PR #20, opened 2019-08-15
Adding codecov to travis

gwaygenomics:


gwaygenomics:
Adding codecov cc @shntnu

PR #72, opened 2020-03-04
Adding Robust Normalization by MAD

gwaygenomics:
The "robustize" normalization strategy is based on [sklearn.preprocessing.RobustScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.RobustScaler.html), which subtracts the median and then **scales data using the interquartile range**.

The original [cytominer "robust" function](https://github.com/cytomining/cytominer/blob/3cde7f08722775226330339d75c6579d6c321e75/R/normalize.R#L78-L81) subtracts the median and then **scales data using the median absolute deviation**.

In this pull request, I add `RobustMAD`. This includes an option to perform robust scaling using the original cytominer method.

### DIFF
--- a/pycytominer/cyto_utils/transform.py
+++ b/pycytominer/cyto_utils/transform.py
@@ -6,6 +6,7 @@ import os
 import numpy as np
 import pandas as pd
 from scipy.linalg import eigh
+from scipy.stats import median_absolute_deviation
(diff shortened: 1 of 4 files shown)


gwaygenomics:
The motivating reason to add this functionality now is described in https://github.com/broadinstitute/lincs-cell-painting/issues/3#issuecomment-591994451

Issue #100, opened 2020-09-24
Add single cell normalization options


gwaygenomics:
The most common entrypoint to pycytominer is in the form of `sqlite` files. These files contain tables for each compartment, which themselves contain single cell morphology measurements.

Increasingly often we need to use normalized single cell profiles. We need to consider how to modify the sqlite files in a way that makes the single cell data easier to compute on.

It is possible that this issue coexists somewhere between `pycytominer` and `cytominer-database`.

gwaygenomics:
My strategy here will be to abstract out the `AggregateProfiles()` class into a new cytominer `singlecell.py` method.

gwaygenomics:
I ended up extracting `AggregateProfiles` from `aggregate.py` into a new script `cyto_utils/cells.py`.

@hillsbury noted that the single cell normalization procedure for aggregate profiles is not likely to work the same as for bulk profiles.

gwaygenomics:
Normalization options for single cell data still need to be explored. The single cell module is now available after merging #111

PR #111, opened 2020-12-10
Adding single cell module

gwaygenomics:
Implementing the strategy outlined in https://github.com/cytomining/pycytominer/issues/100#issuecomment-698893475


gwaygenomics:
todo: add test to `pycytominer.cyto_utils.cells.py` in `merge_single_cells` method

Issue #123, opened 2021-01-28
Add option to aggregate only a subset of the data (such as only specific sites)


shntnu:

shntnu:
@niranjchandrasekaran @gwaygenomics do you recollect where we might have made notes on what to do here? It's possible Ruifan can prioritize this, but he'd need some pointers on where to start.

gwaygenomics:
I don't know if we had any specific notes, but I agree that they would be helpful!

The key places to make this addition will be in [`pycytominer.cyto_utils.cells.load_image()](https://github.com/cytomining/pycytominer/blob/master/pycytominer/cyto_utils/cells.py#L167-L172) to add more columns from image.csv. The next place will be [`pycytominer.cyto_utils.cells.load_compartment()](https://github.com/cytomining/pycytominer/blob/master/pycytominer/cyto_utils/cells.py#L264-L267) to add functionality to perform specific subsets. If the subsetting gets particularly complex, then it might be worth abstracting `load_compartment` into a separate function to import

shntnu:
Great! And you don't expect `aggregate_compartment()` and `merge_single_cells()` to need modifying, right @gwaygenomics ?

gwaygenomics:
shouldn't need modification, correct

gwaybio:
closed by #129

Issue #120, opened 2021-01-07
Make image features modular


gwaygenomics:
Right now, image columns are hardcoded in loading the image csv table

https://github.com/cytomining/pycytominer/blob/91afadc0ec3148e153f6ba5b37bd1100b9eb7577/pycytominer/cyto_utils/cells.py#L167-L172

We need to make this a more modular argument, separate from `self.strata`, that will incorporate more image features than the small subset

PR #129, opened 2021-03-10
Modify aggregate

niranjchandrasekaran:
In this PR, I attempt to add the following functionalities to pycytominer

- Aggregate profiles from only specific sites - https://github.com/cytomining/pycytominer/issues/123 
- Add object count and number of sites to the profiles as per Anne and Shantanu's request on Slack

> Anne: we really should have a bit of QC in the pipeline for each experiment that says how mean cells/sample and sites/sample to be sure that checks out.

> Shantanu: would it be possible to add (counts, number of sites) as outputs of https://github.com/jump-cellpainting/profiling-recipe/blob/refactor_recipe/profiles/profile.py

Instead of outputting the counts separately, I have added them as metadata to the profiles.

### Things changed/added

The following is the description of things I added/changed to the code

#### `pycytominer/aggregate.py`
- `aggregate`
	- Added the parameters `compute_object_count` and `object_feature` to the `aggregate` function.
		- `compute_object_count` defaults to False. This tells `aggregate` whether to count objects or not.
		- `object_feature` defaults to `ObjectNumber`. This stores the name of the metadata feature that distinguishes objects. I believe `ObjectNumber` is used in CellProfiler4 (I need to confirm this). In an older dataset I found this feature's name to be `Metadata_ObjectNumber`.
	- Added code to compute the number of objects. A new metadata feature `Metadata_Object_Count` is added to the profile with the number of objects per sample. 

#### `pycytominer/cyto_utils/util.py`
- Added two functions `check_fields_of_view_format` and `check_fields_of_view`.
	- `check_fields_of_view_format` checks if the input fields of view is an array of integers. If not, it raises TypeError.
	- `check_fields_of_view` checks to ensure that the entered list of fields of view is a subset of the fields of view in the data. If not, it raises a ValueError.

#### `pycytominer/cyto_utils/__init__.py`
- Added the functions `check_fields_of_view_format` and `check_fields_of_view`.

#### `pycytominer/cyto_utils/cells.py`
- `__init__`
	- Imported the functions `check_fields_of_view_format` and `check_fields_of_view` in `pycytominer/cyto_utils/cells.py`.
	- Added the parameters `fields_of_view`, `fields_of_view_feature` and `object_feature`.
		- `fields_of_view` defaults to `all`. This stores the list of fields of view input by the user.
		- `fields_of_view_feature` defaults to `Metadata_FieldID`. This stores the name of the field of view metadata feature. In older data `Metadata_Site` 
		- `object_feature` defaults to `ObjectNumber`. This stores the name of the object number metadata feature.
	- If the user inputs custom fields of view, then `check_fields_of_view_format` is called to check if the input format is correct.
- `load_image`
	- Functionality is added to filter out fields based on the value of `fields_of_view`.
		- `check_fields_of_view` is called to ensure the input fields of view is a subset of the fields of view available in the data before fields are filtered out.
- `aggregate_compartment`
	- Added the parameter `compute_count`. 
		- `compute_count` defaults to False. It should be set to True only when object count and field of view count should be added to the profiles.
		- Added code to compute the number of fields of view. A new metadata feature `Metadata_Fields_Count` is added to the profile with the number of fields of view per sample.
- `aggregate_profiles`
	- When the `aggregate_compartment` is called for the first time, `compute_counts` is set to True. This assumes that the object count is the same for all compartments (@shntnu Is it true that the object count is same for all compartments?)

#### Minor changes

- In the docstring of the function `merge_single_cells` in `pycytominer/cyto_utils/cell.py` I fixed a typo by changing `compression` to changed to `compression_options`.

### Example

I took a single plate of data from @gwaygenomics's pipeline example - https://github.com/cytomining/pipeline-examples/blob/main/0.download.sh and tested the modifications above. The following is an example of how to specify the fields to view to aggregate and how the profiles look like. In the first example only 3 fields are aggregated. In the second example all the available fields are aggregated.

https://nbviewer.jupyter.org/github/niranjchandrasekaran/explore-aggregate/blob/main/run-aggregate.ipynb (I have given up on GitHub's ability to display Jupyter notebooks)

### Things not yet done
Anne had also asked to add the number of channels.

> Anne: And also number of channels (for when we wonder if we had brifghtfield or not)

I was not able to find this information in the sqlite file. @shntnu, I know that we can get this information from the Image.idx.xml file ([like I had done previously](https://github.com/jump-cellpainting/pilot-analysis/issues/41)) but since pycytominer does not interact with the xml file, I was wondering if there is another way to do this.

### DIFF
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -41,7 +41,7 @@ def aggregate(
     compute_object_count : bool, default False
         Whether or not to compute object counts.
     object_feature : str, default "ObjectNumber"
-        Object number feature.
+        Object number feature. Only used if compute_object_count=True.
     subset_data_df : pandas.core.frame.DataFrame
         How to subset the input.
     compression_options : str, optional
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -70,6 +70,8 @@ def aggregate(
     else:
         population_df = population_df.mean().reset_index()
 
+    print(list(population_df.columns))
+
     # Aggregated image number and object number do not make sense
     for col in ["ImageNumber", "ObjectNumber"]:
         if col in population_df.columns:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -48,6 +48,7 @@ def aggregate(
         ).reindex(population_df.columns, axis="columns")
 
     # Subset dataframe to only specified variables if provided
+    print(list(population_df.columns))
     strata_df = population_df.loc[:, strata]
     if features == "infer":
         features = infer_cp_features(population_df)
@@ -69,8 +70,6 @@ def aggregate(
         population_df = population_df.median().reset_index()
     else:
         population_df = population_df.mean().reset_index()
-
-    print(list(population_df.columns))
 
     # Aggregated image number and object number do not make sense
     for col in ["ImageNumber", "ObjectNumber"]:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -79,7 +79,7 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
-        count_object_df = population_df_copy.loc[:, strata + [object_feature]]
+        count_object_df = population_df_copy.loc[:, np.union1d(strata, [object_feature])]
         count_object_df = (
             count_object_df.groupby(strata)[object_feature]
             .count()
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -51,7 +51,7 @@ def aggregate(
 
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
-    count_df = population_df.loc[:, strata + ['Metadata_ObjectNumber']]
+    count_object_df = population_df.loc[:, strata + ['Metadata_ObjectNumber']]
     if features == "infer":
         features = infer_cp_features(population_df)
         population_df = population_df.loc[:, features]
@@ -75,13 +75,13 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
-        count_df = (
-            count_df.groupby(strata)['Metadata_ObjectNumber']
+        count_object_df = (
+            count_object_df.groupby(strata)['Metadata_ObjectNumber']
             .count()
             .reset_index()
             .rename(columns={'Metadata_ObjectNumber': f'Metadata_Object_Count'})
         )
-        population_df = count_df.merge(population_df, on=strata, how='right')
+        population_df = count_object_df.merge(population_df, on=strata, how='right')
 
 
     # Aggregated image number and object number do not make sense
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -23,25 +23,39 @@ def aggregate(
     compression_options=None,
     float_format=None,
 ):
-    """
-    Combine population dataframe variables by strata groups using given operation
+    """Combine population dataframe variables by strata groups using given operation.
 
-    Arguments:
-    population_df - pandas DataFrame to group and aggregate
-    strata - [default: ["Metadata_Plate", "Metadata_Well"]] list indicating the columns to groupby and aggregate
-    features - [default: "all"] or list indicating features that should be aggregated
-    operation - [default: "median"] a string indicating how the data is aggregated
-                currently only supports one of ['mean', 'median']
-    output_file - [default: "none"] if provided, will write aggregated profiles to file
-                  if not specified, will return the aggregated profiles. We recommend
-                  naming the file based on the plate name.
-    compute_object_count - [default: False] determine whether to compute object counts
-    object_feature - [default: "ObjectNumber"] Object number feature
-    subset_data_df - [default: "none"] a pandas dataframe indicating how to subset the input
+    Parameters
+    ----------
+    population_df : pandas.core.frame.DataFrame
+        DataFrame to group and aggregate.
+    strata : list of str, default ["Metadata_Plate", "Metadata_Well"]
+        Columns to groupby and aggregate.
+    features : list of str, default "all"
+        List of features that should be aggregated.
+    operation : str, default "median"
+        How the data is aggregated. Currently only supports one of ['mean', 'median'].
+    output_file : str or file handle, optional
+        If provided, will write aggregated profiles to file. If not specified, will return the aggregated profiles.
+        We recommend naming the file based on the plate name.
+    compute_object_count : bool, default False
+        Whether or not to compute object counts.
+    object_feature : str, default "ObjectNumber"
+        Object number feature.
+    subset_data_df : pandas.core.frame.DataFrame
+        How to subset the input.
+    compression_options : str, optional
+        The mechanism to compress.
+    float_format : str, optional
+        Decimal precision to use in writing output file.
 
-    Return:
-    Pandas DataFrame of aggregated features
+    Returns
+    -------
+    pandas.core.frame.DataFrame
+        DataFrame of aggregated features.
+
     """
+
     # Check that the operation is supported
     operation = check_aggregate_operation(operation)
 
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -53,7 +53,6 @@ def aggregate(
 
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
-    count_object_df = population_df.loc[:, strata + [object_feature]]
     if features == "infer":
         features = infer_cp_features(population_df)
         population_df = population_df.loc[:, features]
@@ -77,6 +76,7 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
+        count_object_df = population_df.loc[:, strata + [object_feature]]
         count_object_df = (
             count_object_df.groupby(strata)[object_feature]
             .count()
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -75,7 +75,7 @@ def aggregate(
             count_object_df.groupby(strata)[object_feature]
             .count()
             .reset_index()
-            .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
+            .rename(columns={f"{object_feature}": "Metadata_Object_Count"})
         )
 
     if features == "infer":
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -51,6 +51,9 @@ def aggregate(
             population_df, how="left", on=subset_data_df.columns.tolist()
         ).reindex(population_df.columns, axis="columns")
 
+    # Create a copy of the population dataframe before Metadata features are removed
+    population_df_copy = population_df.copy()
+
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
     if features == "infer":
@@ -76,7 +79,7 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
-        count_object_df = population_df.loc[:, strata + [object_feature]]
+        count_object_df = population_df_copy.loc[:, strata + [object_feature]]
         count_object_df = (
             count_object_df.groupby(strata)[object_feature]
             .count()
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -14,11 +14,11 @@ from pycytominer.cyto_utils import (
 def aggregate(
     population_df,
     strata=["Metadata_Plate", "Metadata_Well"],
-    compute_object_count=False,
-    object_feature="ObjectNumber",
     features="infer",
     operation="median",
     output_file="none",
+    compute_object_count=False,
+    object_feature="ObjectNumber",
     subset_data_df="none",
     compression_options=None,
     float_format=None,
@@ -29,14 +29,14 @@ def aggregate(
     Arguments:
     population_df - pandas DataFrame to group and aggregate
     strata - [default: ["Metadata_Plate", "Metadata_Well"]] list indicating the columns to groupby and aggregate
-    compute_object_count - [default: False] determine whether to compute object counts
-    object_feature - [default: "ObjectNumber"] Object number feature
     features - [default: "all"] or list indicating features that should be aggregated
     operation - [default: "median"] a string indicating how the data is aggregated
                 currently only supports one of ['mean', 'median']
     output_file - [default: "none"] if provided, will write aggregated profiles to file
                   if not specified, will return the aggregated profiles. We recommend
                   naming the file based on the plate name.
+    compute_object_count - [default: False] determine whether to compute object counts
+    object_feature - [default: "ObjectNumber"] Object number feature
     subset_data_df - [default: "none"] a pandas dataframe indicating how to subset the input
 
     Return:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -15,6 +15,7 @@ def aggregate(
     population_df,
     strata=["Metadata_Plate", "Metadata_Well"],
     compute_object_count=False,
+    object_feature="ObjectNumber",
     features="infer",
     operation="median",
     output_file="none",
@@ -29,6 +30,7 @@ def aggregate(
     population_df - pandas DataFrame to group and aggregate
     strata - [default: ["Metadata_Plate", "Metadata_Well"]] list indicating the columns to groupby and aggregate
     compute_object_count - [default: False] determine whether to compute object counts
+    object_feature - [default: "ObjectNumber"] Object number feature
     features - [default: "all"] or list indicating features that should be aggregated
     operation - [default: "median"] a string indicating how the data is aggregated
                 currently only supports one of ['mean', 'median']
@@ -51,7 +53,7 @@ def aggregate(
 
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
-    count_object_df = population_df.loc[:, strata + ['Metadata_ObjectNumber']]
+    count_object_df = population_df.loc[:, strata + [object_feature]]
     if features == "infer":
         features = infer_cp_features(population_df)
         population_df = population_df.loc[:, features]
@@ -76,13 +78,12 @@ def aggregate(
     # Compute objects counts
     if compute_object_count:
         count_object_df = (
-            count_object_df.groupby(strata)['Metadata_ObjectNumber']
+            count_object_df.groupby(strata)[object_feature]
             .count()
             .reset_index()
-            .rename(columns={'Metadata_ObjectNumber': f'Metadata_Object_Count'})
+            .rename(columns={f'{object_feature}': f'Metadata_Object_Count'})
         )
         population_df = count_object_df.merge(population_df, on=strata, how='right')
-
 
     # Aggregated image number and object number do not make sense
     for col in ["ImageNumber", "ObjectNumber"]:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -51,11 +51,19 @@ def aggregate(
             population_df, how="left", on=subset_data_df.columns.tolist()
         ).reindex(population_df.columns, axis="columns")
 
-    # Create a copy of the population dataframe before Metadata features are removed
-    population_df_copy = population_df.copy()
-
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
+
+    # Only extract single object column in preparation for count
+    if compute_object_count:
+        count_object_df = population_df.loc[:, np.union1d(strata, [object_feature])]
+        count_object_df = (
+            count_object_df.groupby(strata)[object_feature]
+            .count()
+            .reset_index()
+            .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
+        )
+
     if features == "infer":
         features = infer_cp_features(population_df)
         population_df = population_df.loc[:, features]
@@ -79,15 +87,6 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
-        count_object_df = population_df_copy.loc[
-            :, np.union1d(strata, [object_feature])
-        ]
-        count_object_df = (
-            count_object_df.groupby(strata)[object_feature]
-            .count()
-            .reset_index()
-            .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
-        )
         population_df = count_object_df.merge(population_df, on=strata, how="right")
 
     # Aggregated image number and object number do not make sense
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -81,9 +81,9 @@ def aggregate(
             count_object_df.groupby(strata)[object_feature]
             .count()
             .reset_index()
-            .rename(columns={f'{object_feature}': f'Metadata_Object_Count'})
+            .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
         )
-        population_df = count_object_df.merge(population_df, on=strata, how='right')
+        population_df = count_object_df.merge(population_df, on=strata, how="right")
 
     # Aggregated image number and object number do not make sense
     for col in ["ImageNumber", "ObjectNumber"]:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -14,6 +14,7 @@ from pycytominer.cyto_utils import (
 def aggregate(
     population_df,
     strata=["Metadata_Plate", "Metadata_Well"],
+    compute_object_count=False,
     features="infer",
     operation="median",
     output_file="none",
@@ -27,6 +28,7 @@ def aggregate(
     Arguments:
     population_df - pandas DataFrame to group and aggregate
     strata - [default: ["Metadata_Plate", "Metadata_Well"]] list indicating the columns to groupby and aggregate
+    compute_object_count - [default: False] determine whether to compute object counts
     features - [default: "all"] or list indicating features that should be aggregated
     operation - [default: "median"] a string indicating how the data is aggregated
                 currently only supports one of ['mean', 'median']
@@ -48,8 +50,8 @@ def aggregate(
         ).reindex(population_df.columns, axis="columns")
 
     # Subset dataframe to only specified variables if provided
-    print(list(population_df.columns))
     strata_df = population_df.loc[:, strata]
+    count_df = population_df.loc[:, strata + ['Metadata_ObjectNumber']]
     if features == "infer":
         features = infer_cp_features(population_df)
         population_df = population_df.loc[:, features]
@@ -70,6 +72,17 @@ def aggregate(
         population_df = population_df.median().reset_index()
     else:
         population_df = population_df.mean().reset_index()
+
+    # Compute objects counts
+    if compute_object_count:
+        count_df = (
+            count_df.groupby(strata)['Metadata_ObjectNumber']
+            .count()
+            .reset_index()
+            .rename(columns={'Metadata_ObjectNumber': f'Metadata_Object_Count'})
+        )
+        population_df = count_df.merge(population_df, on=strata, how='right')
+
 
     # Aggregated image number and object number do not make sense
     for col in ["ImageNumber", "ObjectNumber"]:
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -79,7 +79,9 @@ def aggregate(
 
     # Compute objects counts
     if compute_object_count:
-        count_object_df = population_df_copy.loc[:, np.union1d(strata, [object_feature])]
+        count_object_df = population_df_copy.loc[
+            :, np.union1d(strata, [object_feature])
+        ]
         count_object_df = (
             count_object_df.groupby(strata)[object_feature]
             .count()
--- a/pycytominer/cyto_utils/__init__.py
+++ b/pycytominer/cyto_utils/__init__.py
@@ -8,7 +8,7 @@ from .util import (
     check_consensus_operation,
     get_pairwise_correlation,
     check_fields_of_view_format,
-    check_fields_of_view
+    check_fields_of_view,
 )
 from .single_cell_ingest_utils import (
     get_default_linking_cols,
--- a/pycytominer/cyto_utils/__init__.py
+++ b/pycytominer/cyto_utils/__init__.py
@@ -7,6 +7,8 @@ from .util import (
     check_aggregate_operation,
     check_consensus_operation,
     get_pairwise_correlation,
+    check_fields_of_view_format,
+    check_fields_of_view
 )
 from .single_cell_ingest_utils import (
     get_default_linking_cols,
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -79,6 +79,7 @@ class SingleCells(object):
         subsample_n="all",
         subsampling_random_state="none",
         fields_of_view="all",
+        fields_of_view_feature = "Metadata_Site",
         object_feature="ObjectNumber",
     ):
         """Constructor method"""
@@ -109,7 +110,7 @@ class SingleCells(object):
         self.compartments = compartments
         self.compartment_linking_cols = compartment_linking_cols
         self.fields_of_view = fields_of_view
-        self.fields_of_view_feature = "Metadata_Site"
+        self.fields_of_view_feature = fields_of_view_feature
         self.object_feature = object_feature
 
         # Confirm that the compartments and linking cols are formatted properly
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -105,7 +105,7 @@ class SingleCells(object):
         self.compartments = compartments
         self.compartment_linking_cols = compartment_linking_cols
         self.fields_of_view = fields_of_view
-        self.fields_of_view_feature = 'Metadata_Site'
+        self.fields_of_view_feature = "Metadata_Site"
         self.object_feature = object_feature
 
         self.image_cols = ["TableNumber", "ImageNumber", "Metadata_Site"]
@@ -183,7 +183,9 @@ class SingleCells(object):
     def load_image(self):
         """Load image table from sqlite file"""
         # Extract image metadata
-        image_query = "select {} from image".format(', '.join(self.image_cols + self.strata))
+        image_query = "select {} from image".format(
+            ", ".join(self.image_cols + self.strata)
+        )
         self.image_df = pd.read_sql(sql=image_query, con=self.conn)
         if self.fields_of_view != "all":
             check_fields_of_view(
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -45,7 +45,7 @@ class SingleCells(object):
     :type subsample_n:, str, int
     :param subsampling_random_state: the random state to init subsample, defaults to "none"
     :type subsampling_random_state: str, int
-    :param fields_of_view: list of field ID to include in the analysis, defaults to "all"
+    :param fields_of_view: list of fields of view to include in the analysis, defaults to "all"
     :type fields_of_view: list, str
 
     .. note::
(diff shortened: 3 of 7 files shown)


pycytominer/aggregate.py
```diff
@@ -49,6 +53,7 @@ def aggregate(
 
     # Subset dataframe to only specified variables if provided
     strata_df = population_df.loc[:, strata]
+    count_object_df = population_df.loc[:, strata + [object_feature]]
```
gwaygenomics, line 56: Can you move this line to only execute if `compute_object_count=True`? I think it shouldn't execute unless we use the `count_object_df` variable.

pycytominer/cyto_utils/__init__.py
```diff
@@ -7,6 +7,8 @@
     check_aggregate_operation,
     check_consensus_operation,
     get_pairwise_correlation,
+    check_fields_of_view_format,
+    check_fields_of_view
```
gwaygenomics, line 11: make sure to run `black`. I spot a missing comma! :)

pycytominer/aggregate.py
```diff
@@ -27,6 +29,8 @@ def aggregate(
     Arguments:
     population_df - pandas DataFrame to group and aggregate
     strata - [default: ["Metadata_Plate", "Metadata_Well"]] list indicating the columns to groupby and aggregate
+    compute_object_count - [default: False] determine whether to compute object counts
```
gwaygenomics, line 32: in preparation for pycytominer release, can you make sure that any comment docstring that you touch is written in numpy sphinx format? A comprehensive example is here: https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

I've used it with some luck in [`cytominer_eval`](https://github.com/cytomining/cytominer-eval/blob/master/cytominer_eval/evaluate.py#L34)

pycytominer/aggregate.py
```diff
@@ -70,6 +75,16 @@ def aggregate(
...
 
+    # Compute objects counts
+    if compute_object_count:
+        count_object_df = (
+            count_object_df.groupby(strata)[object_feature]
+            .count()
+            .reset_index()
+            .rename(columns={f'{object_feature}': f'Metadata_Object_Count'})
+        )
+        population_df = count_object_df.merge(population_df, on=strata, how='right')
```
gwaygenomics, line 86: needs to be blacked

pycytominer/aggregate.py
```diff
@@ -14,6 +14,8 @@
 def aggregate(
     population_df,
     strata=["Metadata_Plate", "Metadata_Well"],
+    compute_object_count=False,
```
gwaygenomics, line 17: Can you move these to between `output_file` and `subset_data_df`? I think that the argument ordering should follow in order of how often a user interacts with it. In my mind:

Top args - most changed
output_file - the last argument in the top args (it has the biggest effect)
output_file args - any args pertaining to how the file is handled

also, do I even use `subset_data_df` in this function? 😬  If not, can you delete it?

pycytominer/cyto_utils/cells.py
```diff
@@ -167,9 +184,12 @@ def set_subsample_random_state(self, random_state):
     def load_image(self):
         """Load image table from sqlite file"""
         # Extract image metadata
-        image_cols = "TableNumber, ImageNumber, {}".format(", ".join(self.strata))
+        image_cols = "TableNumber, ImageNumber, Metadata_Site, {}".format(", ".join(self.strata))
```
gwaygenomics, line 187: I was hoping to eventually move this hardcoded `image_cols` into something that's defined upon class instantiation. Can you make this upgrade now? I think it should be fairly straightforward to move `image_cols` to be defined as `self.image_cols=["TableNumber", "ImageNumber", "Metadata_Site"]` in `SingleCells.__init__()`.

You'll then need to do some wrangling in `load_image()` to get it ready for sql ingestion.

gwaygenomics: see #120

pycytominer/cyto_utils/util.py
```diff
@@ -124,6 +125,43 @@ def check_consensus_operation(operation):
...
+    Return:
+    Correctly formatted fields_of_view variable
+    """
+
+    if fields_of_view != "all":
+        if isinstance(fields_of_view, list):
+            if all(isinstance(x, int) for x in fields_of_view):
+                return fields_of_view
+            else:
+                raise TypeError(f'Variable of type int expected, however type {type(fields_of_view[0])} was passed')
```
gwaygenomics, line 144: i can see this error message frustrating a user if the type of the first element in `fields_of_view` is an int, but one of the other elements is not. Is there a way to account for this in the error message? If not, I don't think we need to try too hard to be specific.

gwaygenomics (changes requested):
Looks great @niranjchandrasekaran - thanks for this important contribution! Love the comprehensive description...

I made several comments in the pull request. I'll need to go through it once more after you are able to address.

A major thing left to do is to add tests.

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: I have moved `image_cols` to `SingleCells.__init__()`. It looked like there will be duplicate columns if `image_cols` and `strata` shared features (for e.g. `Metadata_Site`). So I have also removed duplicates using `np.union1d()`

pycytominer/aggregate.py
niranjchandrasekaran: Done!

pycytominer/cyto_utils/__init__.py
niranjchandrasekaran: Done! :)

pycytominer/aggregate.py
niranjchandrasekaran: Does [this](https://github.com/cytomining/pycytominer/pull/129/commits/b2446a4b7b580474d8ed4efa9511643a8109a5c5) look right? If so I will change the documentation for all the functions in these 3 or 4 files that I have touched.

pycytominer/aggregate.py
niranjchandrasekaran: I wasn't able to because `population_df` does not have all the metadata fields at this point (i.e. no `object_feature`). So I have done something like [this](https://github.com/cytomining/pycytominer/pull/129/commits/45684d4a9b1ac4ea61bbd5082cffcdde0d46175e) but this looks clunky as well.

pycytominer/aggregate.py
niranjchandrasekaran: > Can you move these to between `output_file` and `subset_data_df`? 

Done

> also, do I even use subset_data_df in this function? 😬 If not, can you delete it? 

You do :) So I have left it as is.

pycytominer/cyto_utils/util.py
niranjchandrasekaran: Changed it a little - https://github.com/cytomining/pycytominer/pull/129/commits/44f49eb552f9ff9adff012fb01edcd68f89328f7

pycytominer/aggregate.py
gwaygenomics: oh i see the issue. Yeah, I also agree that the current solution is clunky. We shouldn't create a copy of input data unless we absolutely need to, and especially not by default. Imagine running this function on really large data - watch out!

I think that we can avoid this issue by instead using two `if compute_object_count` statements.

One after defining `strata_df`:

```python
strata_df = population_df.loc[:, strata]

# Only extract single object column in preparation for count
if compute_object_count:
    count_object_df = population_df_copy.loc[:, strata + [object_feature]]
    count_object_df = (
        count_object_df.groupby(strata)[object_feature]
        .count()
        .reset_index()
        .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
    )

if features == "infer":
    features = infer_cp_features(population_df)
    population_df = population_df.loc[:, features]
else:
    population_df = population_df.loc[:, features]
```

Then, after performing the actual aggregation:

```python
if compute_object_count:
    population_df = count_object_df.merge(population_df, on=strata, how="right")
```

How does this sound? I'd prefer a solution avoiding duplicate data while maintaining readability.

pycytominer/aggregate.py
gwaygenomics: looks great 👍 please use consistent punctuation though

pycytominer/cyto_utils/cells.py
gwaygenomics: the `np.union1d()` is a beautiful idea.

Let's go one step further with defining `image_cols` in `SingleCells.__init__()`. How about instead of hardcoding them inside `__init__()` (which is one step better than hardcoding for sure), but we can also add them as a default argument upon class instantiation (two steps better!). In other words, 

```python
class SingleCells(object):
    """yada yada yada
    """
    def __init__(
        self,
        file_or_conn,
        ....,
        image_cols=["TableNumber", "ImageNumber", "Metadata_Site"]
    ):
    self.image_cols = image_cols
```

This will enable us to use any image column in future profiles. Delightfully devilish!

pycytominer/cyto_utils/util.py
gwaygenomics: I'm wondering if we should try to first coerce to integers, and then throw an error if it fails. What do you think about that? Maybe a bad idea, IDK

pycytominer/cyto_utils/cells.py
```diff
@@ -267,14 +292,20 @@ def load_compartment(self, compartment):
         return df
 
     def aggregate_compartment(
-        self, compartment, compute_subsample=False, aggregate_args=None
+        self,
+        compartment,
+        compute_subsample=False,
+        compute_counts=False,
```
gwaygenomics, line 298: instead of adding `compute_subsample` as an argument, can you use `compute_subsample` through `aggregate_args`?

In other words, the current arguments stay the same (preferring a minimalistic core), and the function accesses the variable through `aggregate_args`? Actually no, nvm, I see now that you're also adding functionality directly here. In this case, I think how you have it makes sense. Anything passing through aggregate.py only should be included in `aggregate_args`. 

By that standard, then I think you should add `self.object_feature` to the aggregate args instead of directly to the `aggregate` call.

Something like:

```python
aggregate_args["features"] = features
aggregate_args["object_feature"] = self.object_feature
```

To me, this seems like a slightly more consistent API, but it can go either way really.

pycytominer/cyto_utils/cells.py
```diff
@@ -312,11 +343,28 @@ def aggregate_compartment(
...
             operation=self.aggregation_operation,
             subset_data_df=self.subset_data_df,
-            **aggregate_args
+            object_feature=self.object_feature,
+            **aggregate_args,
         )
 
+        if compute_counts:
+            fields_count_df = self.image_df.loc[
+                :, self.strata + [self.fields_of_view_feature]
```
gwaygenomics, line 355: should we also check the feature union for this?

pycytominer/cyto_utils/cells.py
```diff
@@ -267,14 +292,20 @@ def load_compartment(self, compartment):
...
+        compute_counts=False,
+        aggregate_args=None,
     ):
         """Aggregate morphological profiles. Uses pycytominer.aggregate()
 
         :param compartment: string indicating the specific compartment, defaults to "cells"
         :type compartment: str
         :param compute_subsample: determine if subsample should be computed, defaults to False
         :type compute_subsample: bool
+        :param compute_counts: determine if the number of the objects and fields should be computed, defaults to False
```
gwaygenomics, line 307: can you add to this docstring slightly? can you add one example of what an object and field is? I found myself getting lost a couple of lines down as to what "counts" i was computing and I jumped back here. An example would be crystal clear

pycytominer/cyto_utils/cells.py
```diff
@@ -494,7 +542,9 @@ def aggregate_profiles(
         for compartment in self.compartments:
             if compartment_idx == 0:
                 aggregated = self.aggregate_compartment(
-                    compartment=compartment, compute_subsample=compute_subsample
+                    compartment=compartment,
+                    compute_subsample=compute_subsample,
+                    compute_counts=True,
```
gwaygenomics, line 547: Is this always going to be True? I am usually in favor of keeping the default behavior the most simple as possible. Is it true that adding True would be more complex? Maybe we always want to compute counts though?

gwaygenomics (changes requested):
I added to our original discussion, and made a couple additional comments here on new stuff.

There are three more things to do before I approve:

1. We come to a consensus on the previous discussion points. I do not hold firm opinions on many points.
2. numpy sphinx docstring formatting for all functions. This will help the package a ton to start doing this asap.
3. Pass old and add new tests

pycytominer/aggregate.py
niranjchandrasekaran: This makes sense! Implemented your suggestion -https://github.com/cytomining/pycytominer/pull/129/commits/3dcbe344595981cf8fd45727681c4b8fff617f88

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Ah I see what you mean. Implemented here - https://github.com/cytomining/pycytominer/pull/129/commits/ba0c15ea782df5f103341a13c77b4ff239da1c8e

pycytominer/cyto_utils/util.py
niranjchandrasekaran: Here is my attempt to coerce to integers - https://github.com/cytomining/pycytominer/pull/129/commits/789c6ed3abfe3f4a4fbb8526171448521f5e6401

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Makes sense - https://github.com/cytomining/pycytominer/pull/129/commits/1953cffdf96cb42280ed4f3513f9048a1614ceca

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: 👍 https://github.com/cytomining/pycytominer/pull/129/commits/e759d5902e162f645e1e5e936d280da7022315fc

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/b81169d3c93ddb9768e16ed14e73fc475d864041

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: I guess my goal is to always compute counts and make the two new metadata features part of all profiles generated. In JUMP we wanted to know the values of these two counts often and going back to the sqlite made it difficult. So I decided to add these values to the profiles itself. It is possible that for site-wise or single cell profiles, these metrics may not be meaningful. That would argue against not having `compute_counts` as True always. What do you think?

niranjchandrasekaran:
@gwaygenomics I have replied to your comments. I am going to start working on docstrings and writing tests. I don't have a lot of experience with tests. I will go through the ones you have already written and will let you know if I need your help.

pycytominer/cyto_utils/cells.py
```diff
@@ -96,6 +108,9 @@ def __init__(
         self.is_subset_computed = False
         self.compartments = compartments
         self.compartment_linking_cols = compartment_linking_cols
+        self.fields_of_view = fields_of_view
+        self.fields_of_view_feature = "Metadata_Site"
```
gwaygenomics, line 112: is this always going to be "Metadata_Site" - should we move to default __init__?

pycytominer/cyto_utils/cells.py
```diff
@@ -117,6 +132,9 @@ def __init__(
         # Throw an error if both subsample_frac and subsample_n is set
         self._check_subsampling()
 
+        # Confirm that the input fields of view is valid
+        self.fields_of_view = check_fields_of_view_format(self.fields_of_view)
```
gwaygenomics, line 136: nice! I like this attribute assigned in this way. We can also do as suggested, and then drop line 111 where you're originally assigning the attribute. Super minor, but slightly more pythonic IMO
```suggestion
        self.fields_of_view = check_fields_of_view_format(fields_of_view)
```

pycytominer/cyto_utils/cells.py
gwaygenomics: Cool, thanks for describing. In my view, this is the biggest change introduced in this PR! After merging, pyctominer will forever generate a `Metadata_fields_count` column. I think that this behavior is fine, good behavior even if JUMP requires it often, so I am inclined to agree with keeping `compute_counts` as True.

One other question comes to mind with this default behavior. Should we be adding other metadata fields by default? Cell count is one that immediately comes to mind. I think any discussion of this is beyond scope of the present PR, so I opened #130

gwaygenomics: just to confirm, `compute_counts` defaults to True only in the first pass through aggregate_compartments. This is intended since you only need it once per image.csv, correct?

gwaygenomics (commented):
A couple followup discussion.

Happy to help with tests. Once you get a testing system setup, it is fairly straightforward. I recommend installing pytest, and running locally before pushing any commit and relying on CI to test for you. Some tests take long, so you can specify a path to pytest to only test certain files.

shntnu:
> Is it true that the object count is same for all compartments?

No, they can differ, although only very rarely so.

> I know that we can get this information from the Index.idx.xml file (like I had done previously) but since pycytominer does not interact with the xml file, I was wondering if there is another way to do this.

No, that information is present only in the `Index.idx.xml` file.

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: > just to confirm, compute_counts defaults to True only in the first pass through aggregate_compartments. This is intended since you only need it once per image.csv, correct?

That's correct

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/2d27b5c0e91eb08215ef3bcea05c46cc51ad7e33

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/a8d20b2cc5e929736160bdc3ecff37a4aca1a10d

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: > One other question comes to mind with this default behavior. Should we be adding other metadata fields by default? Cell count is one that immediately comes to mind. I think any discussion of this is beyond scope of the present PR, so I opened #130

I also have added a new field called `Metadata_Object_Count` which counts the number of cells. But I am split on how to proceed. 

I wanted to change it to `Metadata_Cell_Count` as that is essentially what is being calculated. But since `compute_counts` is set to True only for the first pass through `aggregate_compartments()`, if the order of the compartments change for some reason, then `Metadata_Cell_Count` won't be appropriate name for the field. `Metadata_Object_Count` is general enough that it would work.

I don't like the idea of having a single field for counting the number of objects given Shantanu's comment (https://github.com/cytomining/pycytominer/pull/129#issuecomment-7993708690). If the counts in each compartment can sometimes vary, then we may want to have multiple object count fields (one per compartment). I don't particularly like this option either because we are introducing too many new fields. But if this is how we want to proceed then I can make peace with it.

Any thoughts @gwaygenomics?

niranjchandrasekaran:
@gwaygenomics I have replied to your comments. I have also updated the docstring for all functions in the three files that I edited (https://github.com/cytomining/pycytominer/pull/129/commits/03ab9a2dd531cc21533e34ffce012d7b9347a31a). Let me know what you think. Next I will update the tests and add new ones.

gwaygenomics:
docstrings look great! I'll perform my next full review once you add tests. Thanks!

niranjchandrasekaran:
I have made a few more changes since my last comment.

- Changed the name of newly created feature `Metadata_Fields_Count` to `Metadata_Site_Count`. - https://github.com/cytomining/pycytominer/pull/129/commits/3e8518419e438fae5af5602bf5f909f8bfd9bddc
- `Metadata_Site_Count` is not created if `Metadata_Site` is in the `strata` - https://github.com/cytomining/pycytominer/pull/129/commits/551c22f1ad8076608644c7e204c5e419f57282f5
- Added new tests `test_check_fields_of_view` and `test_check_fields_of_view_format` - https://github.com/cytomining/pycytominer/pull/129/commits/92e2ad70c77584ced29461e68d75c2309099dddb
- Modified existing tests - https://github.com/cytomining/pycytominer/pull/129/commits/6a0f650bea3eb97776c087f0df1feee53528d6dd

In general, most of the changes I made to the tests had to do with the order of the columns. They seem to have changed due to all the code I have introduced. Hence I had to sort columns before data frames could be compared.

While modifying existing tests, I also had to remove a couple of tests (commented them out) and commented out a part of another test. @gwaygenomics I will need your help to get them to work. The following are the reasons why I commented them out
- In `pycytominer/tests/test_cyto_utils/test_cells::test_aggregate_comparment` I had to comment out the following because merging single cells now doesn't seem to generate the same output as aggregating profiles. Thus suggests that `merge_single_cells` needs to be modified. Not sure how straightforward this will be.

https://github.com/cytomining/pycytominer/blob/6a0f650bea3eb97776c087f0df1feee53528d6dd/pycytominer/tests/test_cyto_utils/test_cells.py#L397-L400

- I commented out `test_aggregate_subsampling_profile` because in the following line, the similarity between a data frame and string was being checked. I don't think I totally understood this test. So I wanted to check with you before I modified it.

https://github.com/cytomining/pycytominer/blob/6a0f650bea3eb97776c087f0df1feee53528d6dd/pycytominer/tests/test_cyto_utils/test_cells.py#L440

- I commented out `test_aggregate_subsampling_profile_compress` because the result and the expected were totally different from one another. I wasn't sure why this was the case.

Other than that all the other tests seem to pass, which makes me happy :)

pycytominer/cyto_utils/cells.py
```diff
@@ -463,28 +623,25 @@ def aggregate_profiles(
...
-        :Example:
+        Parameters
+        ----------
+        compute_subsample : bool, default False
+            Whether or not to compute subsample. compute_subsample must be specified to perform subsampling.
+            The function aggregate_profiles(compute_subsample=True) will apply subsetting even if subsample is initialized.
+        output_file : str, optional
+            The name of a file to output. We recommended that, if provided, the output file be suffixed with "_augmented".
+        compression_options : str, optional
+            The mechanism to compress.
```
gwaygenomics, line 634:
```suggestion
            compression arguments as input to pandas.to_csv() with pandas version >= 1.2
```

pycytominer/cyto_utils/cells.py
```diff
@@ -328,20 +480,28 @@ def merge_single_cells(
...
+        """Given the linking columns, merge single cell data. Normalization is also supported.
+
+        Parameters
+        ----------
+        compute_subsample : bool, default False
+            Whether or not to compute subsample.
+        sc_output_file : str, optional
+            The name of a file to output.
+        compression_options : str, optional
+            The mechanism to compress.
```
gwaygenomics, line 492:
```suggestion
            compression arguments as input to pandas.to_csv() with pandas version > 1.2
```

pycytominer/cyto_utils/cells.py
gwaygenomics: I think it is ok to have multiple metadata columns - remember, "measure everything" :) I'd add a subtitle to the lab's motto though "Measure everything... using sound data science practices ensuring data reproducibility and provenance" (but this doesn't have as nice a ring to it though 😦 )

If you decide to go this route @niranjchandrasekaran (I do think it is up to you to decide), maybe the correct column name is `f"Metadata_Object_Count_{compartment}"`.

If you do not decide to go this route, I think it is at least worth summarizing this discussion in #130

pycytominer/cyto_utils/util.py
```diff
@@ -17,10 +18,33 @@
...
+    list of str
+        Default compartments.
+
+    """
+
     return ["cells", "cytoplasm", "nuclei"]
 
 
 def check_compartments(compartments):
+    """Checks if the input compartments are non canonical compartments.
```
gwaygenomics, line 34:
```suggestion
    """Checks if the input compartments are noncanonical compartments.
```

pycytominer/cyto_utils/util.py
```diff
@@ -38,16 +62,21 @@ def check_compartments(compartments):
 
 
 def load_known_metadata_dictionary(metadata_file=default_metadata_file):
-    """
-    From a tab separated text file (two columns: ["compartment", "feature"]) load
-    previously known metadata columns per compartment
+    """From a tab separated text file (two columns: ["compartment", "feature"]) load
```
gwaygenomics, line 65:
```suggestion
    """From a tab separated text file (two columns: ["compartment", "feature"]), load
```

pycytominer/cyto_utils/util.py
```diff
@@ -38,16 +62,21 @@ def check_compartments(compartments):
...
+    previously known metadata columns per compartment.
+
+    Parameters
+    ----------
+    metadata_file : str
```
gwaygenomics, line 70:
```suggestion
    metadata_file : str, optional
```

pycytominer/cyto_utils/util.py
```diff
@@ -38,16 +62,21 @@ def check_compartments(compartments):
...
+        File location of the metadata text file.
```
gwaygenomics, line 71:
```suggestion
        File location of the metadata text file. Uses a default dictionary if you do not specify.
```

pycytominer/cyto_utils/util.py
```diff
@@ -124,18 +168,83 @@ def check_consensus_operation(operation):
     return operation
 
 
-def get_pairwise_correlation(population_df, method="pearson"):
+def check_fields_of_view_format(fields_of_view):
+    """Confirm that the input fields of view is valid.
+
+    Parameters
+    ----------
+    fields_of_view : list of int, default "all"
```
gwaygenomics, line 176: This does not currently default to "all" - either delete this comment, or fix function arg

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -98,15 +99,16 @@ def build_random_data(
 new_linking_cols["new"] = {"cytoplasm": "ObjectNumber"}
 
 # Setup SingleCells Class
-ap = SingleCells(file_or_conn=file)
+ap = SingleCells(file_or_conn=file, object_feature='Metadata_ObjectNumber')
```
gwaygenomics, line 102: please `black` this file

gwaygenomics: I also notice that the default in `SingleCells` is set to just `ObjectNumber` (instead of `Metadata_ObjectNumber`). Is plain old `ObjectNumber` better than with a Metadata prefix?

pycytominer/tests/test_cyto_utils/test_util.py
```diff
@@ -144,3 +146,42 @@ def test_get_pairwise_correlation():
     expected_result = -0.8
     x_y_cor = pair_df.query("correlation != 0").round(1).correlation.values[0]
     assert x_y_cor == expected_result
+
+
+def test_check_fields_of_view():
+    data_fields_of_view = [1, 3, 4, 5]
+
+    valid_input_fields_of_view = [1, 3, 4]
+    assert check_fields_of_view(data_fields_of_view, valid_input_fields_of_view) is None
```
gwaygenomics, line 155:
```suggestion
    assert check_fields_of_view(data_fields_of_view, valid_input_fields_of_view)
```

I think you can drop `is None`, but I'm not 100% sure

pycytominer/tests/test_cyto_utils/test_util.py
```diff
@@ -144,3 +146,42 @@ def test_get_pairwise_correlation():
...
+    valid_input_fields_of_view = [4, 3, 1, 5]
+    assert check_fields_of_view(data_fields_of_view, valid_input_fields_of_view) is None
+
+    invalid_input_fields_of_view = [2, 6, 7]
+    with pytest.raises(ValueError) as err:
+        check_fields_of_view(data_fields_of_view, invalid_input_fields_of_view)
+        assert str(err) == "Some of the input fields of view are not present in the image table."
+
+
+def test_check_fields_of_view_format():
```
gwaygenomics, line 169: beautiful 🌞

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -416,50 +424,50 @@ def test_aggregate_subsampling_count_cells():
...
-            "Cytoplasm_d": [762.5, 625.0],
-            "Nuclei_a": [683.5, 171.0],
-            "Nuclei_b": [50.5, 625.0],
-            "Nuclei_c": [431.0, 483.0],
-            "Nuclei_d": [519.0, 286.5],
-        }
-    )
-
-    pd.testing.assert_frame_equal(result, expected_result)
+# def test_aggregate_subsampling_profile():
```
gwaygenomics, line 427: I'm going to look into this now, separately from this review. I'll see if I can figure out what's going on

gwaygenomics (changes requested):
some discussion items and suggested items. Overall, looks very, very good

I have three key comments:

1. I'm going to work on the commented out tests now
2. Please add a test when `compute_object_count=True` - Is it true that the existing tests all assume it's false?
3. Lastly, please test behavior for if the variable specified in `object_feature` does not exist (e.g. SingleCells(object_feature="DOES NOT EXIST").

gwaygenomics:
Niranj,

- [ ] pycytominer/tests/test_cyto_utils/test_cells::test_aggregate_comparment 
  - fixed in 035ef069aad2b2071e58b53b1ea5dfc00081ed3b
  - Because `aggregate_compartment( compute_counts=True)` in `SingleCells.aggregate_profiles()`, I needed to modify the call to the traditional `aggregate()` on the `merge_single_cells()` output.
  - Also, I needed to drop `Metadata_Site_Count` from the output columns of `SingleCells.aggregate_profiles()`. This is also because of  `aggregate_compartment( compute_counts=True)` - we add the column from `fields_count_df`
  - I think that it is ok for `aggregate()` and `SingleCells.aggregate_profiles()` to diverge in metadata columns. They just can't diverge in terms of feature columns!

gwaygenomics (approved):

gwaygenomics:
@niranjchandrasekaran LGTM - Feel free to merge when you're ready

pycytominer/aggregate.py
```diff
@@ -49,6 +67,17 @@ def aggregate(
...
     strata_df = population_df.loc[:, strata]
+
+    # Only extract single object column in preparation for count
+    if compute_object_count:
+        count_object_df = population_df.loc[:, np.union1d(strata, [object_feature])]
+        count_object_df = (
+            count_object_df.groupby(strata)[object_feature]
+            .count()
+            .reset_index()
+            .rename(columns={f"{object_feature}": f"Metadata_Object_Count"})
```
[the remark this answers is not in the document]
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/60e229f53ce86e37dcbed29e48d1c141620155a0

pycytominer/cyto_utils/cells.py
```diff
@@ -43,6 +47,10 @@ class SingleCells(object):
     :type subsample_n:, str, int
     :param subsampling_random_state: the random state to init subsample, defaults to "none"
     :type subsampling_random_state: str, int
+    :param fields_of_view: list of fields of view to include in the analysis, defaults to "all"
```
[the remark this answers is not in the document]
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/bda5dae68b5dccd456ed226e0b0b2ce9e1176fa7

pycytominer/cyto_utils/cells.py
```diff
@@ -262,24 +371,50 @@ def get_subsample(self, df=None, compartment="cells", rename_col=True):
...
-        :type compartment: str
-        :param compute_subsample: determine if subsample should be computed, defaults to False
-        :type compute_subsample: bool
-        :param aggregate_args: additional arguments passed as a dictionary as input to pycytominer.aggregate()
-        :type aggregate_args: None, dict
-        :return: Aggregated single-cell profiles
-        :rtype: pd.DataFrame
+        Parameters
+        ----------
+        compartment : str, default "cells"
```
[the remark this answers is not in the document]
niranjchandrasekaran: I have removed "cells" as default https://github.com/cytomining/pycytominer/pull/129/commits/d354a30c565dbfe10b38ff078d1bdc205bb09e1c

pycytominer/cyto_utils/cells.py
```diff
@@ -262,24 +371,50 @@ def get_subsample(self, df=None, compartment="cells", rename_col=True):
...
+            Compartment to aggregate.
+        compute_subsample : bool, default False
+            Whether or not to subsample.
+        compute_counts : bool, default False
```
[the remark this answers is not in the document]
niranjchandrasekaran: Currently two different kinds of counts are being computed - site count and object count. So I grouped them together. Would it be better if I separate them into `compute_site_count` and compute_object_count` instead of `compute_counts`? It feels like separating them if better as they may be other counts we would want to compute in the future

pycytominer/cyto_utils/cells.py
```diff
@@ -308,15 +443,32 @@ def aggregate_compartment(
             features = aggregate_args["features"]
 
         aggregate_args["features"] = features
+        aggregate_args["object_feature"] = self.object_feature
```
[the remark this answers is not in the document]
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/6bf5310588ee7cb68c3dde400a9e138e3dcf9a10

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/bc1b388f328b0781902a5c8586655f57838fa5b3

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: Done - https://github.com/cytomining/pycytominer/pull/129/commits/bc1b388f328b0781902a5c8586655f57838fa5b3

pycytominer/cyto_utils/util.py
niranjchandrasekaran: I have some thoughts. I will move the discussion to #130

pycytominer/cyto_utils/util.py
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/2116c46fdcfca0f73b0f7d6f0f14169056308ce4

pycytominer/cyto_utils/util.py
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/2116c46fdcfca0f73b0f7d6f0f14169056308ce4

pycytominer/cyto_utils/util.py
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/2116c46fdcfca0f73b0f7d6f0f14169056308ce4

pycytominer/cyto_utils/util.py
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/6bfee56de80393a31ca93fd139bae6265fb7733f

pycytominer/cyto_utils/cells.py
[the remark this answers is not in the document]
gwaygenomics: oh I see, it's probably ok for now as is

pycytominer/tests/test_cyto_utils/test_cells.py
niranjchandrasekaran: https://github.com/cytomining/pycytominer/pull/129/commits/5e5ad2b55ca16d0790b8cd23a148388da832e93c

`ObjectNumber` was the name in the new JUMP pilot data. I am guessing (not certain) this is what is output by newer versions of CellProfiler. Hence the change.

pycytominer/tests/test_cyto_utils/test_util.py
niranjchandrasekaran: Tried dropping `is None` the test failed. So leaving it as is.

pycytominer/tests/test_cyto_utils/test_util.py
niranjchandrasekaran: :)

PR #153, opened 2021-07-07
Add "noise removal" feature selection 

ruifanp:

# Description

Added noise removal feature selection. Demonstrated to make features more informative in multiple sets.

# Checklist

- [ ] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [ ] My code follows the style guidelines of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have deleted all non-relevant text in this pull request template.

### DIFF
--- a/pycytominer/operations/noise_removal.py
+++ b/pycytominer/operations/noise_removal.py
@@ -1,0 +1,35 @@
+"""
+Remove noisy features, as defined by features with excessive standard deviation within the same perturbation group.
+"""
+
+import numpy as np
+import pandas as pd
+from pycytominer.cyto_utils import infer_cp_features
+
+def noise_removal(
+    population_df, features="infer", samples="all", stdev_cutoff=0.8,
+):
+    """
+    Removes features which are insufficiently reproducible within the same perturbation group.
+    Args:
+        population_df:
+        features:
+        samples:
+        stdev_cutoff:
+
+    Returns:
+
+    """
+    # Subset dataframe
+    if samples != "all":
+        population_df = population_df.loc[samples, :]
+
+    if features == "infer":
+        features = infer_cp_features(population_df)
+
+    population_df = population_df.loc[:, features]
+
+    return population_df
+
+df = pd.read_csv('C:/Users/Ruifan/neuronal-cell-painting/1.run-workflows/profiles/NCP_STEM_1/BR_NCP_STEM_1/BR_NCP_STEM_1_normalized.csv.gz')
+print(noise_removal(df))
--- a/pycytominer/operations/noise_removal.py
+++ b/pycytominer/operations/noise_removal.py
@@ -6,20 +6,10 @@ import numpy as np
 import pandas as pd
 from pycytominer.cyto_utils import infer_cp_features
 
+
 def noise_removal(
-    population_df, features="infer", samples="all", stdev_cutoff=0.8,
+        population_df, perturb_list, features="infer", samples="all", stdev_cutoff=0.8,
 ):
-    """
-    Removes features which are insufficiently reproducible within the same perturbation group.
-    Args:
-        population_df:
-        features:
-        samples:
-        stdev_cutoff:
-
-    Returns:
-
-    """
     # Subset dataframe
     if samples != "all":
         population_df = population_df.loc[samples, :]
@@ -29,7 +19,38 @@ def noise_removal(
 
     population_df = population_df.loc[:, features]
 
-    return population_df
+    # Label each row with the identity of its perturbation group
+    population_df['group'] = perturb_list
 
-df = pd.read_csv('C:/Users/Ruifan/neuronal-cell-painting/1.run-workflows/profiles/NCP_STEM_1/BR_NCP_STEM_1/BR_NCP_STEM_1_normalized.csv.gz')
-print(noise_removal(df))
+    # List of features to remove
+    features_to_remove = []
+
+    # Loop through every feature, and find the mean stdev for each feature within a perturbation group. Remove if
+    # it's too high.
+    for feature in features:
+        feature_stdevs = []  # Feature stdev for each perturbation group
+
+        # Populate feature_stdevs with the standard deviations of the feature for every perturbation group
+        for perturb in perturb_list:
+            temp = population_df[population_df['group'] == perturb][feature]
+            temp_values = list(temp.values)
+            feature_stdevs.append(np.std(temp_values))
+
+        # Find mean feature standard deviation. If it's too high, add to the exclude list.
+        feature_mean_stdev = np.mean(feature_stdevs)
+        if feature_mean_stdev > stdev_cutoff:
+            features_to_remove.append(feature)
+
+    return features_to_remove
+
+
+# df = pd.read_csv(
+#     'C:/Users/Ruifan/neuronal-cell-painting/1.run-workflows/profiles/NCP_STEM_1/BR_NCP_STEM_1/BR_NCP_STEM_1_normalized.csv.gz',
+#     index_col=0)
+# df = df.dropna(axis=1, how='any')
+# groups = pd.read_csv(
+#     'C:/Users/Ruifan/neuronal-cell-painting/1.run-workflows/profiles/NCP_STEM_1/BR_NCP_STEM_1/BR_NCP_STEM_1_groups.csv',
+#     index_col=0)['group'].tolist()
+#
+# print(groups)
+# print(noise_removal(df, groups))
--- a/pycytominer/feature_select.py
+++ b/pycytominer/feature_select.py
@@ -17,6 +17,7 @@ from pycytominer.cyto_utils import (
     infer_cp_features,
     drop_outlier_features,
 )
+from pycytominer.operations.noise_removal import noise_removal
 
 
 def feature_select(
@@ -34,6 +35,8 @@ def feature_select(
     float_format=None,
     blocklist_file=None,
     outlier_cutoff=15,
+    perturb_list=None,
+    stdev_cutoff=None,
 ):
     """
     Performs feature selection based on the given operation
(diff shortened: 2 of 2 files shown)


ruifanp:
This adds the noise removal feature selection. Prior to using this step, the data must be normalized first. The operation works by repeating the following calculation on each feature:
1. For each unique perturbation group, find the standard deviation for the feature.
2. Find the weighted mean (based on number of replicates) of the standard deviations.
3. If the weighted mean is higher than the cutoff, the feature is discarded.


image

ruifanp:
Example of this in action:

df = pd.DataFrame(
    {
        "x": [1, 3, 8, 5, 2, 2],
        "y": [1, 2, 8, 5, 2, 1],
        "z": [9, 3, 8, 9, 2, 9],
        "zz": [0, -3, 8, 9, 6, 9],
    }
).reset_index(drop=True)

# Suppose the first 3 are replicates of perturbation a and the last 3 are of perturbation b
perturb_list = ['a','a', 'a', 'b', 'b', 'b']
# Run feature selection
feature_select(df, features=df.columns.tolist(), operation='noise_removal', perturb_list=perturb_list, stdev_cutoff=3)

Note that features can be inferred, but you must always provide the perturbation list.

niranjchandrasekaran (changes requested):
@ruifanp I am looking forward to using this approach for the JUMP datasets. I have added a few suggestions. I hope they make sense and are useful to you.

I am also tagging @gwaygenomics to take a look at this PR. His reviews are legendary 😄  and one can learn a lot from his suggestions.

gwaygenomics:
@ruifanp - amazing to have this contribution! I will provide my review once you address @niranjchandrasekaran 's comments.

BTW, in the future, I don't think that both of us will need to review. Certainly supplemental comments can be helpful, but our CONTRIBUTING document only specifies one required maintainer acceptance:

> All pull requests must be reviewed and approved by at least one project maintainer in order to be merged. 

From https://github.com/cytomining/pycytominer/blob/372f1086318841670a916f20334333b06e6b84c9/CONTRIBUTING.md#pull-requests

gwaygenomics:
I'll run the tests now, to see if anything obvious needs fixing

pycytominer/operations/noise_removal.py
```diff
@@ -0,0 +1,63 @@
...
+    if samples != "all":
+        population_df = population_df.loc[samples, :]
+
+    if features == "infer":
+        features = infer_cp_features(population_df)
+
+    population_df = population_df.loc[:, features]
+
+    # Label each row with the identity of its perturbation group
+    population_df['group'] = perturb_list
```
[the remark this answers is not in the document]
ruifanp: Maybe. I was going for allowing the user to set custom perturbation groupings (for example, count each gene as a unique perturb, or each gene-guide combo. I agree that having an inferred default from some kind of metadata column makes sense.

pycytominer/operations/noise_removal.py
```diff
@@ -0,0 +1,63 @@
...
+        for perturb in perturb_list:
+            temp = population_df[population_df['group'] == perturb][feature]
+            temp_values = list(temp.values)
+            # Check to make sure there are enough data points to calculate stdev
+            assert (~np.isnan(
+                temp_values)).sum() >= 2, "Feature {} for perturbation group {} does not have sufficient non-NaN " \
+                                          "values. Require at least 2 to calculate standard deviation".\
+                format(feature, perturb)
+
+            feature_stdevs.append(np.std(temp_values))
```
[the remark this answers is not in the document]
ruifanp: Yes, you're right, I changed it and can confirm they give the exact same results. The one drawback is that with this method is that all groups are weighted equally (a perturbation with 8 instances gets weighed the same as with 4) whereas in my previous method they would be weighed proportionally. However, the time savings from this change is so massive that it's worth it.

pycytominer/operations/noise_removal.py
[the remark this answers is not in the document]
gwaygenomics: I'm about to dig into this review @ruifanp, but I am curious as to how you're thinking about @niranjchandrasekaran 's suggestion here. I like having the option, but I do agree that it's the simplest thing to use an existing metadata column.

Ok, working on my review now! 🤓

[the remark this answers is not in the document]
gwaygenomics: you can support both options @ruifanp 

```python
if type(perturb) == "str":
    assert perturb in population_df.columns, f"{perturb} not found. Are you sure it is a metadata column?"
    perturb_update = population_df.loc[:, perturb]
else:
   assert len(perturb) == population_df.shape[0], f"The length of input list: {len(perturb)} is not equivalent to your data: {population_df.shape[0]}"
   perturb_update = perturb.copy()

population_df = population_df.assign(group=perturb_update)
```

Note four things:

1. I changed the name of `perturb_list` to a placeholder - if you go this route, which I strongly advise (it's so much easier for the profiling recipe), then you should not indicate that the variable is a list, if it can also be a string (but that's obvious 😄 ) 
2. You will need to reorder when you're subsetting features in the population_df to be after this step
3. If the user inputs a list, it would be smart to assert that the list is the same length as rows in the population_df. This is especially important if the user also sets `samples` to be something other than "all".
4. I did not test this code - it likely needs a fix or two.

pycytominer/operations/noise_removal.py
```diff
@@ -0,0 +1,52 @@
...
+        features = infer_cp_features(population_df)
+
+    population_df = population_df.loc[:, features]
+
+    # Label each row with the identity of its perturbation group
+    population_df['group'] = perturb_list
+
+    # Get the standard deviations of features within each group
+    stdev_df = population_df.groupby('group').apply(lambda x: np.std(x))
+    stdev_means_df = stdev_df.mean()
```
gwaygenomics, line 49:
```suggestion
    stdev_means_df = population_df.groupby('group').apply(lambda x: np.std(x)).mean()
```

Gotta love pandas chaining! keep it up

pycytominer/operations/noise_removal.py
```diff
@@ -0,0 +1,52 @@
...
+    to_remove = stdev_means_df[stdev_means_df > stdev_cutoff].index.tolist()
```
gwaygenomics, line 50:
```suggestion
    # Identify noisy features that have greater standard deviation within group than the threshold
    to_remove = stdev_means_df[stdev_means_df > stdev_cutoff].index.tolist()
```

Just to help someone reading this. Is the text accurate?

pycytominer/tests/test_feature_select.py
```diff
@@ -73,6 +71,41 @@
 ).reset_index(drop=True)
 
 
+def test_feature_select_noise_removal():
```
gwaygenomics, line 74: make sure to run `black`

pycytominer/tests/test_feature_select.py
```diff
@@ -73,6 +71,41 @@
...
+    """
+    Testing noise_removal feature selection operation
+    """
+    # Set perturbation groups for the test dataframes
```
gwaygenomics, line 78: you will also need to write a separate test for when the user specifies a metadata column

pycytominer/tests/test_feature_select.py
```diff
@@ -73,6 +71,41 @@
...
+    expected_result3 = data_df[['x', 'y', 'z', 'zz']]
+    pd.testing.assert_frame_equal(result1, expected_result1)
+    pd.testing.assert_frame_equal(result2, expected_result2)
+    pd.testing.assert_frame_equal(result3, expected_result3)
+
+    # Test on data_unique_test_df
+    data_unique_test_df_groups = []
+    for elem in ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']:
+        data_unique_test_df_groups.append([elem] * 10)
+    data_unique_test_df_groups = [item for sublist in data_unique_test_df_groups for item in sublist]
```
gwaygenomics, line 99: woah, this is hard to read. What's going on here?

gwaygenomics: What specifically are you testing?

gwaygenomics (changes requested):
Thanks again for this contribution @ruifanp - it's especially awesome considering JUMP is likely to benefit from it immediately.

My two main comments are:

1. An enhancement to allow a user to also specify a metadata column
2. A clarification on what you're testing for with `data_unique_test_df` - is this different than what you're testing for earlier in that function? If it is different, then I think you need to at the very least add comments on what each line is doing - very hard to read!

pycytominer/tests/test_feature_select.py
ruifanp: what do you mean?

gwaygenomics (commented):
One required simplification, one question, one typo fix.

Very close!

pycytominer/tests/test_feature_select.py
```diff
@@ -73,6 +71,96 @@
...
+        operation="noise_removal",
+        noise_removal_perturb_groups="perturb_group",
+        noise_removal_stdev_cutoff=500,
+    )
+    expected_result4b = data_unique_test_df2[["a", "b", "perturb_group"]]
+    expected_result5b = data_unique_test_df2[["a", "b", "c", "d", "perturb_group"]]
+    pd.testing.assert_frame_equal(result4b, expected_result4b)
+    pd.testing.assert_frame_equal(result5b, expected_result5b)
+
+    # Test assertions based on type of perturbation groups passed
```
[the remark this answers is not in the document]
ruifanp: 4a and 5b are the same tests as 4 and 5, but I pass in a string for the metadata column instead of passing in metadata directly.

I should have removed that comment. My original intention was to write tests to check for assertion failure like when you pass in invalid perturbation, but I noticed the other tests didn't test for those nor was "import unittest" used (seems like using the unittest framework is the only way to test assertion errors). I think it's a good ideal to test for these anyways especially in light of your comment above. Can I import unittests for this?

niranjchandrasekaran (approved):
Looks great @ruifanp!

Issue #152, opened 2021-07-01
Code refactor to limit memory usage?


sjfleming:
Hi all,

I've been working with Carmen @carmendv from the Precision Cardiology Lab on using `pycytominer` to aggregate and normalize data which has been saved in an SQLite database format by running
```python
cytominer-database ingest /data/directory sqlite:///backend.sqlite -c ingest_config.ini
```

Our strategy once the database has been created is then to run the following code snippet using `pycytominer`
```python
from pycytominer.cyto_utils.cells import SingleCells

sc = SingleCells('sqlite:///backend.sqlite',
                 aggregation_operation='median')
sc.aggregate_profiles().to_csv(output_filename)
```

However, as we've discussed with @niranjchandrasekaran and @gwaygenomics , the first step of this operation seems to be to load the entire SQLite database into memory.  It seems like this happens either here
https://github.com/cytomining/pycytominer/blob/c0d3e86aa64de8b1c6c3213d48937aab8e9d1c1d/pycytominer/cyto_utils/cells.py#L147
or here (if `load_image_data=False` upon object instantiation)
https://github.com/cytomining/pycytominer/blob/c0d3e86aa64de8b1c6c3213d48937aab8e9d1c1d/pycytominer/cyto_utils/cells.py#L430

These database files can be quite large, many tens of GB or even hundreds of GB.  At some point it becomes untenable to have a machine with enough memory to handle this, as well as prohibitively expensive.  This seems to be related to this issue #142 

As I understand it, the current strategy is:

1. create the SQLite database
2. load the entire database into memory as a pandas dataframe
3. use pandas to manipulate the dataframe and perform aggregation

An alternative strategy I am proposing would be:

1. create the SQLite database
2. perform slightly more complex queries on the database on disk in order to perform the aggregation directly via SQL

This would have the advantage of truly making use of the database, and could really minimize the memory footprint.

I was thinking of something like this (https://www.sqlitetutorial.net/sqlite-group-by/) though I am no expert at SQL.

Is there interest in this?  Does this seem like a good direction for `pycytominer`?

I do think it would probably require a decent amount of refactoring, since the `aggregate()` function in aggregate.py takes a pandas dataframe as an input, and a lot of the code currently involves fetching and formatting that dataframe.  My current thinking is that `aggregate()` would take the database connection as an input argument, rather than an in-memory dataframe.

gwaygenomics:
[...]

Issue #142, opened 2021-05-06
Pycytominer fails when running aggregate from large SQLite tables


bethac07:
Reading in one table from an ~25-30GB SQLite file (so that table should have been 8-10GB total), it ran completely through all the 64 GB of memory on my machine, leading to the error below.

Happily, adding a chunksize parameter to `cyto_utils/cells.py/load_compartment` as below seems to avoid the worst of the issues (I did still see spikes into the 30s of GB on htop, not sure if it was during pandas.concat or during the actual aggregation itself).  Reporting here as an issue rather than just making a PR because 
a) not sure if you think there is anything worth investigating, I'd say probably not but your call and 
b) not sure if you want pycytominer to ALWAYS use chunk sizes (and if so, if you want to use this particular chunk size); I can also imagine also doing chunksize as a user-passed option or pycytominer trying to assess database size in some way and then choosing whether and how to chunk.

Working in the `profiling` conda env (python 3.7.1) on an ubuntu 14 machine; pip freeze at the bottom.

```
        dflist=[]
        for chunk in pd.read_sql(sql=compartment_query, con=self.conn,chunksize=10000):
            dflist.append(chunk)
        df = pd.concat(dflist)
```

```
Traceback (most recent call last):
  File "profiling-recipe/profiles/profiling_pipeline.py", line 28, in <module>
    process_profile(batch=batch, plate=plate, pipeline=pipeline)
  File "/home/ubuntu/ebs_tmp/work/projects/workflow_demo/workspace/software/workflow_demo_data/profiling-recipe/profiles/profile.py", line 124, in process_profile
    aggregate_args=aggregate_args,
  File "/home/ubuntu/miniconda3/envs/profiling/lib/python3.7/site-packages/pycytominer/cyto_utils/cells.py", line 661, in aggregate_profiles
    compute_counts=True,
  File "/home/ubuntu/miniconda3/envs/profiling/lib/python3.7/site-packages/pycytominer/cyto_utils/cells.py", line 433, in aggregate_compartment
    self.load_compartment(compartment=compartment),
  File "/home/ubuntu/miniconda3/envs/profiling/lib/python3.7/site-packages/pycytominer/cyto_utils/cells.py", line 392, in load_compartment
    df = pd.read_sql(sql=compartment_query, con=self.conn)
  File "/home/ubuntu/miniconda3/envs/profiling/lib/python3.7/site-packages/pandas/io/sql.py", line 516, in read_sql
    chunksize=chunksize,
[...]

PR #156, opened 2021-07-14
`cells.aggregate_compartment()` queries the database in chunks, reducing memory

sjfleming:
# Description

These changes were made directly in response to #152 , and may impact #142 .

The intent is to limit memory usage during `pycytominer.cyto_utils.cells.aggregate_compartment()` calls, and other functions which call this method, including `pycytominer.cyto_utils.cells.aggregate_profiles()`.  One explicit stipulation was to touch `pycytominer.aggregate.aggregate()` as little as possible.  (I did have to make one tiny change.)

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ If you have used pycytominer in your project, please cite us as:
 
 ```
 @software{pycytominer,
-  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Fleming, S.J., Carpenter, A.E., Singh, S.},
+  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Fleming, S.J., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Carpenter, A.E., Singh, S.},
   title = {Pycytominer: Data processing functions for profiling perturbations},
   url = {https://github.com/cytomining/pycytominer},
   version = {0.1},
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ If you have used pycytominer in your project, please cite us as:
 
 ```
 @software{pycytominer,
-  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Carpenter, A.E., Singh, S.},
+  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Fleming, S.J., Carpenter, A.E., Singh, S.},
   title = {Pycytominer: Data processing functions for profiling perturbations},
   url = {https://github.com/cytomining/pycytominer},
   version = {0.1},
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -62,7 +62,7 @@ def aggregate(
     # Subset the data to specified samples
     if isinstance(subset_data_df, pd.DataFrame):
         population_df = subset_data_df.merge(
-            population_df, how="left", on=subset_data_df.columns.tolist()
+            population_df, how="inner", on=subset_data_df.columns.tolist()
         ).reindex(population_df.columns, axis="columns")
 
     # Subset dataframe to only specified variables if provided
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -583,9 +583,6 @@ class SingleCells(object):
                     [3]         | [1, 2]
                     [4]         | [1]
 
-                dtypes:
-                    {'TableNumber': 'integer', 'ImageNumber': 'integer', 'thing3': 'integer'}
-
                 _sqlite_strata_conditions(df,
                                           dtypes={'TableNumber': 'integer', 'ImageNumber': 'integer'},
                                           n=1):
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -1,6 +1,5 @@
 import numpy as np
 import pandas as pd
-import psutil
 from sqlalchemy import create_engine
 from pycytominer import aggregate, normalize
 from pycytominer.cyto_utils import (
@@ -402,7 +401,7 @@ class SingleCells(object):
         compartment,
         compute_subsample=False,
         compute_counts=False,
-        n_strata=None,
+        n_strata=1,
     ):
         """Aggregate morphological profiles. Uses pycytominer.aggregate()
 
@@ -415,9 +414,9 @@ class SingleCells(object):
         compute_counts : bool, default False
             Whether or not to compute the number of objects in each compartment
             and the number of fields of view per well.
-        n_strata : Optional[int], default None
+        n_strata : int, default 1
             Number of unique strata to pull from the database into working memory
-            at once.  Default of None will attempt to use available memory efficiently.
+            at once.  A larger number does not lead to faster compute times
 
         Returns
         -------
@@ -496,7 +495,7 @@ class SingleCells(object):
     def _compartment_df_generator(
         self,
         compartment,
-        n_strata=None,
+        n_strata=1,
     ):
         """A generator function that returns chunks of the entire compartment
         table from disk. Attempts to scale SQLite queries to use most of the
@@ -509,9 +508,9 @@ class SingleCells(object):
         ----------
         compartment : str
             Compartment to aggregate.
-        n_strata : Optional[int], default None
-            Number of unique strata to query from the database in one iteration.
-            Input value None will attempt to use the available memory efficiently.
+        n_strata : int, default 1
+            Number of unique strata to pull from the database into working memory
+            at once.  A larger number does not lead to faster compute times
 
         Yields
         ------
@@ -521,6 +520,10 @@ class SingleCells(object):
             between chunks, and thus groupby aggregations are valid
 
         """
+
+        assert (
+            n_strata > 0
+        ), "Number of strata to pull into memory at once (n_strata) must be > 0"
 
         # Columns of the compartment table we will obtain
         cols = "*"
@@ -623,20 +626,6 @@ class SingleCells(object):
             df=df_unique_mergecols.head(1),
             dtypes=dtype_dict,
         )[0]
-
-        if n_strata is None:
-
-            # Grab first unique stratum and see how large the dataframe is
-            single_stratum_query = (
-                f"select {cols} from {compartment} where {first_stratum_condition}"
-            )
-            df_stratum1 = pd.read_sql(sql=single_stratum_query, con=self.conn)
-
-            # Compute a reasonable number of unique strata to grab at once
-            avail_mem = psutil.virtual_memory().available
-            single_stratum_mem = df_stratum1.memory_usage(deep=True).sum()
-            approx_n = int(np.floor(0.75 * avail_mem / single_stratum_mem).item())
-            n_strata = min(len(df_unique_mergecols), max(1, approx_n))
 
         # Group the unique strata values into a list of SQLite condition strings
         # Find unique aggregated strata for the output
@@ -802,7 +791,7 @@ class SingleCells(object):
         output_file="none",
         compression_options=None,
         float_format=None,
-        n_strata=None,
+        n_strata=1,
     ):
         """Aggregate and merge compartments. This is the primary entry to this class.
 
@@ -817,9 +806,9 @@ class SingleCells(object):
             Compression arguments as input to pandas.to_csv() with pandas version >= 1.2.
         float_format : str, optional
             Decimal precision to use in writing output file.
-        n_strata : Optional[int], default None
+        n_strata : int, default 1
             Number of unique strata to pull from the database into working memory
-            at once.  Default of None will attempt to use available memory efficiently.
+            at once.  A larger number does not lead to faster compute times
 
         Returns
         -------
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -401,7 +401,7 @@ class SingleCells(object):
         compartment,
         compute_subsample=False,
         compute_counts=False,
-        n_strata=1,
+        n_aggregation_memory_strata=1,
     ):
         """Aggregate morphological profiles. Uses pycytominer.aggregate()
 
@@ -414,9 +414,9 @@ class SingleCells(object):
         compute_counts : bool, default False
             Whether or not to compute the number of objects in each compartment
             and the number of fields of view per well.
-        n_strata : int, default 1
+        n_aggregation_memory_strata : int, default 1
             Number of unique strata to pull from the database into working memory
-            at once.  A larger number does not lead to faster compute times
+            at once.  Typically 1 is fastest.  A larger number uses more memory.
 
         Returns
         -------
@@ -438,7 +438,8 @@ class SingleCells(object):
         # Iteratively call aggregate() on chunks of the full compartment table
         object_dfs = []
         for compartment_df in self._compartment_df_generator(
-            compartment=compartment, n_strata=n_strata
+            compartment=compartment,
+            n_aggregation_memory_strata=n_aggregation_memory_strata,
         ):
 
             population_df = self.image_df.merge(
@@ -495,7 +496,7 @@ class SingleCells(object):
     def _compartment_df_generator(
         self,
         compartment,
-        n_strata=1,
+        n_aggregation_memory_strata=1,
     ):
         """A generator function that returns chunks of the entire compartment
         table from disk.
@@ -507,13 +508,13 @@ class SingleCells(object):
         ----------
         compartment : str
             Compartment to aggregate.
-        n_strata : int, default 1
+        n_aggregation_memory_strata : int, default 1
             Number of unique strata to pull from the database into working memory
-            at once.  A larger number does not lead to faster compute times
+            at once.  Typically 1 is fastest.  A larger number uses more memory.
 
-        Yields
-        ------
-        image_df : pandas.core.frame.DataFrame
+        Returns
+        -------
+        image_df : Iterator[pandas.core.frame.DataFrame]
             A generator whose __next__() call returns a chunk of the compartment
             table, where rows comprising a unique aggregation stratum are not split
             between chunks, and thus groupby aggregations are valid
@@ -521,10 +522,10 @@ class SingleCells(object):
         """
 
         assert (
-            n_strata > 0
-        ), "Number of strata to pull into memory at once (n_strata) must be > 0"
+            n_aggregation_memory_strata > 0
+        ), "Number of strata to pull into memory at once (n_aggregation_memory_strata) must be > 0"
 
-        # Columns of the compartment table we will obtain
+        # Obtain data types of all columns of the compartment table
         cols = "*"
         compartment_row1 = pd.read_sql(
             sql=f"select {cols} from {compartment} limit 1",
@@ -550,85 +551,12 @@ class SingleCells(object):
             .reset_index(drop=True)
         )
 
-        def _sqlite_strata_conditions(df, dtypes, n=1):
-            """Given a dataframe with columns as strata and rows as unique value
-            combinations, return a list of strings which constitute valid SQLite
-            conditional statements.
-
-            Parameters
-            ----------
-            df : pandas.core.frame.DataFrame
-                A dataframe with columns as strata and rows as unique value combos
-
-            dtypes : Dict[str, str]
-                Dictionary to look up SQLite datatype based on column name
-
-            n : int
-                Number of rows of the input df to combine in each conditional
-                statement
-
-            Returns
-            -------
-            grouped_conditions : List[str]
-                A list of strings, each string being a valid SQLite conditional
-
-            Example
-            -------
-
-                df:
-                    TableNumber | ImageNumber
-                    -------------------------
-                    [1]         | [1]
-                    [2]         | [1, 2, 3]
-                    [3]         | [1, 2]
-                    [4]         | [1]
-
-                _sqlite_strata_conditions(df,
-                                          dtypes={'TableNumber': 'integer', 'ImageNumber': 'integer'},
-                                          n=1):
-                    ['(TableNumber in (1) and ImageNumber in (1))',
-                     '(TableNumber in (2) and ImageNumber in (1, 2, 3))',
-                     '(TableNumber in (3) and ImageNumber in (1, 2))',
-                     '(TableNumber in (4) and ImageNumber in (1))']
-
-                _sqlite_strata_conditions(df,
-                                          dtypes={'TableNumber': 'text', 'ImageNumber': 'integer'},
-                                          n=2):
-                    ["(TableNumber in ('1') and ImageNumber in (1))
-                      or (TableNumber in ('2') and ImageNumber in (1, 2, 3))",
-                     "(TableNumber in ('3') and ImageNumber in (1, 2))
-                      or (TableNumber in ('4') and ImageNumber in (1))"]
-            """
-            conditions = []
-            for row in df.iterrows():
-                series = row[1]
-                values = [
-                    [f"'{a}'" for a in y] if dtypes[x] == "text" else y
-                    for x, y in zip(series.index, series.values)
-                ]  # put quotes around text entries
-                condition_list = [
-                    f"{x} in ({', '.join([str(a) for a in y]) if len(y) > 1 else y[0]})"
-                    for x, y in zip(series.index, values)
-                ]
-                conditions.append(f"({' and '.join(condition_list)})")
-            grouped_conditions = [
-                " or ".join(conditions[i : (i + n)])
-                for i in range(0, len(conditions), n)
-            ]
-            return grouped_conditions
-
-        # Translate the first strata combination into a SQLite condition string
-        first_stratum_condition = _sqlite_strata_conditions(
-            df=df_unique_mergecols.head(1),
-            dtypes=dtype_dict,
-        )[0]
-
         # Group the unique strata values into a list of SQLite condition strings
         # Find unique aggregated strata for the output
         strata_conditions = _sqlite_strata_conditions(
             df=df_unique_mergecols,
             dtypes=dtype_dict,
-            n=n_strata,
+            n=n_aggregation_memory_strata,
         )
 
         # The generator, for each group of compartment values
@@ -787,7 +715,7 @@ class SingleCells(object):
         output_file="none",
         compression_options=None,
         float_format=None,
-        n_strata=1,
+        n_aggregation_memory_strata=1,
     ):
         """Aggregate and merge compartments. This is the primary entry to this class.
 
@@ -802,9 +730,9 @@ class SingleCells(object):
             Compression arguments as input to pandas.to_csv() with pandas version >= 1.2.
         float_format : str, optional
             Decimal precision to use in writing output file.
-        n_strata : int, default 1
+        n_aggregation_memory_strata : int, default 1
             Number of unique strata to pull from the database into working memory
-            at once.  A larger number does not lead to faster compute times
+            at once.  Typically 1 is fastest.  A larger number uses more memory.
 
         Returns
         -------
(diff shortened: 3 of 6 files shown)


sjfleming:
I performed some tests using a 37GB SQLite database file.  I realized that, when it comes to memory usage, more is _not_ better.  I had expected that using more of the available memory would lead to faster aggregation, but this is in fact the opposite of what happens.  I suppose the whole task is I/O bound, and the actual aggregation in pandas is very very fast.  So putting more in memory at once does not help.

Test cases were carried out by specifying `n_strata`, a parameter which I have added to `SingleCells.aggregate_profiles()`.  `n_strata` controls how many strata (of the unique specified aggregation strata) are pulled from the database into memory at once.  Tests were performed on a 2cpu 8GB memory machine with a solid-state hard drive, running Ubuntu 16.04.

Results of tests --- runtime in (minutes:seconds)
```
n_strata = 1 ------------- 18:21
n_strata = 2 ------------- 19:43
n_strata = 4 ------------- 25:31
n_strata max ------------- 26:45
```

Based on these results, I am going to remove the auto-scaling of `n_strata` to be the max it can be to fit in memory, and I will also remove the related dependency I introduced (`psutil`).  I will set the default `n_strata` to be 1.

sjfleming:
I have checked the final outputs of aggregation on a 37GB test database: 
- run with the previous code
- run with this branch

and they seem to be identical.

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+        """
+
+        assert (
+            n_strata > 0
+        ), "Number of strata to pull into memory at once (n_strata) must be > 0"
+
+        # Columns of the compartment table we will obtain
+        cols = "*"
+        compartment_row1 = pd.read_sql(
+            sql=f"select {cols} from {compartment} limit 1",
```
gwaygenomics, line 530: what does `limit 1` mean? it will pull only the first row?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+            con=self.conn,
+        )
+        typeof_str = ", ".join([f"typeof({x})" for x in compartment_row1.columns])
+        compartment_dtypes = pd.read_sql(
+            sql=f"select {typeof_str} from {compartment} limit 1",
+            con=self.conn,
+        )
+        dtype_dict = dict(
+            zip(
+                [s[7:-1] for s in compartment_dtypes.columns],  # strip typeof( )
```
gwaygenomics, line 540: hmm, i don't know what is going on here. You're pulling in only columns of a certain dtype, finding their names, and then creating a dictionary such that you do not need to infer dtypes in subsequent `pandas.read_sql()` calls?

Why do you need to subset [7:-1]? I am always weary of these hardcoded calls!

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+                compartment_dtypes.iloc[0].values,
+            )
+        )
+
+        # Obtain all valid strata combinations, and their merge_cols values
+        df_unique_mergecols = (
+            self.image_df[self.strata + self.merge_cols]
```
gwaygenomics, line 547: @niranjchandrasekaran - when you review, can you pay slightly more attention to this line? I think that it is ok, but will `self.strata` and `self.merge_cols` always be in `self.image_df`? Will the two always be unique lists?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+            .groupby(self.strata)
+            .agg(lambda s: np.unique(s).tolist())
+            .reset_index(drop=True)
+        )
+
+        def _sqlite_strata_conditions(df, dtypes, n=1):
+            """Given a dataframe with columns as strata and rows as unique value
```
gwaygenomics, line 554: not sure what `columns as strata` means

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+            combinations, return a list of strings which constitute valid SQLite
+            conditional statements.
+
+            Parameters
+            ----------
+            df : pandas.core.frame.DataFrame
+                A dataframe with columns as strata and rows as unique value combos
+
```
gwaygenomics, line 562:
```suggestion
```

I think you can remove empty new lines in docstrings

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+            dtypes : Dict[str, str]
+                Dictionary to look up SQLite datatype based on column name
+
```
gwaygenomics, line 565:
```suggestion
```

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+            n : int
+                Number of rows of the input df to combine in each conditional
```
gwaygenomics, line 567: Hmm, i don't understand this description

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+                    -------------------------
+                    [1]         | [1]
+                    [2]         | [1, 2, 3]
+                    [3]         | [1, 2]
+                    [4]         | [1]
+
+                _sqlite_strata_conditions(df,
+                                          dtypes={'TableNumber': 'integer', 'ImageNumber': 'integer'},
+                                          n=1):
+                    ['(TableNumber in (1) and ImageNumber in (1))',
```
gwaygenomics, line 589: is this standard convention for an example? I think I'm supposed to read `_sqlite_strata_conditions(df,
                                          dtypes={'TableNumber': 'integer', 'ImageNumber': 'integer'},
                                          n=1)` as the function call and then the items right of `:` as the output. Is that correct? Is there a more intuitive way of showing what the example output might be?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+        )
+
+        def _sqlite_strata_conditions(df, dtypes, n=1):
+            """Given a dataframe with columns as strata and rows as unique value
+            combinations, return a list of strings which constitute valid SQLite
+            conditional statements.
+
+            Parameters
+            ----------
+            df : pandas.core.frame.DataFrame
```
gwaygenomics, line 560: Is this a compartment data frame? Is this merged with image_df? Can you be more specific in the docstring about what columns to expect in this dataframe?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+                conditions.append(f"({' and '.join(condition_list)})")
+            grouped_conditions = [
+                " or ".join(conditions[i : (i + n)])
+                for i in range(0, len(conditions), n)
+            ]
+            return grouped_conditions
+
+        # Translate the first strata combination into a SQLite condition string
+        first_stratum_condition = _sqlite_strata_conditions(
+            df=df_unique_mergecols.head(1),
```
gwaygenomics, line 622: it took me a while to figure out that `df_unique_mergecols` is defined out of scope of this function. Should you include as a function argument in `_sqlite_strata_conditions`?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+                ]
+                conditions.append(f"({' and '.join(condition_list)})")
+            grouped_conditions = [
+                " or ".join(conditions[i : (i + n)])
+                for i in range(0, len(conditions), n)
+            ]
+            return grouped_conditions
+
+        # Translate the first strata combination into a SQLite condition string
+        first_stratum_condition = _sqlite_strata_conditions(
```
gwaygenomics, line 621: where are you using `first_stratum_condition`?

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+
+        # Obtain all valid strata combinations, and their merge_cols values
+        df_unique_mergecols = (
+            self.image_df[self.strata + self.merge_cols]
+            .groupby(self.strata)
+            .agg(lambda s: np.unique(s).tolist())
+            .reset_index(drop=True)
+        )
+
+        def _sqlite_strata_conditions(df, dtypes, n=1):
```
gwaygenomics, line 553: Based on my read, this function is the most critical contribution in this PR (cc @niranjchandrasekaran). It's probably because I'm not an expert in generator functions and how they're used, but I am having some trouble following the logic.

I make a couple comments in this function to try to help my understanding, but if you can see where my knowledge gap is, please let me know. Right now, I think this function will be difficult to maintain unless we provide a couple more tests.

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -490,6 +490,41 @@ def test_aggregate_subsampling_profile_compress():
...
 
 
+def test_aggregate_subsampling_profile_compress_multiple_queries():
+    compress_file = os.path.join(tmpdir, "test_aggregate_compress.csv.gz")
+
+    _ = ap_subsample.aggregate_profiles(
+        output_file=compress_file,
+        compute_subsample=True,
+        compression_options={"method": "gzip"},
+        n_strata=1,  # this will force multiple queries from each compartment
```
gwaygenomics, line 500: can you run this test multiple times looping through different `n_strata`:

E.g.

```python
for n_strata in [1, 2, 3]:
    ...
   aggregate_profiles(n_strata=n_strata)
```

and confirm that the expected result doesn't change

gwaygenomics: can you also add a separate test where you set `n_strata` to an invalid argument?

gwaygenomics (changes requested):
Really an important contribution @sjfleming - thank you! I have made several suggestions and added many points of discussion that we'll need to talk through before merging.

@niranjchandrasekaran is also planning on talking a look at the whole PR, and I've tagged him at two specific places where I think an extra careful eye is needed.

My main comment is that I'd really like to understand more about `_sqlite_strata_conditions()` and test it _slightly_ more thoroughly. I'd also like to modify some variable names and docstrings, and I have a couple of one-off suggestions/comments.

I think we are close to merging!

pycytominer/aggregate.py
```diff
@@ -62,7 +62,7 @@ def aggregate(
     # Subset the data to specified samples
     if isinstance(subset_data_df, pd.DataFrame):
         population_df = subset_data_df.merge(
-            population_df, how="left", on=subset_data_df.columns.tolist()
+            population_df, how="inner", on=subset_data_df.columns.tolist()
```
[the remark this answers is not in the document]
sjfleming: Great question!  Without this change, my chunked aggregation strategy fails for the following reason: for each chunk, I call `pycytominer.aggregate()`, and I create one aggregated data frame per chunk, which I then concatenate at the end into the final output.  If the aggregation per chunk occurs with a "left" merge, then all the rows of `subset_data_df` that are not part of this chunk will get `NaN` entries.  For example:

```
# using "left" merge for a chunk containing only well A01
well             image              value
A01             1                       1.543
A01             2                      2.442
A02             1                      NaN

# using "inner" merge for a chunk containing only well A01
well             image              value
A01             1                       1.543
A01             2                      2.442
```
and then after the final concatenation step
```
# using "left" merge
well             image              value
A01             1                       1.543
A01             2                      2.442
A02             1                      NaN
A02             1                      0.234

# using "inner" merge
well             image              value
A01             1                       1.543
A01             2                      2.442
A01             2                      2.442
```

pycytominer/cyto_utils/cells.py
```diff
@@ -401,6 +401,7 @@ def aggregate_compartment(
         compartment,
         compute_subsample=False,
         compute_counts=False,
+        n_strata=1,
```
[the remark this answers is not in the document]
sjfleming: Yes, I'm happy to re-think this name.  I also wasn't very happy with it.

pycytominer/cyto_utils/cells.py
```diff
@@ -411,7 +412,11 @@ def aggregate_compartment(
         compute_subsample : bool, default False
             Whether or not to subsample.
         compute_counts : bool, default False
-            Whether or not to compute the number of objects in each compartment and the number of fields of view per well.
+            Whether or not to compute the number of objects in each compartment
+            and the number of fields of view per well.
+        n_strata : int, default 1
+            Number of unique strata to pull from the database into working memory
```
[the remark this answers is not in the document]
sjfleming: What it really is, is the number of unique aggregation strata being pulled into one chunk.  For example, if we are aggregating by "well", then `n_strata=1` means that one "well" will be pulled from the SQLite database into memory at a time.

pycytominer/cyto_utils/cells.py
```diff
@@ -411,7 +412,11 @@ def aggregate_compartment(
...
+            at once.  A larger number does not lead to faster compute times
```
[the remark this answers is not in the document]
sjfleming: According to my (limited) experiments, a larger number actually leads to slower compute times (see my comment in the PR above).  It seems the best thing to do is probably always to have `n_strata=1`

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+
+        Parameters
+        ----------
+        compartment : str
+            Compartment to aggregate.
+        n_strata : int, default 1
+            Number of unique strata to pull from the database into working memory
+            at once.  A larger number does not lead to faster compute times
+
+        Yields
```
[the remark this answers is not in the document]
sjfleming: This might actually not work as a sphinx autodoc doctoring title, I didn't check actually.  It looks like the right thing to do might be to use "Returns" with an `Iterator` specified, as in
https://stackoverflow.com/questions/24461529/is-there-a-sphinx-rest-python-docstring-field-for-yields

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+        ------
+        image_df : pandas.core.frame.DataFrame
+            A generator whose __next__() call returns a chunk of the compartment
+            table, where rows comprising a unique aggregation stratum are not split
+            between chunks, and thus groupby aggregations are valid
+
+        """
+
+        assert (
+            n_strata > 0
```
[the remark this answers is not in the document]
sjfleming: The real max value would be the total number of aggregation strata present in the dataset.  For example, if aggregating by "well", then the max possible `n_strata` would be the total number of "well"s in the SQLite.  However, if the user inputs a larger number, the code will still work fine (though this statement is untested).

pycytominer/cyto_utils/cells.py
sjfleming: Yes, that's right

pycytominer/cyto_utils/cells.py
sjfleming: This is doing the following:
1. get the first row of the `compartment` table from the SQLite, so that I know all the column names
2. get the `dtype` of every column in the SQLite (which I have to request by name... that's why I needed the names)

These `dtype`s are needed to construct valid queries (specifically, `TEXT` must include quotations).

The subset `[7:-1]` I agree is ugly... the reason for it is that the columns come back with names like `typeof(feature1)` instead of `feature1`.  I wanted to strip out "typeof(" from the beginning and ")" from the end.  The columns always come back from the SQL request named in exactly this way... so I think the code is robust... but it is hard to understand.

sjfleming: It might be possible to replace
```python
            zip(
                [s[7:-1] for s in compartment_dtypes.columns],  # strip typeof( )
                compartment_dtypes.iloc[0].values,
            )
```
with
```python
            zip(
                [s for s in compartment_row1.columns],
                compartment_dtypes.iloc[0].values,
            )
```
but I am not sure that the columns _always_ come back in the same order as the request.

pycytominer/cyto_utils/cells.py
sjfleming: Great questions.  I was assuming that the two are unique lists.

pycytominer/cyto_utils/cells.py
sjfleming: I meant to say that the input dataframe's rows should represent unique strata, and that the input dataframe's columns should be the elements of `self.merge_cols`, so this is incorrect

pycytominer/cyto_utils/cells.py
sjfleming: If `n=1`, then each row of the input dataframe becomes a single "conditional statement" string in SQL.  If `n=2`, then pairs of rows of the input dataframe are combined into "conditional statement" strings.

pycytominer/cyto_utils/cells.py
sjfleming: You are correct about my intent.  I was not sure what is the standard convention for an example.  Maybe I should make it more standard.

pycytominer/cyto_utils/cells.py
sjfleming: This is a very strange dataframe that is neither of the above.  It is a new dataframe constructed with the sole purpose of figuring out how to divide the `compartment` table into chunks.

The rows represent unique `self.strata` in the data.  The columns are `self.merge_cols`

pycytominer/cyto_utils/cells.py
sjfleming: `df_unique_mergecols` is in scope of `_compartment_df_generator()`, but out of scope of `_sqlite_strata_conditions()`

pycytominer/cyto_utils/cells.py
sjfleming: I am not.  Good catch!  This is an old thing I didn't entirely delete.

pycytominer/cyto_utils/cells.py
sjfleming: I agree that there should be some dedicated tests to demonstrate the function of this function

pycytominer/tests/test_cyto_utils/test_cells.py
sjfleming: Yes this is a good idea in terms of testing multiple values for `n_strata`.

We can test invalid `n_strata` values... but the only truly invalid values should be `< 1`, and there is an explicit assertion for that.  But we can still include a test.

pycytominer/cyto_utils/cells.py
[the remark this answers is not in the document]
sjfleming: "Yields" might work, I'm not sure
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html
I'm not super experienced at this part...

sjfleming:
I have made an attempt to address all your comments and make relevant changes @gwaygenomics .  I added a couple of tests to address a few comments as well.

pycytominer/cyto_utils/cells.py
```diff
@@ -846,3 +775,67 @@ def aggregate_profiles(
             )
         else:
             return aggregated
+
+
+def _sqlite_strata_conditions(df, dtypes, n=1):
```
gwaygenomics, line 780: should this be a separately defined function? When one imports `SingleCells` will it automatically know this function definition? I thought that you had it as a private method of the class before? Any reason you decided to make it separate?

pycytominer/cyto_utils/cells.py
```diff
@@ -846,3 +775,67 @@ def aggregate_profiles(
...
+    """Given a dataframe where columns are merge_cols and rows are unique
+    value combinations that appear as aggregation strata, return a list
+    of strings which constitute valid SQLite conditional statements.
+
+    Parameters
+    ----------
+    df : pandas.core.frame.DataFrame
+        A dataframe where columns are merge_cols and rows represent
+        unique aggregation strata of the compartment table
```
gwaygenomics, line 789:
```suggestion
        the specific image IDs grouped by unique aggregation strata (e.g. Metadata_well) of the image table
```

Is this more accurate?

pycytominer/cyto_utils/cells.py
```diff
@@ -846,3 +775,67 @@ def aggregate_profiles(
...
+    dtypes : Dict[str, str]
+        Dictionary to look up SQLite datatype based on column name
+    n : int
+        Number of rows of the input df to combine in each output
```
gwaygenomics, line 793: this is really great documentation - very appreciated

pycytominer/cyto_utils/cells.py
```diff
@@ -846,3 +775,67 @@ def aggregate_profiles(
...
+        conditional statement. n=1 means each row of the input will
+        correspond to one string in the output list. n=2 means one
+        string in the output list is comprised of two rows from the
+        input df.
+
+    Returns
+    -------
+    grouped_conditions : List[str]
+        A list of strings, each string being a valid SQLite conditional
```
gwaygenomics, line 802: can you add a couple words describing how this is meant to be used?

"being a valid SQLite conditional to facilitate chunked aggregation?" 🤷

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -6,7 +6,7 @@
 from sqlalchemy import create_engine
 
 from pycytominer import aggregate, normalize
-from pycytominer.cyto_utils.cells import SingleCells
+from pycytominer.cyto_utils.cells import SingleCells, _sqlite_strata_conditions
```
gwaygenomics, line 9: ah, I see, you're needing to input this as a new function. I typically think of `_single_leading_underscore` as being private methods, and [they're defined as "internal use" characters](https://pep8.org/#descriptive-naming-styles) in pep8. I'm wondering if I'm missing any critical insight that led you to separate out this method (This is kind of a duplicate comment of one previous)

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -525,6 +525,60 @@ def test_aggregate_subsampling_profile_compress_multiple_queries():
...
+
+    n1_expected_output = [
+        "(TableNumber in (1) and ImageNumber in (1))",
+        "(TableNumber in (2) and ImageNumber in (1, 2, 3))",
+        "(TableNumber in (3) and ImageNumber in (1, 2))",
+        "(TableNumber in (4) and ImageNumber in (1))",
+    ]
+    out1 = _sqlite_strata_conditions(
+        df=df,
+        dtypes={"TableNumber": "integer", "ImageNumber": "integer"},
```
gwaygenomics, line 563: will the `dtype` ever be a string?

gwaygenomics: oh, "text" - I see below, gotcha!

gwaygenomics (commented):
really great to see this happening. I made a couple comments.

@niranjchandrasekaran - are you able to take a look? I think it's worth having your eyes on the full PR, but I did note particular focus areas.

pycytominer/cyto_utils/cells.py
sjfleming: When one imports `SingleCells`, yes, a `SingleCells` object will automatically be able to see that function.  I did have it as a private method of the class before.  I can move it back if we want.  The only reason I moved it out is that I realized it doesn't actually rely on anything from `SingleCells` and it was just a tiny bit easier to import it for testing as a separate function.  Though it could still be imported for testing either way

pycytominer/cyto_utils/cells.py
sjfleming: Haha yeah, I can add a bit.  Really what I mean is that it will return a list of strings (called `conditions`), where each of the strings can be used in an SQL statement like this:
```python
f"select {cols} from {compartment} where {condition}"
```
Here `condition` is one element of the list of strings `conditions`, and the above SQL statement will grab a single chunk for aggregation.  So by iterating over the conditions in `grouped_conditions` and making SQL queries, we get the whole table back in chunks:
```python
for condition in conditions:
    do_sql_query(f"select {cols} from {compartment} where {condition}")
```

pycytominer/tests/test_cyto_utils/test_cells.py
sjfleming: No real insight, no.  Yeah I just did it that way so I could easily import it to test it here.  And the leading underscore should denote to a user that it's private and never to be used directly for anything.

pycytominer/tests/test_cyto_utils/test_cells.py
sjfleming: Yeah, it seems like the `dtype`s for SQL are `null`, `real`, `text`, `integer`, and `blob`.  But the TableNumber and ImageNumber values I think in practice are always of `integer` type, though I am not 100% sure

pycytominer/cyto_utils/cells.py
sjfleming: Yes that's right.  Only I think it turns out to be whatever is in `merge_cols`, so "ImageNumber" and "TableNumber"

gwaygenomics:
a quick note @sjfleming - can you also add your name to the [software citation section](
https://github.com/cytomining/pycytominer/tree/3edd0c766b2f10dea40be96120cae75cf6f886eb#citation) in this PR? Thanks!

sjfleming:
I rebased on top of the current master branch to get the latest changes just now

sjfleming:
Thanks @gwaygenomics , I took a guess about where to put myself in that author list, please do correct me if it's the wrong place :)

README.md
```diff
@@ -80,7 +80,7 @@ If you have used pycytominer in your project, please cite us as:
 
 ```
 @software{pycytominer,
-  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Carpenter, A.E., Singh, S.},
+  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Fleming, S.J., Carpenter, A.E., Singh, S.},
```
gwaygenomics, line 83:
```suggestion
  author = {Way, G.P., Chandrasekaran, S.N., Bornholdt, M., Fleming, S.J., Tsang, H., Adeboye, A., Cimini, B., Weisbart, E., Ryder, P., Stirling, D., Jamali, N., Carpenter, A.E., Singh, S.},
```

Thanks Stephen! I moved you based on lines contributed - a poor metric, I know, but one we have right now. We do have desires to create a more equitable and representative solution, however (discussed in #158 )

sjfleming: Fine by me!  Happy to be included :)

pycytominer/cyto_utils/cells.py
niranjchandrasekaran: I think `self.strata` and `self.merge_cols` will always be in `self.image_df` and they will be unique. I trying to imagine when the two lists will not be unique. We could aggregate at the image-level but for that we will use `Metadata_Site` in `self.strata` and not `ImageNumber`. Other than that, I can't think of an overlap between the two.

pycytominer/cyto_utils/cells.py
```diff
@@ -430,46 +435,210 @@ def aggregate_compartment(
...
+        compartment : str
+            Compartment to aggregate.
+        n_strata : int, default 1
+            Number of unique strata to pull from the database into working memory
+            at once.  A larger number does not lead to faster compute times
+
+        Yields
+        ------
+        image_df : pandas.core.frame.DataFrame
+            A generator whose __next__() call returns a chunk of the compartment
```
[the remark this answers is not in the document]
niranjchandrasekaran: ☝️ I agree with Greg :)

niranjchandrasekaran (commented):
LGTM! I don't any additional comments

pycytominer/cyto_utils/cells.py
sjfleming: I don't know of a way to make a request that doesn't return the information in this way.  One thing I could do to improve readability is to make a little helper function or lambda function whose purpose is clearly and explicitly to remove `typeof()`... that way it's more readable.  For example, replace this

```python
            zip(
                [s[7:-1] for s in compartment_dtypes.columns],  # strip typeof( )
                compartment_dtypes.iloc[0].values,
            )
```
with this
```python
            # strip the characters "typeof(" from the beginning and ")" from the end of 
            # column names returned by SQLite
            strip_typeof = lambda s: s[7:-1]  
            zip(
                [strip_typeof(s) for s in compartment_dtypes.columns],  
                compartment_dtypes.iloc[0].values,
            )
```

pycytominer/cyto_utils/cells.py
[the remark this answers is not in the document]
sjfleming: Added the example

gwaygenomics (approved):
Looks great @sjfleming - I think we can merge this in.

One last comment (sorry I should have mentioned this sooner!) but can you make sure to run [`black`](https://github.com/psf/black) on the files you've touched?

sjfleming:
@gwaygenomics Great!  Yes, I have already run `black` already and everything is currently formatted that way.

gwaygenomics:
Running final tests now!

gwaygenomics:
thanks again Stephen!

PR #187, opened 2022-02-05
Add first cut on publishing to pypi

staylorx:
# Description

From the dicussion, I've added a GitHub action file for building sdist and wheels, publishing releases to PyPI. I've also made a small change to the setup.py to use the requirements.txt file for required packages in setup.py. Finally the /dist folder can be ignored.

Actions can fail on duplicate versions pushed to production, but duplicate fails to the test PyPI are expected.

# Checklist

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,37 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+on: [release]
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.8]
+    
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+        
+    - name: Install pypa/build
+      run: python -m pip install setuptools build wheel --user
+
+    - name: Build a binary wheel
+      run: python setup.py sdist bdist_wheel
+        
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+        skip_existing: true
+        
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,24 @@
 import pathlib
+import os
 from setuptools import setup
 from setuptools import find_packages
 
 with open("README.md", encoding="utf-8") as readme_file:
     long_description = readme_file.read()
+
+# The directory containing this file
+my_current_dir = pathlib.Path(__file__).parent
+
+# Pull requirements from the text file
+requirement_path = os.path.join(my_current_dir, "requirements.txt")
+
+install_requires = []
+if os.path.isfile(requirement_path):
+    with open(requirement_path) as f:
+        install_requires = f.read().splitlines()
+
+testing_requires = []
+install_requires = install_requires + testing_requires
 
 about = {}
 with open(pathlib.Path("pycytominer/__about__.py")) as fp:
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,37 @@
...
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+        skip_existing: true
+        
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
gwaygenomics, line 37:
```suggestion
        password: ${{ secrets.PYPI_API_TOKEN }}

```

Just adding new line char to remove red icon

setup.py
```diff
@@ -1,10 +1,25 @@
...
     long_description = readme_file.read()
 
+# The directory containing this file
+my_current_dir = pathlib.Path(__file__).parent
+
+# Pull requirements from the text file
+requirement_path = os.path.join(my_current_dir, "requirements.txt")
+
+install_requires = []
+if os.path.isfile(requirement_path):
```
gwaygenomics, line 16: if this fails (i.e. `requirement_path` is not a file) , then it looks like nothing installs. We don't want that right? It will push to pypi and then not check requirements upon install?

Maybe instead add an assertion that it is a file?

gwaygenomics: (i've wanted to do this for a while btw - reduce redundancy between this script and requirements, so, thank you!)

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Install pypa/build
+      run: python -m pip install setuptools build wheel --user
+
+    - name: Build a binary wheel
+      run: python setup.py sdist bdist_wheel
+        
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
gwaygenomics, line 29: is there any reason to post to Test PyPI if we're going to automatically post to regular PyPI? In other words, will our test PyPI be helpful here?

Or, maybe we use Test PyPI differently (push periodically to test an auto-release trigger?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,37 @@
...
+        repository_url: https://test.pypi.org/legacy/
+        skip_existing: true
+        
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
```
gwaygenomics, line 34: Is this going to push only the latest tag?

gwaygenomics (commented):
Looks good @staylorx ! 

A couple comments, but we should be good to merge soon.

setup.py
staylorx: If line 16 doesn't catch it as a file, then the install_requires is empty by line 15. We could assert, but we'll check it clearnly on line 16 anyway.

.github/workflows/publish-to-pypi.yml
staylorx: The test system is used mostly to ensure that the publishing scripts work. Once those are solid, then yes, we could just publish regularly. Still, it's always chicken and the egg.

There's always lots of room for innovation here. :)

.github/workflows/publish-to-pypi.yml
staylorx: Yes, only the branch or tag, and I'm using startsWith to lock it into the tag, not the branches. 

`github.ref` is the... well, I'll let these folks say it better. 😃 

<html>
<body>


github.ref | string | The branch or tag ref that triggered the workflow run. For branches this is the format refs/heads/<branch_name>, and for tags it is refs/tags/<tag_name>.
-- | -- | --



</body>
</html>

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

setup.py
gwaygenomics: but we don't want `install_requires` to be empty, right?

My concern is that this when this fails (i.e. an empty `install_requires`) it will fail silently. Maybe just removing line 16 is sufficient, since then if it fails, it will fail loudly

.github/workflows/publish-to-pypi.yml
gwaygenomics: got it!

gwaygenomics (commented):
Thanks for these changes @staylorx 

I think we're ready to merge after two things:

- Can you commit the new line character in my suggestion?
- Can you make the requirements.txt file fail loudly?

Then I'll approve the PR and the workflow run, merge, and create a new release tag! woo!

staylorx:
I've gone too wide on my changes. I'm going to close this one, and resubmit.

PR #196, opened 2022-05-17
Upload Logo

vincerubinetti:
image

- Design logo and create variations in SVG and PNG format.
- Provide readme for logo.
- Add logo image to main readme.


gwaygenomics (commented):
Looks great! 🤩 

Two comments:

- I'm getting a 404 error in the README image
- There's a lot of extra transparency in some of the `.svg`s (see example below). Is this standard?

Screen Shot 2022-05-17 at 9 26 45 AM

vincerubinetti:
- The 404 is because this pr isn't merged yet. I'm just linking to the location of where the logo-with-text file will be when this pr is merged.
 - The SVG doesn't have a defined width or height, and some SVG viewing programs get confused by this and render it as a perfect square. That's probably what's happening here. But don't worry, the `viewBox` attribute (what defines the boundaries of the image) is cropped correctly to the image. Depending on where and how you end up using the SVG, you may have to finagle some settings or manually define `width=""` and `height=""` attributes that match the aspect ratio of the `viewBox`. I can consult when necessary.

gwaygenomics:
Sounds good! Thanks for checking back in with this detail. I'll go ahead and merge now (cc @niranjchandrasekaran )

Issue #200, opened 2022-05-22
Add Automated Linting Check for Black Formatting


d33bs:
An automated linting check for [Black](https://github.com/psf/black) formatting would provide visibility into whether repo changes adhere standards mentioned in CONTRIBUTING.md. This lint check would fail if related code does not meet Black formatting standards.

This issue comes as the result of conversations within #195 .

d33bs:
Merged with #208

d33bs:
Reopening as the merge was with `develop`

kenibrewer:
@d33bs I was browsing old issues for things that might interest me and I think this one can be closed.

Issue #201, opened 2022-05-23
Provide Guidance or Tooling for Local Development Testing and Linting


d33bs:
This repo utilizes various testing and linting (including code-style) frameworks which are also sometimes incorporated as part of automated CI/CD workflows (for example, pytest). Developers could benefit from better understanding which frameworks are used, how they are used, and standardization on how to use them. This would help increase reproducibility and enhance development capabilities for contributors.

Examples and ideas:
- Add section(s) to `readme.md` describing what and how to test or lint.
- Add section(s) to `contributing.md` describing what and how to test or lint.
- [Act](https://github.com/nektos/act) for testing Github workflows
- [Dagger.io](https://dagger.io/) pipelining (including related Github workflows)

kenibrewer:
This issue is partially completed in #279 and #264

axiomcura:
Bump @d33bs

d33bs:
I think this is mostly good to close. One final closing pass (in my opinion) would be to make the contribution documentation, where most of this information lives, available in the readthedocs deployments so it does not get lost / is harder to miss. Maybe this could be added to the existing docs pr @gwaybio ?

gwaybio:
Good point! I'll make these updates now

gwaybio:
Added in d60f0fa90a4f085ccd2227e986bb8c8db9940224

PR #208, opened 2022-06-09
Add Black Requirement and Workflow Check

d33bs:
# Description

This change adds a [Black](https://github.com/psf/black) linting/formatting check as part of the Github workflow actions associated with this repo. This change is intended to address #200, making black formatting checks automatic and more visible for development collaboration.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ pandas>=1.2.0
 scikit-learn>=0.21.2
 sqlalchemy>=1.3.6
 pytest>=5.0.1
+black>=22.3.0
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -31,6 +31,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: ✨ Lint with black
+      run: |
+        black --check ./pycytominer
     - name: Test with pytest
       run: |
         pytest


gwaybio:
It looks like elements of #203 are sneaking into this PR as well. Before I review, can you rebase `black-linting-check` to remove commits added in #203?

You likely just need to sync your fork with the current upstream develop

gwaybio:
It looks like elements of #203 are sneaking into this PR as well. Before I review, can you rebase `black-linting-check` to remove commits added in #203?

You likely just need to sync your fork with the current upstream develop

d33bs:
Apologies and thank you @gwaybio . I've just rebased as you asked.

requirements.txt
```diff
@@ -4,3 +4,4 @@ pandas>=1.2.0
 scikit-learn>=0.21.2
 sqlalchemy>=1.3.6
 pytest>=5.0.1
+black>=22.3.0
```
gwaybio, line 7: is it standard to include a package required for a workflow as a full package requirement? Is there another way to do this, or are you not concerned by adding an extra dependency in this case? (I sometimes get weary of adding dependencies!)

d33bs: In my experience it's fairly common (though maybe not always perfect) to include development, testing, and linting tooling in a single requirements file for the repo. I think the idea is roughly that running and writing the code are close enough that they may be lumped together so as to not be lost. I might say the single `requirements.txt` file is an MVP - _minimum viable Python_ for a Python project 🙂 .

That said, there are paradigms which expand away from single requirements files to allow for "non-production" dependency management, among other features (as "developer requirements" or other specialized tooling specific to environments or versioning). Maybe we can dig into some of these as part of another issue or part of #201 ? The increased complexity for any of these is usually I think a bit of a balancing act (time/code vs. environment isolation/reproducibility/etc). See below:

- A simple way to do this is by using a `requirements-dev.txt` file (for ex., I happened to notice the [Modin repo](https://github.com/modin-project/modin) uses this pattern).
- Dependency or environment managers:
  - These involve a bigger investment towards alternative file formats and additional dependencies developers must maintain at a benefit of improving things beyond what's provided with Python by default. In and of themselves they involve an additional learning curve and time for migration (to or from). The advantage is usually that developers have improved reproducibility when it comes to their code (where otherwise we rely on intuition/experience, IDE's, etc for this) and hopefully(!) less time spent attempting to resolve dependency conflicts where pip may fall short. Some examples below.
  - [Poetry grouping](https://python-poetry.org/docs/master/managing-dependencies/) allows specification of one or many groups of dependencies for whatever need may arise. My own personal experience lands more lately with Poetry, which I've found to be nice to work with and straightforward to install (but again, not without a learning curve).
  - There are a number of others in this category, such as `pdm`, `pyenv`, `pipenv`, `pip-tools`, etc. Some of these fall in and out of favor as time goes on, and also as communities change. OS differences seem also to play a hand in how well developers are able to take advantage of these. Towards this thought, containerization and compatibility within that idea is important to think about too.
- There are some proposals within the Python community which maybe seek to address some of the ideas in your comment (see [PEP 665](https://peps.python.org/pep-0665/) of July 2021). While it looks like this PEP was rejected, there's some interesting content to consider that may be relevant for guiding the way your projects develop. I wouldn't be surprised to find out that there's another PEP or ongoing work towards this - as noted it's something which many languages (or libraries) already do or have figured out along the way.

Issue #215, opened 2022-07-20
Image based QC prior to aggregate_profiles


kvshams:
Are there any QC procedure that could be done prior to aggregating well.

In my case any images that have less cell dense region would create an artifact as the cells become larger and larger. I want to avoid those images from the aggregation steps. 

Or is there any way to get the entire db to be covert to one data frame including all features and metadata?. This would be more usable for the QC and exclude identified outliers from db and perform the downstream aggregation and analysis?.

gwaybio:
Pycytominer doesn't perform any QC at the moment.

You might consider looking into [bioprofiling.jl](https://doi.org/10.1093/bioinformatics/btab853). IIRC they have some QC ability.

You can also look into [this paper](https://doi.org/10.1371/journal.pone.0267280), which proposes some QC ideas (not yet implemented in pycytominer, see rohban-lab/Image-based-cell-profiling-enhancement-via-data-cleaning-methods#1), including one which may be helpful for adjusting for cell density.

Pycytominer does have functionality to acquire full db (SQLite) here: https://github.com/cytomining/pycytominer/blob/b4d32d39534c949ad5165f0b98b79537c2a7ca58/pycytominer/cyto_utils/cells.py#L25

kvshams:
@gwaybio Thank you for point out to the insightful method. It is a naive request.  How to create the data frame of single cell df after loading the db. 
` sc = SingleCells('sqlite:///Data/database.sqlite') # this is by default get the strata=['Metadata_Plate', 'Metadata_Well']`
ie, How I can create a data-frame contains raw single cell level data for the qc, from sqlite output created by the `ingest` (used ingest function to combine parallel processed data)

gwaybio:
We have a function inside the `SingleCells` class to merge single cells. See https://pycytominer.readthedocs.io/en/latest/pycytominer.cyto_utils.html#pycytominer.cyto_utils.cells.SingleCells.merge_single_cells

However, we recently recognized some memory issues in this function (see #195), which we're working to solve by moving away from SQLite to parquet (#213).

We'd welcome any insights and experience you have with this method

gwaybio:
@kvshams - I wanted to provide an update that the merge_single_cells() functionality is now working well. It now takes 15 minutes to merge whereas previously it was taking several hours.

Issue #195, opened 2022-05-13
Potential memory leak in SingleCell's `.merge_single_cells()` method


axiomcura:
Currently the [`.merge_single_cells()`](https://github.com/cytomining/pycytominer/blob/b2510d1229cd42495d93a605cdeb0eb2e437de70/pycytominer/cyto_utils/cells.py#L604) method in the `SingleCell` class uses extraordinarily high amount of memory for a 10GB file.

To replicate this:

```python
from pycytominer.cyto_utils.cells import SingleCells


sql_path = "SQ00014614.sqlite"
sql_url = f"sqlite:///{sql_path}"

sc_p = SingleCells(
    sql_url,
    strata=["Image_Metadata_Plate", "Image_Metadata_Well"],
    image_cols=["TableNumber", "ImageNumber"],
    fields_of_view_feature=[],
)

merged_sc = sc_p.merge_single_cells()
```

The data used in this example can be downloaded from [here](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672)

Specifically, I have downloaded the smallest file, which is `SQ00014613.sqlite` file

## Memory benchmarking

To locate where the high memory consumption is occurring, I conducted a memory profile using [memory_profiler](https://github.com/pythonprofilers/memory_profiler)

I copied the `.merge_single_cells()` source code and placed it into a script.
Then, I replaced `self` with the `SingleCell` instance variable, which is `sc_p` in this example

Below is the memory profile output:

```text

298.078 MiB  146.570 MiB           2       sc_p = SingleCells(
    30  151.508 MiB    0.000 MiB           1           sqlite_file, strata=strata, image_cols=image_cols, fields_of_view_feature=[]
    31                                             )
    32
    33  298.078 MiB    0.000 MiB           1       sc_df = ""
    34  298.078 MiB    0.000 MiB           1       linking_check_cols = []
    35  298.078 MiB    0.000 MiB           1       merge_suffix_rename = []
    36  298.078 MiB    0.000 MiB           1       for left_compartment in sc_p.compartment_linking_cols:
    37  298.078 MiB    0.000 MiB           1           for right_compartment in sc_p.compartment_linking_cols[left_compartment]:
    38
    39  298.078 MiB    0.000 MiB           1               print(right_compartment)
    40                                                     # linking checking
    41  298.078 MiB    0.000 MiB           1               linking_check = "-".join(sorted([left_compartment, right_compartment]))
[...]

PR #194, opened 2022-05-04
Fix #192 warnings for scipy mad func and other deprecated params

johnarevalo:
# Description

Change deprecated methods from external libraries.

## What is the nature of your change?

- [x] Bug fix (fixes an issue).

# Checklist

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have deleted all non-relevant text in this pull request template.

PR #219, opened 2022-08-16
Fix of `.merge_single_cells()` to Load Single-Cell Data into Dataframes

bunnech:
# Description

This pull request addresses issue [#195](https://github.com/cytomining/pycytominer/issues/195), concerning a potential memory leak in the `.merge_single_cells()` function as mentioned in pull request [#194](https://github.com/cytomining/pycytominer/pull/194#issuecomment-1179340277) and issue [#215](https://github.com/cytomining/pycytominer/issues/215#issuecomment-1192028633).

The updated function `.merge_single_cells()` combined different compartments, image data, and metadata and loads them into a Pandas dataframes. The old version did not finish converting a >10 Gb `.sqlite` files into dataframes within 4 hours. The current fix finishes the task in < 15 minutes.
This was achieved by removing the dependency to `pd.read_sql` (as suggested by @johnarevalo) and by getting rid of temporarily created dataframes of compartments to be merged which took additional memory. There might be better solutions, but this is working well for me currently. 

Let me know what you think!

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).

# Checklist


pycytominer/cyto_utils/cells.py
```diff
@@ -397,6 +397,34 @@ def get_subsample(self, df=None, compartment="cells", rename_col=True):
...
 
+    def is_feature_col(self, col):
+        """Check if column is a feature."""
+        return (
+            col.startswith("Cell")
+            or col.startswith("Cytoplasm")
+            or col.startswith("Nuclei")
+        )
+
+    def count(self, table):
```
gwaybio, line 408: please use a more descriptive name.

Maybe something like `count_sql_table_rows()`

pycytominer/cyto_utils/cells.py
```diff
@@ -397,6 +397,34 @@ def get_subsample(self, df=None, compartment="cells", rename_col=True):
...
+        """Count total number of rows for a table."""
+        (num_rows,) = next(self.conn.execute(f"SELECT COUNT(*) FROM {table}"))
+        return num_rows
+
+    def get_columns(self, table):
```
gwaybio, line 413: more descriptive name please.

For this, how about `get_sql_table_col_names()`?

pycytominer/cyto_utils/cells.py
```diff
@@ -410,9 +438,30 @@ def load_compartment(self, compartment):
...
+
+        # Use pre-allocated np.array for data
+        feats = np.empty(shape=(num_cells, num_feats), dtype=np.float64)
+        # Use pre-allocated pd.DataFrame for metadata
+        metas = pd.DataFrame(columns=meta_cols, index=range(num_cells))
+
+        # Query database for selected columns of chosen compartment
+        columns = ", ".join(meta_cols + feat_cols)
+        query = f"select {columns} from {compartment}"
+        resultset = self.conn.execute(query)
```
gwaybio, line 455: can you name `resultset` more descriptively?

pycytominer/cyto_utils/cells.py
```diff
@@ -410,9 +438,30 @@ def load_compartment(self, compartment):
...
+
+        # Load data row by row for both meta information and features
+        for i, row in enumerate(resultset):
+            metas.loc[i] = row[:num_meta]
+            feats[i] = row[num_meta:]
```
gwaybio, line 460: Is there any reason to not include the row _and_ column indexing here?

(e.g. `metas.loc[i, :] = row[:num_meta=]`, and something equivalent for `feats`)

To me, it's easier to understand that the variables you're building are 2d if we explicitly set both indices. 

We shouldn't implement my suggestion if it slows processing

pycytominer/cyto_utils/cells.py
```diff
@@ -410,9 +438,30 @@ def load_compartment(self, compartment):
...
+
+        # Return concatenated data and metainformation of compartment
+        return pd.concat(
```
gwaybio, line 463: metadata should be indexed first, right? Am I reading correctly that it is indexed _after_ morphology features in this setup?

pycytominer/cyto_utils/cells.py
```diff
@@ -658,26 +707,27 @@ def merge_single_cells(
                 ]
 
                 if isinstance(sc_df, str):
-                    initial_df = self.load_compartment(compartment=left_compartment)
+                    sc_df = self.load_compartment(compartment=left_compartment)
```
gwaybio, line 710: nice!!

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -235,10 +235,12 @@ def test_SingleCells_count():
...
 def test_load_compartment():
     loaded_compartment_df = ap.load_compartment(compartment="cells")
-    pd.testing.assert_frame_equal(loaded_compartment_df, cells_df)
+    pd.testing.assert_frame_equal(
+        loaded_compartment_df, cells_df, check_dtype=False)
 
     # Test non-canonical compartment loading
-    pd.testing.assert_frame_equal(new_compartment_df, ap_new.load_compartment("new"))
+    pd.testing.assert_frame_equal(
+        new_compartment_df, ap_new.load_compartment("new"), check_dtype=False)
```
gwaybio, line 243: I am surprised that this passes the tests, given this function:

```python
def is_feature_col(self, col):
    """Check if column is a feature."""
    return (
        col.startswith("Cell")
        or col.startswith("Cytoplasm")
        or col.startswith("Nuclei")
    )
```

Is it possible that all features prefixed with `new` are considered metadata features in your implementation?

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -307,6 +309,7 @@ def test_merge_single_cells():
                 pd.testing.assert_frame_equal(
                     norm_method_df.sort_index(axis=1),
                     manual_merge_normalize.sort_index(axis=1),
+                    check_dtype=False
```
gwaybio, line 312: This is a bit concerning to me. Can you report what changed and why ignoring column dtype is now necessary?

pycytominer/cyto_utils/cells.py
```diff
@@ -397,6 +397,34 @@ def get_subsample(self, df=None, compartment="cells", rename_col=True):
 
         self.is_subset_computed = True
 
+    def is_feature_col(self, col):
```
gwaybio, line 400: I made a comment about this function in the proposed test update.

I think that you can easily update this function to iterate through `self.compartments` rather than hardcoding canonical compartments here. (i.e. what about non-canonical compartments?)

gwaybio (changes requested):
Thanks for this contribution! I made several comments throughout that we should discuss prior to merging.

I have two additional discussion items as well. Can you comment on:

1. I'm wondering why we need to separate metadata features and morphology features at all. Would it make sense to process all features the same way? Ah, but maybe if we do that they won't be in the correct order in the final sc_df?
2. I am inclined to ask for tests for the three new functions you added (prior to name change, `is_feature_col`, `count`, and `get_columns`). I know that they are tested through the other, existing tests, but I think they will be more robust to future changes if they themselves are tested.

pycytominer/cyto_utils/cells.py
shntnu: Some quick context here for @bunnech 

The so-called canonical -- or default -- compartments are `cells`, `cytoplasm`, and `nuclei` (and their corresponding column name prefixes are `Cells_`, `Cytoplasm_`, and `Nuclei_`, respectively).

But in general, it's possible to 
- have other ("non-canonical") compartments, and 
- (I think), not have some of these canonical compartments

pycytominer/tests/test_cyto_utils/test_cells.py
bunnech: In order to read in features faster, I load them into a `np.array` (which I initialize with `dtype=np.float64`). Original metadata might be a `np.int64`, but this should not matter.

pycytominer/cyto_utils/cells.py
bunnech: I compared with `self.compartments`. Thanks for the suggestion @gwaybio  and explanation @shntnu!

pycytominer/tests/test_cyto_utils/test_cells.py
bunnech: Fixed it by checking whatever is provided in `self.compartments`.

bunnech:
> 2. I am inclined to ask for tests for the three new functions you added (prior to name change, `is_feature_col`, `count`, and `get_columns`). I know that they are tested through the other, existing tests, but I think they will be more robust to future changes if they themselves are tested.

I removed `is_feature_col` now. The other two functions are tested via pytests on `load_compartment`. Not sure which additional test to design.

bunnech:
@gwaybio, I pushed the edits suggested by you. Let me know if anything else is required!

Thanks again!

pycytominer/tests/test_cyto_utils/test_cells.py
gwaybio: sounds good!

pycytominer/cyto_utils/cells.py
bunnech: Ah, we can equally write:

```
return pd.concat([metas, pd.DataFrame(columns=feat_cols, data=feats)], axis=1)
```

Is this preferred?

bunnech: We are only using indexing via `:num_meta` or `num_meta:` for the SQL query result.

gwaybio (commented):
> > I'm wondering why we need to separate metadata features and morphology features at all. Would it make sense to process all features the same way? Ah, but maybe if we do that they won't be in the correct order in the final sc_df?

> We are required to load the morphology features as a np.array as this is much faster than adding it via a pd.DataFrame. Then I merge both into a joint pd.DataFrame, which I return per compartment. Metadata features might contain strings or objects, so a np.array here is not possible. That's why we separate them.

Got it! Makes perfect sense, thanks for your explanation.

> > I am inclined to ask for tests for the three new functions you added (prior to name change, is_feature_col, count, and get_columns). I know that they are tested through the other, existing tests, but I think they will be more robust to future changes if they themselves are tested.

> I removed is_feature_col now. The other two functions are tested via pytests on load_compartment. Not sure which additional test to design.

I like the elegance of removing `is_feature_col`. The other two functions I think can be tested rather easily:

## `count_sql_table_rows()`

```python
# INSERT TESTING FUNCTION DEFINITION
def test_sc_count_sql_table():  # or something
    # Some python pseudocode testing logic:
    sc = SingleCells()  # Initiate class equivalently as we do in other

    # Iterate over initialized compartments
    for compartment in sc.compartments:
         result_row_count = sc.count_sql_table_rows(table=compartment)
         assert  result_row_count == whatever_example_initialized as
```

## `get_sql_table_col_names()`

```python
# INSERT TESTING FUNCTION DEFINITION
def test_get_sql_table_col_names():  # or something
    # Some python pseudocode testing logic:
    sc = SingleCells()  # Initiate class equivalently as we do in other

    # Iterate over initialized compartments
    for compartment in sc.compartments:
         meta_cols, feat_cols = sc.get_sql_table_col_names(table=compartment)
         assert  meta_cols == expected_meta_cols
         assert  feat_cols == feat_cols
```

pycytominer/cyto_utils/cells.py
gwaybio: The standard is to have metadata features before morphology features in profile data.

>  Ah, we can equally write:
> return pd.concat([metas, pd.DataFrame(columns=feat_cols, data=feats)], axis=1)
> Is this preferred?


If I am correct in my read of the `pd.concat()` call that metadata will be added as the last columns, then yes please.

bunnech:
> The other two functions I think can be tested rather easily:

Thanks! Added some pytests!

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -235,10 +235,26 @@ def test_SingleCells_count():
...
+    pd.testing.assert_frame_equal(
+        loaded_compartment_df, cells_df, check_dtype=False)
 
     # Test non-canonical compartment loading
-    pd.testing.assert_frame_equal(new_compartment_df, ap_new.load_compartment("new"))
+    pd.testing.assert_frame_equal(
+        new_compartment_df, ap_new.load_compartment("new"), check_dtype=False)
+
+
+def test_sc_count_sql_table():  # or something
```
gwaybio, line 246:
```suggestion
def test_sc_count_sql_table():
```

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -235,10 +235,26 @@ def test_SingleCells_count():
...
+    # Iterate over initialized compartments
+    for compartment in ap.compartments:
+        result_row_count = ap.count_sql_table_rows(table=compartment)
+        assert result_row_count == 100
+
+
+def test_get_sql_table_col_names():  # or something
```
gwaybio, line 253:
```suggestion
def test_get_sql_table_col_names():
```

Please also assert `feat_cols`

gwaybio (approved):
Two suggestions, one additional request, and one unresolved comment.

After you address these three things, I will merge (pending passing tests)

pycytominer/tests/test_cyto_utils/test_cells.py
bunnech: For this I have no consistent ground truth.

gwaybio: won't this (or something close) work?

```python
assert feat_cols == [f"{compartment}_a", f"{compartment}_b", f"{compartment}_c", f"{compartment}_d"]
```

bunnech: Cytoplasm breaks the cycle as it has some additional things. But I could do:

```
assert f"{compartment.capitalize()}_a" in feat_cols
```

bunnech:
> Two suggestions, one additional request, and one unresolved comment.

Addressed them! Thanks a lot for the feedback and guidance!

pycytominer/cyto_utils/cells.py
gwaybio: Just to clarify. My read of `pd.concat()` is incorrect? Metadata features are inserted as the first columns?

bunnech: No, currently not. Because the existing pytests assume that the structure is morphology features, then meta features.

gwaybio: Interesting! I added #220 (this fix is beyond scope of your PR)

bunnech: I changed the existing pytests to agree with the order meta features, then morphological features, and swapped the order of the concat.

gwaybio: wonderful!

gwaybio:
closes #220 

Thanks again @bunnech !

Issue #205, opened 2022-06-08
Build SQLite conversion tool


gwaybio:
In #202 we discuss how building a SQLite2XXX (where XXX represents our preferred pycytominer data type) would enable us to sunset SQLite dependencies in this repo. 

Any tool that we build will likely contain some (or all) of the SQLite hygiene functions @d33bs wrote in #203 

I think this tool would be a new package in `cytomining` and exist outside of pycytominer

d33bs:
If it doesn't already somehow exist, it might be beneficial to consider making data type mapping decisions for the `cytomining` repos with this issue when it comes to data at-rest (files) and in-use (memory). For example, Pandas infers Pythonic datatypes per SQLite types when using `read_sql`. This can have impacts downstream, as discussed in #198 . Types could be cast or validated within the conversion efforts.

A good example of why this matters are Datetimes:
1. SQLite may store these as strings or numbers
2. Pandas (or other in-memory formats) can interpret these as strings or numbers (in addition to `datetime64` etc)
3. XXX may have yet more options

A type mapping table could look something like the following. We might also want to call out how non-values (None, NULL, np.NaN, pd.NA, etc) are handled.
SQLite Type | XXX Type
--- | ---
TEXT | XXX Object/String Type
REAL | XXX Float Type
... | ...

gwaybio:
I just had a chat with @staylorx - @d33bs I think you're all set to start up on this. Exciting! Thanks

d33bs:
Hi @gwaybio - I'm excited to dive into this! In thinking about this a bit more I'm wondering if there's any schema or standardized structure for the source data. A few questions to this effect:

- What is the "original" data format prior to arriving in SQLite (application, file format, signals etc)?
- Does this data have a formal structure (including columns, datatypes, etc) which could be referenced as part of this issue? - Also, if this structure is available, how dynamic is it (how often should we expect to grow, contract, or change at-will)?

gwaybio:
Sorry for the delayed reply, my answers as follows:

> What is the "original" data format prior to arriving in SQLite (application, file format, signals etc)?

Prior to SQLite, the data are a collection of `.csv` files derived from the program [CellProfiler](https://github.com/CellProfiler/CellProfiler). The data represent single cell morphological profiles - in other words, various different measurements (hundreds or thousands of measurements) of what single cells look like based on microscopy images.

> Does this data have a formal structure (including columns, datatypes, etc) which could be referenced as part of this issue.

[...]

PR #228, opened 2022-08-22
Add to Function Converting SQLite to Pandas DataFrame and Merging of Additional Metadata

bunnech:
# Description

This pull request is a follow-up on issue #205 and pull-request #213 after discussion with @d33bs, @shntnu, and @gwaybio, currently only a draft or proposal, which requires further discussion and development!

# Checklist


gwaybio:
Thanks @bunnech!

@d33bs - over to you. I imagine that adding a couple things will be necessary to merge this PR:

1. Modify the function to include all parameters available to the `SingleCells()` class.
2. Adding `parquet` option to https://github.com/cytomining/pycytominer/blob/master/pycytominer/cyto_utils/output.py#L12
3. Adding the `output()` argument to the function
4. Importing this function in `cells.py` and refactoring `.merge_single_cells()` to include this merge functionality only once. (I'm not convinced that this step is necessary, so please use your discretion here)
5. Add example testing file (I think you created this already in #213)
6. Add relevant tests

I'm probably missing some elements as well.

We also don't need _perfection_ here, good enough will suffice (in case this is helpful to explicitly state!). The plan is still to port this over to `pycytominer-transform`.

gwaybio:
cc @axiomcura - Erik, see the function Charlotte wrote in this PR. As you're developing the single cell normalization project and working with SQLite files, I bet that you can use (and probably heavily modify) this function to output `parquet` files in a new module in https://github.com/WayScience/cell-health-data.

We'll then add version 6 of [this data](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672/5) as parquet files.

In other words, I don't think that you need to wait on this PR to be merged to move forward.

d33bs:
Thank you @gwaybio! In looking through the changes here and your comment I wanted to follow up to ask about the following:

>2. Adding parquet option to https://github.com/cytomining/pycytominer/blob/master/pycytominer/cyto_utils/output.py#L12
>3. Adding the output() argument to the function

Would you like to constrain parquet exports beyond (or within) the options provided by [`pd.DataFrame.to_parquet`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_parquet.html)?

gwaybio:
> Would you like to constrain parquet exports beyond (or within) the options provided by [pd.DataFrame.to_parquet](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_parquet.html)?

For now, I think constraining options to those within `pd.DataFrame.to_parquet()` is good enough.

shntnu:
https://github.com/cytomining/pycytominer/pull/228#issuecomment-1224154250

Similarly, I'll tag @EchteRobert for his use in https://github.com/broadinstitute/FeatureAggregation_single_cell/issues.

Robert: The logic in Charlotte's SQLite to pandas df converter will fix the issue we discussed regarding compartment joins. 

@d33bs, you've now got a formidable team of alpha testers ready to go when it's time ;)

d33bs:
Awesome, thanks so much for your help @shntnu! 

Hi @gwaybio and/or @bunnech - within the existing `write_parquet.sqlite_to_df` function in this PR there's a reference to a metadata CSV which may be merged with the output of `SingleCells.merge_single_cells`. A few questions about this metadata file:

- What is the data source of this metadata file?
- Is there an example of a file that is used for this purpose which I may reference for further implementation?
- If there are no examples, could I ask for guidance on generating a file that could be used for testing?

Thanks in advance!

bunnech:
> What is the data source of this metadata file?

This is mostly for metainformation concerning the compounds used.

> Is there an example of a file that is used for this purpose which I may reference for further implementation?

So here I expect a table which contains information for each well of the plate. Often this contains average features per well and additional metadata on compound and gene.

Here is one example: https://github.com/jump-cellpainting/2021_Chandrasekaran_submitted/blob/main/profiles/2020_11_04_CPJUMP1/BR00116991/BR00116991_augmented.csv.gz

Let me know if you need more details. Thanks so much for looking into this!

shntnu:
> > Is there an example of a file that is used for this purpose which I may reference for further implementation?
> 
> So here I expect a table which contains information for each well of the plate. Often this contains average features per well and additional metadata on compound and gene.

It turns out that this file Charlotte is referring to is created using `pycytominer.annotate` so it is better to use the same input as `pycytominer.annotate` to perform this annotation task.

However, you could drop this metadata argument and pipe the output into `pycytominer.annotate` to produce the same outcome.

d33bs:
Understood, thank you so much @bunnech and @shntnu! I'll explore using `pycytominer.annotate` as you mentioned.

d33bs:
Hi @gwaybio, @shntnu, and @bunnech - I've made some additions just now towards the goals mentioned in this PR. I'd love your feedback and insights, please don't hesitate to let me know what may be changed to assist with this PR.

Core additions:
- enabling parquet-based output and including this as part of `merge_single_cells`
- adding optional annotation for `merge_single_cells`
- adding example cytominer-database SQLite
- relevant tests
- relevant docs additions
- enable test documentation creation

I also wanted to add some notes as I included other items noticed along the way which may otherwise look out of place:
- Local testing was showing errors regarding pycytominer-database missing and added this as an explicit dependency
- I noticed that sphinx documentation was experiencing issues due to a missing __authors__ from __about__, and added this as best I could but we may need changes here as appropriate (this becomes the copyright footer in html content, so some balancing here with the license of this repo).
- I noticed that a slew of sphinx documentation for tests were being excluded and made a change here to include it in order to make a page for the example/test SQLite file available.

docs/conf.py
```diff
@@ -43,7 +43,7 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ["**test**"]
+exclude_patterns = []
```
gwaybio, line 46: I had originally excluded tests b/c I thought that doing so would stop sphinx from compiling documentation for the tests.

Am I not understanding correctly? What does removing this do?

pycytominer/__about__.py
```diff
@@ -1,3 +1,4 @@
 __project__ = "pycytominer"
 __version__ = "0.2.0"
 __license__ = "BSD 3-Clause License"
+__author__ = "Broad Institute of MIT, Harvard, and Pycytominer Contributors"
```
gwaybio, line 4:
```suggestion
__author__ = "Pycytominer Contributors"
```

Good to know that removing this breaks sphinx. Thanks for adding it back! I originally had it just as myself, but as the project grew obviously this was no longer true.

I like simply stating pycytominer contributors here. There are many outside-the-Broad contributors.

pycytominer/cyto_utils/cells.py
```diff
@@ -452,8 +454,7 @@ def load_compartment(self, compartment):
             feats[i] = row[num_meta:]
 
         # Return concatenated data and metainformation of compartment
-        return pd.concat(
-            [metas, pd.DataFrame(columns=feat_cols, data=feats)], axis=1)
+        return pd.concat([metas, pd.DataFrame(columns=feat_cols, data=feats)], axis=1)
```
gwaybio, line 457: this looks like a format update. Is this what black recommends for the file? (we use black)

pycytominer/cyto_utils/cells.py
```diff
@@ -644,12 +645,14 @@ def _compartment_df_generator(
 
     def merge_single_cells(
         self,
-        compute_subsample=False,
-        sc_output_file="none",
-        compression_options=None,
-        float_format=None,
-        single_cell_normalize=False,
-        normalize_args=None,
+        compute_subsample: bool = False,
```
gwaybio, line 648: We haven't used this function argument declaration syntax in pycytominer, historically. I'd love to move towards it, but I wonder if it's separate pull request to do so. In other words, is there a downside to having mixed function argument declarations?

pycytominer/cyto_utils/cells.py
```diff
@@ -667,11 +670,14 @@ def merge_single_cells(
...
             Additional arguments passed as input to pycytominer.normalize().
+        platemap: str or pd.DataFrame, default None
+            optional platemap filepath str or pd.DataFrame to be used with results via annotate
 
         Returns
         -------
-        pandas.core.frame.DataFrame
-            Either a dataframe (if output_file="none") or will write to file.
+        pandas.core.frame.DataFrame or str
+            if output_file="none") returns a Pandas dataframe
```
gwaybio, line 679:
```suggestion
            if output_file="none" returns a Pandas dataframe
```

pycytominer/cyto_utils/cells.py
```diff
@@ -779,12 +785,20 @@ def merge_single_cells(
 
             sc_df = normalize(profiles=sc_df, **normalize_args)
 
+        # In case platemap metadata is provided, use pycytominer.annotate for metadata
+        if platemap is not None:
+            sc_df = annotate(
+                profiles=sc_df, platemap=platemap, output_file="none", **kwargs
```
gwaybio, line 791: we've historically stayed away from `kwargs` - not for any particular reason except to maintain more explicit function arguments.

Besides decreasing optional function arguments, is there a good reason to introduce `kwargs` back? If so, then we probably should name the argument `annotation_kwargs`.

pycytominer/cyto_utils/output.py
```diff
@@ -2,19 +2,21 @@
...
-    sep=",",
-    float_format=None,
-    compression_options={"method": "gzip", "mtime": 1},
+    df: pd.DataFrame,
+    output_filename: str,
+    output_type: str = "csv",
+    sep: str = ",",
+    float_format: Optional[str] = None,
+    compression_options: Union[str, Dict] = {"method": "gzip", "mtime": 1},
+    **kwargs,
```
gwaybio, line 19: Why is adding `kwargs` here necessary? Am I missing something?

pycytominer/cyto_utils/output.py
```diff
@@ -74,19 +78,29 @@ def output(
...
+            index=False,
+            float_format=float_format,
+            compression=compression_options,
+        )
+
+    elif output_type == "parquet":
+
+        # note: compression options will be validated against pd.DataFrame.to_parquet options
+        # raising errors and tested through Pandas, PyArrow, etc. as necessary.
+        df.to_parquet(path=output_filename, compression=compression_options)
```
gwaybio, line 98: Does `pd.DataFrame.to_parquet()` take the same compression options as `pd.DataFrame.to_csv()`?

I'm ok with being very opinionated about parquet compression. i.e. maybe it is best to always use one specifically (or never use one).

pycytominer/tests/test_annotate.py
```diff
@@ -1,19 +1,20 @@
...
+import tempfile
+
 import pandas as pd
 from pycytominer.annotate import annotate
 
 random.seed(123)
 
 # Get temporary directory
-tmpdir = tempfile.gettempdir()
+TMPDIR = tempfile.gettempdir()
```
gwaybio, line 11: Thanks for making these changes! The test suite really does need a makeover...

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -429,10 +437,102 @@ def test_merge_single_cells_subsample():
...
+    expected_sc_merged_df = annotate(
+        profiles=AP.merge_single_cells(),
+        platemap=PLATEMAP_DF,
+        join_on=["Metadata_well_position", "Metadata_Well"],
+    )
+    sc_merged_df = AP.merge_single_cells(
+        platemap=PLATEMAP_DF, join_on=["Metadata_well_position", "Metadata_Well"]
+    )
+
+    pd.testing.assert_frame_equal(sc_merged_df, expected_sc_merged_df)
```
gwaybio, line 454: I'm a bit confused how this is working. I don't see any annotate arguments in line 450?

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -429,10 +437,102 @@ def test_merge_single_cells_subsample():
...
+    )
+
+    # gather base merge_single_cells df
+    merged_sc = sc_p.merge_single_cells()
+
+    # test csv output from merge_single_cells
+    result_file = sc_p.merge_single_cells(
+        sc_output_file=os.path.join(
+            TMPDIR,
+            "test_SQ00014613.csv.gz",
```
gwaybio, line 483: Currently, this test is a bit circular.

To break the circularity, you can output this file, validate it by hand, then save it within the test suite as well (just like we did with the parquet file). Then, replace the circularity by loading the preconfigured file.

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -429,10 +437,102 @@ def test_merge_single_cells_subsample():
...
+        strata=["Image_Metadata_Plate", "Image_Metadata_Well"],
+        image_cols=["TableNumber", "ImageNumber"],
+    )
+
+    # gather base merge_single_cells df
+    merged_sc = sc_p.merge_single_cells()
+
+    # test csv output from merge_single_cells
+    result_file = sc_p.merge_single_cells(
+        sc_output_file=os.path.join(
```
gwaybio, line 481: please switch from `os` to `pathlib`

I'm not sure how much `os` legacy code exists, but we currently much prefer pathlib as a direct replacement to os.

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -429,10 +437,102 @@ def test_merge_single_cells_subsample():
...
+            TMPDIR,
+            "test_SQ00014613.parquet",
+        ),
+        output_type="parquet",
+        compression_options="snappy",
+    )
+    # note: pd.DataFrame datatypes sometimes appear automatically changed on-read, so we cast
+    # the result_file dataframe using the base dataframe's types.
+    pd.testing.assert_frame_equal(
```
gwaybio, line 504: woooooo!! 🎉 🎉

pycytominer/tests/test_cyto_utils/test_output.py
```diff
@@ -39,64 +39,88 @@
...
+    output_filename = os.path.join(TMPDIR, "test_output.parquet")
+
+    # test with base output arguments and
+    # kwargs output arguments for pd.DataFrame.to_parquet
+    output_result = output(
+        df=DATA_DF,
+        output_filename=output_filename,
+        output_type="parquet",
+        compression_options="snappy",
+        engine="auto",
```
gwaybio, line 93: ah, is this a `kwarg` example? Wait no, what is `engine`? Sorry, I missed this in my review of these updates.

I think we can be opinionated about parquet outputs (am I not thinking about downsides of this?)

pycytominer/tests/test_data/cytominer_database_example_data/shrink_SQ00014613.sqlite_for_tests.py
```diff
@@ -0,0 +1,40 @@
+"""
```
gwaybio, line 1: I want to confirm that this function isn't executed during testing. Is that correct?

requirements.txt
```diff
@@ -1,6 +1,8 @@
+cytominer-database>=0.3.4
```
gwaybio, line 1: We specifically made this an optional requirement. (see #160)

Do we need to add it here?

gwaybio (changes requested):
Looks great! Thanks @d33bs (and @bunnech !)

I added several comments, requests, and discussion items. Excited to get this merged in! (cc @axiomcura )

pycytominer/__about__.py
d33bs: Thank you for the clarification here - understood! I'll add this to a batch of changes.

pycytominer/cyto_utils/cells.py
d33bs: Yes, this ended up being what Black recommended from my local development environment (VS Code autoformat on save, terminal-based `black --check ...` shows all good). It may be helpful at some point to consider the use of [Black configuration specifications within a pyproject.toml file](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file) to make sure local linting / formatting changes are consistent (especially when it comes to Python or Black version differences). We could also pull the Black (or other linting) Github Action checks from the `develop` branch into `main/master` for greater assurances here.

pycytominer/cyto_utils/cells.py
d33bs: Thank you! I'll add this to a batch of changes.

pycytominer/cyto_utils/output.py
d33bs: Pandas `to_parquet()` vs `to_csv()` compression options are pretty different, `gzip` I think being the only overlapping option. Would you like to use the current Pandas default for `to_parquet()` compression here: `snappy`? I believe there are slight performance implications (read and write speeds) depending on what's used.

pycytominer/tests/test_annotate.py
d33bs: Cheers, thank you for the chance to make the changes!

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: With these changes `merge_single_cells` will attempt to pass kwargs to `annotate` when the platemap argument receives a dataframe, using the `merge_single_cells` dataframe result as the `annotate.profiles` argument dataframe (within the `SingleCells` method merging occurs first, then annotation). This test seeks to demonstrate an isolated/independent `merge_single_cells` + `annotate` result along with an "embedded" version of the same thing.

pycytominer/tests/test_cyto_utils/test_output.py
d33bs: `engine` came from or was for a `pd.DataFrame.to_parquet()` kwargs test demonstration, though I ended up needing to remove the flexibility downstream within `output`. Pandas with a PyArrow-based `to_parquet` implementation throws errors when unrecognized kwargs are passed. Either way, we should remove the `engine` argument here.

pycytominer/cyto_utils/cells.py
d33bs: My understanding from the Python docs and in practice is that including type hints is mostly for developer courtesy (increasing code readability/understandability) and potentially linting alerts/blocks (for example, with [Mypy](http://mypy-lang.org/)). See this note from [`typing`](https://docs.python.org/3/library/typing.html#module-typing):
> Note: The Python runtime does not enforce function and variable type annotations.

Adding to these thoughts: there are some references in the wild about "Gradual Typing" which could be adopted here ([example presentation on this topic](https://www.youtube.com/watch?v=zKre4DKAB30&t=1111s&ab_channel=PyCascades)). It may come down to feelings/attitude towards dynamic vs static typing and community developer comfort level.

I made these changes because I was focused on this method already and so as to assist Pycytominer developers in their future work here 🙂 . To your point and if you'd like, we could redact the type hinting changes for this PR and move forward on that work within a dedicated issue (maybe this should happen either way?).

pycytominer/cyto_utils/cells.py
d33bs: The conversation within the PR led me to include `output` and `annotate` as part of `merge_single_cells`, though I may have misunderstood the intent. Using `kwargs` here allowed for some flexibility and somewhat sidesteps the notion of having "too many parameters" for a method or function (specifically avoiding having to rewrite them within the function where they're otherwise already outlined). This is more of a general code sensibility topic than a strict standard, and `*args/**kwargs` blurs this line a bit I feel.

We could rename here to `annotation_kwargs` as you suggested. I can understand `kwargs` being a difficulty too, we could lean on the developer to create their own merge -> annotate -> output flow within their code, calling each function separately. Thoughts?

Also wanted to include a reference on function parameter count from Python Code Quality Authority (PyCQA):
- [Pylint R0913](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/too-many-arguments.html): "Used when a function or method takes too many arguments." This is only linting, so it's up to each community to decide what's ultimately best in terms of approach.

pycytominer/cyto_utils/output.py
d33bs: Apologies, I originally intended to use `kwargs` for `to_parquet` parameter handling but found that the related PyArrow functions are unable to flex to this and ended up removing. I'll remove `kwargs` as appropriate here.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Understood, will do, apologies about this.

docs/conf.py
d33bs: This is correct in that it avoids rendering any of the `pycytominer.tests.*.rst` documents as HTML. I changed this to include `pycytominer.tests.test_data.rst` when documentation is rendered, providing context and information on the sample data. We could store this elsewhere too, if we shouldn't render those docs. For a visual of what renders to HTML when removing the `exclude_pattern`, see the following screenshot: 

image

requirements.txt
d33bs: I missed the notes in the `readme.md`, my apologies. I had added the dependency due to having a base installation of pycytominer without the collate extra and running into errors with test_collate as a result. I'll back this out so as to not cause any complications here.

pycytominer/tests/test_data/cytominer_database_example_data/shrink_SQ00014613.sqlite_for_tests.py
d33bs: Correct, this file doesn't get executed during testing due to not having a matching [Pytest default filename pattern](https://docs.pytest.org/en/7.1.x/getting-started.html#run-multiple-tests). This may be validated through the command `pytest --collect-only`. To be extra careful I'll remove `tests` from the filename to make this more explicit. We could get more specific here with a [Pytest configuration file](https://pytest.org/en/latest/reference/customize.html) as well, though this may be better suited for an external issue's focus.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Great point, thank you, I'll make this change for both the CSV and parquet files.

pycytominer/cyto_utils/cells.py
d33bs: Just another quick note here: in looking at this further, I noticed that `output.output_type` is why we utilize `**kwargs` for both annotation and output within `merge_single_cells` (as well as in those respective functions). It may be best to leave this as a generalized name so as to not confuse.

pycytominer/cyto_utils/output.py
d33bs: A quick follow up and correction here: `**kwargs` are used in this instance to enable `merge_single_cells` to utilize both annotation and output arguments together (and not conflict when seeing additional unrecognized arguments).

docs/conf.py
gwaybio: Gotcha! Yes, I think it's best to remove tests from the docs. I do not think I've ever seen tests in docs before.

pycytominer/cyto_utils/cells.py
gwaybio: Gotcha! Yes, we'll want to standardize. Likely outside scope of the present PR.

pycytominer/cyto_utils/cells.py
gwaybio: Thanks for linking the video! Yes, let's embrace gradual typing. I will keep this in mind in all future pull request reviews.

pycytominer/cyto_utils/cells.py
gwaybio: I am in favor of reducing confusion. @shntnu mentions many alpha testers available. I'm in favor merging this in as you feel reduces most confusion, then receiving feedback from these testers (including Erik) 

> we could lean on the developer to create their own merge -> annotate -> output flow within their code, calling each function separately.

I am generally in favor of leaning on the developer, but I also lean towards making a developer's life easier. See comment above about alpha testers. If they deem kwargs too confusing (or if they end up establishing their own flow we can drop kwargs in the future.

pycytominer/cyto_utils/output.py
gwaybio: yes, for now, let's use default pandas `snappy` compression.

I think that updating `compression_options` for `to_parquet()` might be slightly out of scope.

pycytominer/tests/test_cyto_utils/test_cells.py
gwaybio: Gotcha!

gwaybio (commented):
I answered all discussion items @d33bs - Looks great!

I think we're nearly ready to merge this in (maybe a couple more changes necessary based on the comments).

We can then test this functionality to see if any more tweaks are needed (the testers are encouraged to contribute these changes directly!)

One additional comment: Thanks for adding the testing files, do you know why the `parquet` file is so much larger than the `.csv`?

image

docs/conf.py
d33bs: Understood - I'll add the exclude flag for that suite of documents back in with a related commit.

pycytominer/cyto_utils/cells.py
d33bs: Sounds good, I opened #230 to help address this in the future.

pycytominer/cyto_utils/cells.py
d33bs: Sounds great! Thank you.

pycytominer/tests/test_cyto_utils/test_output.py
d33bs: Just a note here: Engine was removed as this kwarg is no longer available to be used within the output function.

pycytominer/tests/test_data/cytominer_database_example_data/shrink_SQ00014613.sqlite_for_tests.py
d33bs: Just a note here: filename was changed to avoid any issues with pytest runs.

requirements.txt
d33bs: Just a note here: dependency removed in a further commit.

pycytominer/cyto_utils/cells.py
d33bs: Sounds good - I'd also love to hear back from the testers on preference here, deferring to those who make most use of it.

Additional food for thought (and may not apply or be debatable): ["Preference for Deep Classes" by John Ousterhout](https://youtu.be/bmSAYlu0NcY?t=753).

pycytominer/cyto_utils/output.py
d33bs: Understood, thank you. I'll update this with an additional commit.

d33bs:
Cheers, thank you @gwaybio for the great comments and discussion.

> One additional comment: Thanks for adding the testing files, do you know why the `parquet` file is so much larger than the `.csv`?

This can boil down to the differences in purpose for the compression types being used: gzip and snappy. My understanding is that gzip is more optimized for storage compression (lower filesizes) and snappy is more optimized for compressed reads + writes in addition to being splittable. In very small scenarios there are a few things to note: parquet comes along with additional metadata and has features (columnar focus, row groups, data pages) which make the storage consequences a bit different than CSV. A quick example provided at the bottom shows how storage is effected as row sizes increase. CSV beats parquet storage with 4 rows but grows quickly as more rows are added.

I did some testing using `pd.DataFrame.to_parquet` with gzip, and brotli, both of which failed to yield a smaller file. There are other configuration tweaks I'm investigating here too, but wanted to follow up with this comment for now.

Some references:
- This comment from Stack Overflow has a good overview (more of a Spark context, but informative): https://stackoverflow.com/a/56410326
- From the [Snappy Wikipedia page](https://en.wikipedia.org/wiki/Snappy_(compression)): 
  > "[Snappy] does not aim for maximum compression, or compatibility with any other compression library; instead, it aims for very high speeds and reasonable compression."

Example mentioned above:

```python
import pathlib

import pandas as pd

pq_filename = "test.parquet"
csv_filename = "test.csv"

# 4 row dataframe
df = pd.DataFrame({"a": [0, 0, 1, 1], "b": [0, 1, 0, 1]})

df.to_csv(csv_filename)
df.to_parquet(pq_filename)

print("4 row dataframe")
print("CSV size:", pathlib.Path(csv_filename).stat().st_size)
print("Parquet size:", pathlib.Path(pq_filename).stat().st_size)

# 400 row dataframe
df = pd.DataFrame({"a": [0, 0, 1, 1] * 100, "b": [0, 1, 0, 1] * 100})

df.to_csv(csv_filename)
df.to_parquet(pq_filename)

print("\n400 row dataframe")
print("CSV size:", pathlib.Path(csv_filename).stat().st_size)
print("Parquet size:", pathlib.Path(pq_filename).stat().st_size)

# expected output (bytes):
# 4 row dataframe
# CSV size: 29
# Parquet size: 2238
# 
# 400 row dataframe
# CSV size: 3095
# Parquet size: 2256
```

gwaybio (approved):

vincerubinetti (approved):

falquaddoomi (approved):

gwaybio:
Alrighty, we're all set to merge!

Thanks @bunnech and @d33bs - OK @shntnu and @axiomcura, we're all set to start testing this

shntnu:
This PR is a treasure trove of engineering and design insights! Thank you, @d33bs and @gwaybio for the discussions!

I was wondering: what's a good way of demonstrating the new functionality offered by this PR, to alpha testers? 

Do you think pointing them to the test https://github.com/cytomining/pycytominer/blob/f3206dae848bfd25b57b11c4f276d2a3de419fd1/pycytominer/tests/test_cyto_utils/test_cells.py#L458 in specific, and `test_cells.py` in general is the way to go?

And the tl;dr is: 

> You can now load a SQLite file, (optionally) subsample, merge it and store it in a single dataframe in a jiffy (~10mins for a typical ~10Gb file), and (optionally) add metadata columns when you are at it. You can also pass options to write to disk (Parquet is preferred for speed). 

Right?

d33bs:
Hi @shntnu - thank you so much for the kind words! Truly appreciate your help in guiding and testing for this PR as well!

Addressing your questions:
- Testing: Yes, `test_merge_single_cells_cytominer_database_test_file()` I believe represents good examples of how to use the new functionality offered by this PR. It may be worthwhile to also test the following from `test_output.py`  https://github.com/cytomining/pycytominer/blob/master/pycytominer/tests/test_cyto_utils/test_output.py#L80 
- tl;dr: Yes, that's a good summary. I'd add: "Merge single cells data from SQLite to parquet (or CSV) in a jiffy with optional annotation all from one method".

Issue #233, opened 2022-09-30
Operating system kernel kills `merge_single_cells()` process due to out of memory error


axiomcura:
Hello everyone,

I have been using the newly developed `merge_single_cells()` (introduced in #219) to convert my `.sqlite` files into parquet. 

I am currently running into an issue where large files (specifically over 20GB) are being killed by Operating System due to out of memory (OOM) error. 

![Killed Message](https://user-images.githubusercontent.com/31600622/193365101-8f0d1e3d-52f7-49ed-9c84-3a659b56e806.png)

The image shows that some `sqlite` files were successfully converted into parquet files but the OS kernel raised a `killed` message on `../data/SQ00014611.sqlite.`

To figure out what error caused the OS kernel to kill the process, I typed: 

```
sudo dmesg -T| grep -E -i -B100 'killed process'
```

and returned: 

```
Out of memory: Killed process 137056 (python) total-vm:58479492kB
```

Based on the message it, seems the `.merge_single_cells()` at some point uses 58.5GB of memory on a 20.78GB file.

Here is an image of `PopOS` resource monitor when using `.merge_single_cells()` on `../data/SQ00014611.sqlite.` dataset a few seconds before the OS kernel killing the job and raising OOM error:

![ PopOS Monitor](https://user-images.githubusercontent.com/31600622/193365170-15c6ef08-4483-458a-aa3c-d32ac67b7f62.png)

Below is the source code used and a download link that points to the dataset that causes this issue. 

 SQ00014611 dataset download [link](https://nih.figshare.com/ndownloader/files/18508583)

```python
# creating single cell object
sqlite_path = "./SQ00014611.sqlite"
sql_url = f"sqlite:///{sqlite_path}"
sc_p = SingleCells(
    sql_url,
    strata=["Image_Metadata_Plate", "Image_Metadata_Well"],
    image_cols=["TableNumber", "ImageNumber"],
)

# setting up output paths and naming
f_name = Path(sqlite_path).stem
save_dir = Path("../cell-health-parquet-data")
save_dir.mkdir(exist_ok=True)
save_path = save_dir / f"{f_name}.parquet"

# converting to parquet files
sc_p.merge_single_cells(
    sc_output_file=save_path,
    output_type="parquet",
    join_on=["Metadata_well_position", "Image_Metadata_Well"],
)
```
## Memory benchmarking

[...]

PR #234, opened 2022-10-06
Reduce memory consumption of SingleCells.merge_single_cells

d33bs:
# Description

These changes seek to address #233 memory challenges when using `SingleCells.merge_single_cells` by segmenting Pandas merges using dynamic argument `chunksize` and related dataframe concatenation within the method. In brief testing using small datasets with these changes I saw around 12-18% reduction in peak memory use (using `chunksize` default).

Using this segmented merge approach we're making a tradeoff between memory and time; in large data scenarios we may avoid out-of-memory issues and the procedure may take longer due to more (but smaller) merges being required. `chunksize` is an optional argument which when left to it's default of `None` will automatically be set to 1/3rd of the first compartment dataframe rowlength. Performance may vary with `chunksize` contingent on the dataset and is the reasoning for this being an argument which may be optionally set by a developer.

__Caveats:__
- Leveraging existing suffix column naming within the merges along with the increase in number of merges increases the number of warnings emitted during testing. (`FutureWarning: Passing 'suffixes' which cause duplicate columns {'ObjectNumber_cytoplasm'} in the result is deprecated and will raise a MergeError in a future version.`).
- Testing via `test_cells.py` needed updates in order for the changes within `cells.py` to pass pytesting. With these changes I focused on ensuring the data itself was as expected. I'd ask for double-checks here especially to make sure everything looks good.
- Performance will likely vary based on the resources of the system running pycytominer and the dataset involved. Using the `chunksize` default does not guarantee avoidance of out-of-memory issues and instead provides flexibility for developer implementation.

Thank you @axiomcura for the excellent writeup on the performance issue!

- [ ] Enhancement (adds functionality)

# Checklist

- [ ] My changes generate no new warnings. (see caveats above)

### DIFF
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -419,13 +419,16 @@ class SingleCells(object):
 
         return meta_cols, feat_cols
 
-    def load_compartment(self, compartment):
+    def load_compartment(self, compartment, float_datatype: type = np.float64):
         """Creates the compartment dataframe.
 
         Parameters
         ----------
         compartment : str
             The compartment to process.
+        float_datatype: type, default np.float64
+            Numpy floating point datatype to use for load_compartment and resulting dataframes.
+            Please note: using any besides np.float64 are experimentally unverified.
 
         Returns
         -------
@@ -439,7 +442,7 @@ class SingleCells(object):
         num_meta, num_feats = len(meta_cols), len(feat_cols)
 
         # Use pre-allocated np.array for data
-        feats = np.empty(shape=(num_cells, num_feats), dtype=np.float64)
+        feats = np.empty(shape=(num_cells, num_feats), dtype=float_datatype)
         # Use pre-allocated pd.DataFrame for metadata
         metas = pd.DataFrame(columns=meta_cols, index=range(num_cells))
 
@@ -653,6 +656,7 @@ class SingleCells(object):
         normalize_args: Optional[Dict] = None,
         platemap: Optional[Union[str, pd.DataFrame]] = None,
         sc_merge_chunksize: Optional[int] = None,
+        float_datatype: type = np.float64,
         **kwargs,
     ):
         """Given the linking columns, merge single cell data. Normalization is also supported.
@@ -677,6 +681,9 @@ class SingleCells(object):
             Chunksize for merge and concatenation operations to help address performance issues
             note: if set to None, will infer a chunksize which is the roughly 1/3 the row length
             of first component df.
+        float_datatype: type, default np.float64
+            Numpy floating point datatype to use for load_compartment and resulting dataframes.
+            Please note: using any besides np.float64 are experimentally unverified.
 
         Returns
         -------
@@ -710,7 +717,9 @@ class SingleCells(object):
                 ]
 
                 if sc_df.empty:
-                    sc_df = self.load_compartment(compartment=left_compartment)
+                    sc_df = self.load_compartment(
+                        compartment=left_compartment, float_datatype=float_datatype
+                    )
 
                     # if chunksize was not set, set it to roughly
                     # one third the size of our initial compartment
@@ -733,7 +742,9 @@ class SingleCells(object):
                 # sc_merge_chunksize to help constrain memory
                 sc_df = pd.concat(
                     [
-                        self.load_compartment(compartment=right_compartment).merge(
+                        self.load_compartment(
+                            compartment=right_compartment, float_datatype=float_datatype
+                        ).merge(
                             right=right_chunk,
                             # note: we reverse left and right for join key merge order reference
                             left_on=self.merge_cols + [right_link_col],
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -652,6 +652,7 @@ class SingleCells(object):
         single_cell_normalize: bool = False,
         normalize_args: Optional[Dict] = None,
         platemap: Optional[Union[str, pd.DataFrame]] = None,
+        chunksize: Optional[int] = None,
         **kwargs,
     ):
         """Given the linking columns, merge single cell data. Normalization is also supported.
@@ -672,6 +673,10 @@ class SingleCells(object):
             Additional arguments passed as input to pycytominer.normalize().
         platemap: str or pd.DataFrame, default None
             optional platemap filepath str or pd.DataFrame to be used with results via annotate
+        chunksize: int, default None
+            chunksize for merge and concatenation operations to help address performance issues
+            note: if set to None, will infer a chunksize which is the roughly 1/3 the row length
+            of first component df.
 
         Returns
         -------
@@ -681,7 +686,7 @@ class SingleCells(object):
         """
 
         # Load the single cell dataframe by merging on the specific linking columns
-        sc_df = ""
+        sc_df = pd.DataFrame()
         linking_check_cols = []
         merge_suffix_rename = []
         for left_compartment in self.compartment_linking_cols:
@@ -704,8 +709,12 @@ class SingleCells(object):
                     left_compartment
                 ]
 
-                if isinstance(sc_df, str):
+                if sc_df.empty:
                     sc_df = self.load_compartment(compartment=left_compartment)
+
+                    # if chunksize was not set,
+                    if chunksize is None:
+                        chunksize = round(len(sc_df) / 3)
 
                     if compute_subsample:
                         # Sample cells proportionally by self.strata
@@ -719,20 +728,25 @@ class SingleCells(object):
                             sc_df, how="left", on=subset_logic_df.columns.tolist()
                         ).reindex(sc_df.columns, axis="columns")
 
-                    sc_df = sc_df.merge(
-                        self.load_compartment(compartment=right_compartment),
-                        left_on=self.merge_cols + [left_link_col],
-                        right_on=self.merge_cols + [right_link_col],
-                        suffixes=merge_suffix,
-                    )
-
-                else:
-                    sc_df = sc_df.merge(
-                        self.load_compartment(compartment=right_compartment),
-                        left_on=self.merge_cols + [left_link_col],
-                        right_on=self.merge_cols + [right_link_col],
-                        suffixes=merge_suffix,
-                    )
+                # perform a segmented merge using pd.concat and
+                # chunksize to help constrain memory
+                sc_df = pd.concat(
+                    [
+                        self.load_compartment(compartment=right_compartment).merge(
+                            right=right,
+                            # note: we reverse left and right for join key merge order reference
+                            left_on=self.merge_cols + [right_link_col],
+                            right_on=self.merge_cols + [left_link_col],
+                            # note: we reverse left and right for join keys
+                            suffixes=reversed(merge_suffix),
+                            how="inner",
+                        )
+                        for right in [
+                            sc_df[i : i + chunksize]
+                            for i in range(0, sc_df.shape[0], chunksize)
+                        ]
+                    ]
+                )
 
                 linking_check_cols.append(linking_check)
 
@@ -759,8 +773,18 @@ class SingleCells(object):
             self.load_image()
             self.load_image_data = True
 
+        # perform a segmented merge using pd.concat and
+        # chunksize to help constrain memory
         sc_df = (
-            self.image_df.merge(sc_df, on=self.merge_cols, how="right")
+            pd.concat(
+                [
+                    self.image_df.merge(right=right, on=self.merge_cols, how="right")
+                    for right in [
+                        sc_df[i : i + chunksize]
+                        for i in range(0, sc_df.shape[0], chunksize)
+                    ]
+                ]
+            )
             # pandas rename performance may be improved using copy=False, inplace=False
             # reference: https://ryanlstevens.github.io/2022-05-06-pandasColumnRenaming/
             .rename(
@@ -769,6 +793,10 @@ class SingleCells(object):
                 self.full_merge_suffix_rename, axis="columns", copy=False, inplace=False
             )
         )
+
+        # reset the index to address above concat merges and memory conservation (inplace)
+        sc_df.reset_index(inplace=True, drop=True)
+
         if single_cell_normalize:
             # Infering features is tricky with non-canonical data
             if normalize_args is None:
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -712,7 +712,8 @@ class SingleCells(object):
                 if sc_df.empty:
                     sc_df = self.load_compartment(compartment=left_compartment)
 
-                    # if chunksize was not set,
+                    # if chunksize was not set, set it to roughly
+                    # one third the size of our initial compartment
                     if chunksize is None:
                         chunksize = round(len(sc_df) / 3)
 
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -652,7 +652,7 @@ class SingleCells(object):
         single_cell_normalize: bool = False,
         normalize_args: Optional[Dict] = None,
         platemap: Optional[Union[str, pd.DataFrame]] = None,
-        chunksize: Optional[int] = None,
+        sc_merge_chunksize: Optional[int] = None,
         **kwargs,
     ):
         """Given the linking columns, merge single cell data. Normalization is also supported.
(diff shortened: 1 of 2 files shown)


pycytominer/cyto_utils/cells.py
```diff
@@ -652,6 +652,7 @@ def merge_single_cells(
         single_cell_normalize: bool = False,
         normalize_args: Optional[Dict] = None,
         platemap: Optional[Union[str, pd.DataFrame]] = None,
+        chunksize: Optional[int] = None,
```
gwaybio, line 655:
```suggestion
        sc_merge_chunksize: Optional[int] = None,
```

☝️ just a suggestion - it would be great to have a more specific variable name. Please update my suggestion if you can think of a better name.

pycytominer/cyto_utils/cells.py
```diff
@@ -672,6 +673,10 @@ def merge_single_cells(
             Additional arguments passed as input to pycytominer.normalize().
         platemap: str or pd.DataFrame, default None
             optional platemap filepath str or pd.DataFrame to be used with results via annotate
+        chunksize: int, default None
```
gwaybio, line 676: Update docs to account for updated variable name

pycytominer/cyto_utils/cells.py
```diff
@@ -719,20 +729,25 @@ def merge_single_cells(
...
-                        left_on=self.merge_cols + [left_link_col],
-                        right_on=self.merge_cols + [right_link_col],
-                        suffixes=merge_suffix,
-                    )
+                # perform a segmented merge using pd.concat and
+                # chunksize to help constrain memory
+                sc_df = pd.concat(
+                    [
+                        self.load_compartment(compartment=right_compartment).merge(
+                            right=right,
```
gwaybio, line 737:
```suggestion
                            right=right_chunk,
```

pycytominer/cyto_utils/cells.py
```diff
@@ -719,20 +729,25 @@ def merge_single_cells(
...
+                            # note: we reverse left and right for join key merge order reference
+                            left_on=self.merge_cols + [right_link_col],
+                            right_on=self.merge_cols + [left_link_col],
+                            # note: we reverse left and right for join keys
+                            suffixes=reversed(merge_suffix),
+                            how="inner",
+                        )
+                        for right in [
```
gwaybio, line 745:
```suggestion
                        for right_chunk in [
```

To increase clarity

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -300,21 +304,20 @@ def test_merge_single_cells():
...
-    assert (
-        manual_merge.Metadata_ObjectNumber_cells.tolist()
-        == sc_merged_df.Metadata_ObjectNumber.tolist()
+    assert_cols = [
+        "Metadata_ObjectNumber",
+        "Metadata_ObjectNumber_cytoplasm",
+        "Metadata_ObjectNumber_cells",
+    ]
+    # check that we have the same data using same cols, sort and a reset index
+    pd.testing.assert_frame_equal(
```
gwaybio, line 313: this doesn't looked blacked to me, please confirm

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -493,8 +517,8 @@ def test_merge_single_cells_cytominer_database_test_file():
     # note: pd.DataFrame datatypes sometimes appear automatically changed on-read, so we cast
     # the result_file dataframe using the base dataframe's types.
     pd.testing.assert_frame_equal(
-        pd.read_csv(csv_path).astype(merged_sc.dtypes.to_dict()),
-        pd.read_csv(result_file).astype(merged_sc.dtypes.to_dict()),
+        pd.read_csv(csv_path).astype(merged_sc.dtypes.to_dict())[merged_sc.columns],
```
gwaybio, line 520: why is this necessary now?

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -300,21 +304,20 @@ def test_merge_single_cells():
...
-        == sc_merged_df.Metadata_ObjectNumber.tolist()
+    assert_cols = [
+        "Metadata_ObjectNumber",
+        "Metadata_ObjectNumber_cytoplasm",
+        "Metadata_ObjectNumber_cells",
+    ]
+    # check that we have the same data using same cols, sort and a reset index
+    pd.testing.assert_frame_equal(
+        left=manual_merge[assert_cols]
+        .sort_values(by=assert_cols, ascending=True)
```
gwaybio, line 315: I am a bit concerned that sorting is giving us undue confidence of equivalent results. Am I reading this wrong? How should I be thinking about this?

Can you add an explicit check that the object number is the same? Oh, but maybe this code block is doing exactly that (and more!)?

pycytominer/cyto_utils/cells.py
```diff
@@ -759,8 +774,18 @@ def merge_single_cells(
...
             self.load_image_data = True
 
+        # perform a segmented merge using pd.concat and
+        # chunksize to help constrain memory
         sc_df = (
-            self.image_df.merge(sc_df, on=self.merge_cols, how="right")
+            pd.concat(
+                [
+                    self.image_df.merge(right=right, on=self.merge_cols, how="right")
+                    for right in [
```
gwaybio, line 783: right_chunk

gwaybio (changes requested):
Looks good! Please address my comments (focusing on the testing comments)

Also, would reducing to `float32` here help a lot as well? This is within scope of the name of your PR

pycytominer/cyto_utils/cells.py
d33bs: Thanks! I like the one you suggest better, adding to the commits.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: I ran black on this, and `--check` doesn't seem to flag it (may be profile related). I'm seeing that there's a list end bumping up against the comment, I'll add in a newline there with new changes. Are you seeing anything else that looks stylistically out of place which I may change?

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Great comment, thank you! The chunked merges made the data inconsistent with the existing tests. The check here now looks at columns "Metadata_ObjectNumber", "Metadata_ObjectNumber_cytoplasm", "Metadata_ObjectNumber_cells", sorted by the same columns in the same order for data equivalency and does include an Metadata_ObjectNumber check as a result. 

Would users of `merge_single_cells` expect for sorted output by the same columns or similar? We could include this as an additional step within the method, which would likely avoid needing to do that manually within testing here.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Thank you for calling this out! For inner merges, Pandas "...preserve[s] the order of the left keys." ([reference: pd.merge(..., how)](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.merge.html)). The column specification here is provided to account for the now differently ordered columns due to the "right" and "left" being swapped within the compartment merges. I also felt that accounting for exact column ordering over time may be unwieldy, so this may provide benefit towards development velocity. That said, we may need a change; do you think we should enforce strict column order within `merge_single_cells` to remove the need for specification here?

d33bs:
Thank you @gwaybio for the comments and feedback! I added comment + changes for individual thoughts and wanted to follow up here to your mention about `float32`. Converting from `float64` (a default which I believe cascades from numpy to pandas) to `float32` would very likely reduce the memory footprint of this method. Because this could mean data loss under some circumstances, would it be beneficial to make this an additional argument to allow development options?

gwaybio:
Yes, good idea to make an optional/experimentally-untested argument

d33bs:
@gwaybio and @axiomcura - I made some changes to allow optional float_dtype specification for merge_single_cells. The argument, `float_datatype`, defaults to using `np.float64` to ensure no accidental data loss (developers must specify a smaller or different datatype). When using `np.float32` instead of the default `np.float64` I'm seeing around 48% reduction in memory use with small tests. This may vary contingent on the compartments, features, and overall data size being used by the method.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: @gwaybio - just wanted to follow up here with some thoughts about `merge_single_cells` column order specification. If we make explicit column sorting a part of the method, we should be able to reduce the variability of data produced. Note: tests may still require column filtering (for example, if certain data do not exist in one dataframe vs another).

I wrote the following as an example of what we could do here:

```python
import pandas as pd

# create an example dataframe with mixed order columns
df = pd.DataFrame(
    {
        "Image_Metadata_1": [0],
        "TableNumber": [0],
        "ImageNumber": [0],
        "Cytoplasm_Data_1": [0],
        "Nuclei_Data_1": [0],
        "Cells_Data_1": [0],
        "Mito_Data_1": [0],
        "Actin_Data_1": [0],
        "Image_Data_1": [0],
    }
)

# print df columns as a python list, representing new order
print("Initial order:")
print(df.columns.tolist())


def custom_sort(value: str):
    """
    A custom sort for Pycytominer merge_single_cells
    pd.dataframe columns
    """

    # lowercase str which will be used for comparisons
    # to avoid any capitalization challenges
    value_lower = value.lower()

    # first sorted values (by list index)
    sort_first = ["tablenumber", "imagenumber"]

    # middle sort value
    sort_middle = "metadata"

    # sorted last (by list order enumeration)
    sort_later = [
        "cells",
        "cytoplasm",
        "nuclei",
        "image",
    ]

    # if value is in the sort_first list
    # return the index from that list
    if value_lower in sort_first:
        return sort_first.index(value_lower)

    # if sort_middle is anywhere in value return
    # next index value after sort_first values
    elif sort_middle in value_lower:
        return len(sort_first)

    # if any sort_later are found as the first part of value
    # return enumerated index of sort_later value (starting from
    # relative len based on the above conditionals and lists)
    elif any(value_lower.startswith(val) for val in sort_later):
        for k, v in enumerate(sort_later, start=len(sort_first) + 1):
            if value_lower.startswith(v):
                return k

    # else we return the total length of all sort values
    return len(sort_first) + len(sort_later) + 1


# inner sorted alphabetizes any columns which may not be part of custom_sort
# outer sort provides pycytominer-specific column sort order
df = df[sorted(sorted(df.columns), key=custom_sort)]

# print df columns as a python list, representing new order
print("\nSorted order:")
print(df.columns.tolist())
```

Which has printed output:

```shell
Initial order:
['Image_Metadata_1', 'TableNumber', 'ImageNumber', 'Cytoplasm_Data_1', 'Nuclei_Data_1', 'Cells_Data_1', 'Mito_Data_1', 'Actin_Data_1', 'Image_Data_1']

Sorted order:
['TableNumber', 'ImageNumber', 'Image_Metadata_1', 'Cells_Data_1', 'Cytoplasm_Data_1', 'Nuclei_Data_1', 'Image_Data_1', 'Actin_Data_1', 'Mito_Data_1']
```

gwaybio:
@d33bs - sorry to have lost track of this PR. Can you remind me where we are with this? is it ready to merge?

d33bs:
> @d33bs - sorry to have lost track of this PR. Can you remind me where we are with this? is it ready to merge?

Thanks @gwaybio and no worries! I feel that this PR should be closed without a merge, with justifications and detailed thoughts below.

Justifications:
- We've bound at least three major changes within this PR which could each otherwise be treated separately: **1.** Pandas merge performance enhancements, **2.** floating point datatype options and defaults, **3.** column order within merged results (definition as well as implementation of order).
- (pertaining to change **1.**) As-is, this PR I think may add greater complexity surrounding how the Pandas merge operations function with `merge_single_cells`. Adding greater complexity with Pandas merge operations is risky (lowering understandability, somewhat unique to Pandas data handling, etc) and could prove unsustainable for development in pycytominer. There may be other more desirable chunking options (for example, [as suggested here](https://github.com/cytomining/pycytominer/issues/142#issuecomment-1344483212)).

Forward momentum:
1. (pertaining to change **2.**) The floating point datatype options and defaults feel like the most clear way to address performance issues, accepting the caveat that this is a temporary and maybe sometimes scientifically unverified path forward. This may also be the quickest way to enable progress for wayscience/cell-health-data#10 (>=20 GB dataset resource exceptions).
    - Suggested action items:
        1. Open a new issue to outline the focus and need here
        2. Open a new PR for review with roughly the content of [298bf3a](https://github.com/cytomining/pycytominer/pull/234/commits/298bf3aab28e4ffed53512debc91dbaba9a11a3c) and [b4634e2](https://github.com/cytomining/pycytominer/pull/234/commits/b4634e271d498c5c70486a57939ced903f23f7c2).
2. (pertaining to change **3.**) Would it be valuable to programmatically address the column order of `merge_single_cells` results within pycytominer (for example, [see comment](https://github.com/cytomining/pycytominer/pull/234#discussion_r996191619))? As things stand, some of this is implied through Pandas merge and join operations. I'm uncertain about what the downstream effects might be to implementing changes here, but it may be a useful challenge to face in order to address general merge/join performance with reproducibility in mind.
    - Suggested action items:
        1. If it's deemed valuable, open a new issue to outline the focus and need here.
4.

d33bs:
> @d33bs - sorry to have lost track of this PR. Can you remind me where we are with this? is it ready to merge?

Thanks @gwaybio and no worries! I feel that this PR should be closed without a merge, with justifications and detailed thoughts below. Very open to other ideas and discussion here! 🙂 

**Justifications on Closing without Merge:**
- We've bound at least three major changes within this PR which could each otherwise be treated separately: **1.** Pandas merge performance enhancements, **2.** floating point datatype options and defaults, **3.** column order within merged results (definition as well as implementation of order).
- (pertaining to change **1.**) As-is, this PR I think may add greater complexity surrounding how Pandas merge operations function within `merge_single_cells`. Adding greater complexity with Pandas merge operations may be risky (lowering understandability, somewhat unique to Pandas data handling, etc) and could prove unsustainable for development in pycytominer. There may be other more desirable chunking options (for example, [as suggested here](https://github.com/cytomining/pycytominer/issues/142#issuecomment-1344483212) and below).

**Forward momentum:**
- (pertaining to change **2.**) The floating point datatype options and defaults feel like the most clear way to address performance issues, accepting the caveat that this is a temporary and maybe sometimes scientifically unverified path forward. This may also be the quickest way to enable progress for wayscience/cell-health-data#10 (>=20 GB dataset resource exceptions).
    - Suggested action items:
        1. We open a new issue to outline the focus and need here
        2. We open a new PR for review with roughly the content of [298bf3a](https://github.com/cytomining/pycytominer/pull/234/commits/298bf3aab28e4ffed53512debc91dbaba9a11a3c) and [b4634e2](https://github.com/cytomining/pycytominer/pull/234/commits/b4634e271d498c5c70486a57939ced903f23f7c2).
- (pertaining to change **3.**) Would it be valuable to programmatically address the column order of `merge_single_cells` results within pycytominer (for example, [see comment](https://github.com/cytomining/pycytominer/pull/234#discussion_r996191619))? As things stand, some of this is implied through Pandas merge and join operations. I'm uncertain about what the downstream effects might be to implementing changes here, but it may be a useful challenge to face in order to address general merge/join performance with reproducibility in mind.
    - Suggested action items:
        1. If it's valuable, we open a new issue to outline the focus and need here.
- (pertaining to change **1.**) The items above would provide greater definition for compartment merge performance considerations. Combining thoughts from other conversations, I wonder what performance gains might be possible through the use of an in-memory [DuckDB client](https://duckdb.org/docs/api/python/overview) for SQLite data extraction and chunking as [PyArrow Tables](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html). This may only make sense in scenarios where we export to a file (as otherwise we may face too many penalties in cross-conversions).
    - Brainstorming: one somewhat in-place approach could involve adding an optional parameter for `load_compartment` to return Arrow (defaulting to Pandas). If Arrow is optionally used within `merge_single_cells`, we could define a chunked join to compare performance. It could also make sense to perform the join work without loading each compartment first. The optional nature of the in-memory data format selection could help address backwards compatibility (with Pandas) while acknowledging and striving for greater design scalability (with Arrow).

What are your thoughts? Should we move forward with any of the above as outlined? Please let me know if you have questions/suggestions/comments throughout.

gwaybio:
Thanks @d33bs for catching me up!

> What are your thoughts? Should we move forward with any of the above as outlined?

No. I feel that because of limited bandwidth, all of our efforts should be placed toward pycytominer-transform at the moment. I feel that once we have transform, we can start to deprecate SingleCells functionality, which will invariably make these short term patches a waste of time.

However, if you sense that any of these patches might be really quick, then let's do them! There is an unfortunate time trade-off here, and we need to be really thoughtful about were to allocate attention.

I'll close now and we'll use a short part of our meeting this morning to discuss any of these short-term pycytominer targets. Thanks again!

PR #238, opened 2022-11-03
Change drop_outlier default to 500

### LINKED ISSUE #237: Questions about drop_outlier_features
I was looking at some code @fefossa is writing, and came across `drop_outlier_features`, which was added in #62 but somehow escaped my notice (it isn't on in the profiling template, for example). This is potentially super useful to us, but I was wondering about the default value (15). I know I'm asking about decisions made in December 2019, a far simpler time, but do you recall how this value was chosen? It doesn't appear at first pass to be documented, and this seems to me to be WAY too low - we routinely see features we believe in above that absolute value. Fernanda is going to check on some profiles with and without that feature reduction step added to get a sense of in her data how much effect it is having, but it is possibly significant.

I know the parameter is configurable, so we can override it if we'd like, but in general I assume we want defaults to be broadly applicable for the vast set of use cases, and I don't know that this one is.

bethac07:

# Description
Resolves #237 by changing the drop_outliers default cutoff from 15 to 500, for reasons more thoroughly detailed in that issue. 

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -142,7 +142,7 @@ def count_na_features(population_df, features):
 
 
 def drop_outlier_features(
-    population_df, features="infer", samples="all", outlier_cutoff=15
+    population_df, features="infer", samples="all", outlier_cutoff=500
 ):
     """Exclude a feature if its min or max absolute value is greater than the threshold.
 
@@ -154,7 +154,7 @@ def drop_outlier_features(
         Features present in the population dataframe. If "infer", then assume Cell Painting features are those that start with "Cells_", "Nuclei_", or "Cytoplasm_"
     samples : list of str or str, default "all"
         Samples to perform the operation on
-    outlier_cutoff : int or float, default 15
+    outlier_cutoff : int or float, default 500
         Threshold to remove features if absolute values is greater
 
     Returns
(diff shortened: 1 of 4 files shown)


pycytominer/cyto_utils/features.py
```diff
@@ -154,7 +154,7 @@ def drop_outlier_features(
         Features present in the population dataframe. If "infer", then assume Cell Painting features are those that start with "Cells_", "Nuclei_", or "Cytoplasm_"
     samples : list of str or str, default "all"
         Samples to perform the operation on
-    outlier_cutoff : int or float, default 15
+    outlier_cutoff : int or float, default 500
```
gwaybio, line 157:
```suggestion
    outlier_cutoff : int or float, default 500
    see https://github.com/cytomining/pycytominer/issues/237 for details.
```

Adding this explicit pointer for potential future reference

pycytominer/tests/test_cyto_utils/test_feature_drop_outlier.py
```diff
@@ -23,29 +23,34 @@
...
 
 def test_outlier_default():
     result = drop_outlier_features(data_df)
-    expected_result = ["Cells_zz", "Nuclei_z"]
+    expected_result = ["Cells_zz"]
     assert sorted(result) == sorted(expected_result)
 
 
-def test_outlier_high_cutoff():
+def test_outlier_30_cutoff():
```
gwaybio, line 37: This test and `test_outlier_default()` seem essentially the same to me. We can probably delete this one, especially given the addition of `test_outlier_15_cutoff()`

gwaybio (approved):
LGTM, a couple suggestions.

@niranjchandrasekaran - would you like to take a look?

bethac07:
@gwaybio Addressed your comments, once the tests rerun should we wait for Niranj to check or just pull?

Issue #240, opened 2022-11-13
Annotated Drops Metadata_Well_Position column when the Platemap and Profile dataset use the same name


kenibrewer:
After running pycytominer.annotate on a platemap and profiles_df that already had matching well_position column names, I noticed that the Metadata_Well column disappeared. 

I tracked it down to this line of code: https://github.com/cytomining/pycytominer/blob/8e5d27f592b10b10a73c55b6335e30e1c6c492fd/pycytominer/annotate.py#L82

This doesn't take into account the possibility that the two columns might be the same when dropping one. I'll submit a pull request implementing this fix.

kenibrewer:
Below is an example code fragment modified from the test/test_annotate.py file that illustrates this problem. This code fails if the PLATEMAP_WELL_COLUMN is set to 'Metadata_Well'. 

```{python}
import pathlib
import random
import tempfile

import pandas as pd
from pycytominer.annotate import annotate

random.seed(123)

# Get temporary directory
TMPDIR = tempfile.gettempdir()

# Setup a testing file
OUTPUT_FILE = pathlib.Path(f"{TMPDIR}/test.csv")

PLATEMAP_WELL_COLUMN = "Metadata_well_position"

# Build data to use in tests
DATA_DF = pd.concat(
    [
        pd.DataFrame(
            {"Metadata_Well": ["A01", "A02", "A03"], "x": [1, 3, 8], "y": [5, 3, 1]}
        ),
        pd.DataFrame(
            {"Metadata_Well": ["B01", "B02", "B03"], "x": [1, 3, 5], "y": [8, 3, 1]}
        ),
    ]
).reset_index(drop=True)

PLATEMAP_DF = pd.DataFrame(
    {
        PLATEMAP_WELL_COLUMN: ["A01", "A02", "A03", "B01", "B02", "B03"],
        "gene": ["x", "y", "z"] * 2,
    }
).reset_index(drop=True)


def test_annotate():

    # create expected result prior to annotate to distinguish modifications
    # performed by annotate to provided dataframes.
    expected_result = (
        PLATEMAP_DF.merge(DATA_DF, left_on=PLATEMAP_WELL_COLUMN, right_on="Metadata_Well")
        .rename(columns={"gene": "Metadata_gene"})
        .drop(PLATEMAP_WELL_COLUMN, axis="columns")
    )

    result = annotate(
        profiles=DATA_DF,
        platemap=PLATEMAP_DF,
[...]

PR #241, opened 2022-11-13
Added check before dropping well column

kenibrewer:
# Description

Simple bugfix for #240  where a missing check for identical column names results in an inappropriately dropped column.

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/load.py
+++ b/pycytominer/cyto_utils/load.py
@@ -75,7 +75,7 @@ def load_platemap(platemap, add_metadata_id=True):
         except FileNotFoundError:
             raise FileNotFoundError(f"{platemap} platemap file not found")
     else:
-        # Setting platemap to a copy to prevent column name changes from back-propogating
+        # Setting platemap to a copy to prevent column name changes from back-propagating
         platemap = platemap.copy()
 
     if add_metadata_id:
--- a/pycytominer/cyto_utils/load.py
+++ b/pycytominer/cyto_utils/load.py
@@ -74,6 +74,9 @@ def load_platemap(platemap, add_metadata_id=True):
             platemap = pd.read_csv(platemap, sep=delim)
         except FileNotFoundError:
             raise FileNotFoundError(f"{platemap} platemap file not found")
+    else:
+        # Setting platemap to a copy to prevent column name changes from back-propogating
+        platemap = platemap.copy()
 
     if add_metadata_id:
         platemap.columns = [
--- a/pycytominer/tests/test_annotate.py
+++ b/pycytominer/tests/test_annotate.py
@@ -37,7 +37,6 @@ def test_annotate():
 
     # create expected result prior to annotate to distinguish modifications
     # performed by annotate to provided dataframes.
-
     expected_result = (
         PLATEMAP_DF.merge(DATA_DF, left_on="well_position", right_on="Metadata_Well")
         .rename(columns={"gene": "Metadata_gene"})
@@ -56,7 +55,6 @@ def test_annotate():
 def test_annotate_platemap_naming():
 
     # Test annotate with the same column name in platemap and data.
-
     platemap_df = PLATEMAP_DF.copy().rename(columns={"well_position": "Metadata_Well"})
 
     expected_result = platemap_df.merge(
--- a/pycytominer/tests/test_annotate.py
+++ b/pycytominer/tests/test_annotate.py
@@ -46,11 +46,33 @@ def test_annotate():
     result = annotate(
         profiles=DATA_DF,
         platemap=PLATEMAP_DF,
-        join_on=["Metadata_well_position", "Metadata_Well"],
+        join_on=["well_position", "Metadata_Well"],
     )
 
     pd.testing.assert_frame_equal(result, expected_result)
 
+
+def test_annotate_platemap_naming():
+    
+    # Test annotate with the same column name in platemap and data.
+    
+        # create expected result prior to annotate to distinguish modifications
+    # performed by annotate to provided dataframes.
+    expected_result = (
+        PLATEMAP_DF.copy()
+        .rename(columns={"well_position":"well"})
+        .merge(DATA_DF, left_on="well", right_on="Metadata_Well")
+        .rename(columns={"gene": "Metadata_gene"})
+        .drop("well", axis="columns")
+    )
+
+    result = annotate(
+        profiles=DATA_DF,
+        platemap=PLATEMAP_DF.rename(columns={"well_position":"Well"}),
+        join_on=["Metadata_Well", "Metadata_Well"],
+    )
+
+    pd.testing.assert_frame_equal(result, expected_result)
 
 def test_annotate_output():
     annotate(
@@ -93,3 +115,7 @@ def test_annotate_output_compress():
     )
     expected_result = pd.read_csv(compress_file)
     pd.testing.assert_frame_equal(result, expected_result)
+
+
+if __name__ == "__main__":
+    test_annotate_platemap_naming()
--- a/pycytominer/tests/test_annotate.py
+++ b/pycytominer/tests/test_annotate.py
@@ -37,16 +37,19 @@ def test_annotate():
 
     # create expected result prior to annotate to distinguish modifications
     # performed by annotate to provided dataframes.
+    platemap_df = PLATEMAP_DF.copy()
+    data_df = DATA_DF.copy()
+    
     expected_result = (
-        PLATEMAP_DF.merge(DATA_DF, left_on="well_position", right_on="Metadata_Well")
+        platemap_df.merge(data_df, left_on="well_position", right_on="Metadata_Well")
         .rename(columns={"gene": "Metadata_gene"})
         .drop("well_position", axis="columns")
     )
 
     result = annotate(
-        profiles=DATA_DF,
-        platemap=PLATEMAP_DF,
-        join_on=["well_position", "Metadata_Well"],
+        profiles=data_df,
+        platemap=platemap_df,
+        join_on=["Metadata_well_position", "Metadata_Well"],
     )
 
     pd.testing.assert_frame_equal(result, expected_result)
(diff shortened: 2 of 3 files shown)


gwaybio:
Thanks for your contribution @kenibrewer - can you add a small test that would have failed the previous functionality that this contribution now fixes?

gwaybio:
Actually, maybe the simpler solution would be to drop the expression `.drop(join_on[0], axis="columns"`

But, then we run into the issue with auto suffixes changing both column names. We'd have to add something like: `suffixes=("_platemap", "")`

What do you think?

kenibrewer:
@gwaybio I added an example to the related issue page that illustrates the problem. I also added a test_case to test_annotate.py that crashes with the old code and works with the new code. In the process, I also discovered that a failure to work on fresh dataframe copies in individual tests was causing operations in one test to affect the others. So I fixed that too.

> Actually, maybe the simpler solution would be to drop the expression `.drop(join_on[0], axis="columns"`
> 
> But, then we run into the issue when `join_on[0] == join_on[1]` with auto suffixes changing both column names. We'd have to add something like: `suffixes=("_platemap", "")`
> 
> What do you think?

I think there is value to eliminating redundant information by having dropping the platemap version of the well column name. So I think we should still drop the column when appropriate. However, even if we drop the expression `.drop(join_on[0], axis="columns"`, we won't get any additional columns with auto-suffixes than we would anyways. 

Addressing the auto_suffix problem separately, I actually do really like the idea of adding the `suffixes=("_platemap", "")` parameter regardless. An additional nice feature would be if after the merge any new '_platemap' columns are checked against their matching data column and are dropped if they are identical.

pycytominer/tests/test_annotate.py
```diff
@@ -37,34 +37,63 @@ def test_annotate():
 
     # create expected result prior to annotate to distinguish modifications
     # performed by annotate to provided dataframes.
+    platemap_df = PLATEMAP_DF.copy()
```
gwaybio, line 40: I'm a bit confused why this is necessary - our tests were passing before.

I think you can remove the copies and revert the changed lines back to the original data.

pycytominer/tests/test_annotate.py
```diff
@@ -37,34 +37,63 @@ def test_annotate():
...
 
     pd.testing.assert_frame_equal(result, expected_result)
 
 
+def test_annotate_platemap_naming():
+    
+    # Test annotate with the same column name in platemap and data.
+    
+    platemap_df = PLATEMAP_DF.copy().rename(columns={"well_position":"well"})
+    data_df = DATA_DF.copy()
```
gwaybio, line 63: delete this line please

pycytominer/tests/test_annotate.py
```diff
@@ -37,34 +37,63 @@ def test_annotate():
...
     )
 
     pd.testing.assert_frame_equal(result, expected_result)
 
 
+def test_annotate_platemap_naming():
+    
+    # Test annotate with the same column name in platemap and data.
+    
+    platemap_df = PLATEMAP_DF.copy().rename(columns={"well_position":"well"})
```
gwaybio, line 62:
```suggestion
    platemap_df = PLATEMAP_DF.copy(deep=True)
    platemap_df = platemap_df.rename(columns={"well_position":"well"})
```

To confirm that any modifications that you perform on the copy doesn't bleed through to the original `PLATEMAP_DF`

pycytominer/tests/test_annotate.py
```diff
@@ -37,34 +37,63 @@ def test_annotate():
...
+        platemap_df
+        .merge(data_df, left_on="well", right_on="Metadata_Well")
+        .rename(columns={"gene": "Metadata_gene"})
+        .drop("well", axis="columns")
+    )
+
+    result = annotate(
+        profiles=data_df,
+        platemap=platemap_df,
+        join_on=["Metadata_well", "Metadata_Well"],
```
gwaybio, line 75: These two entries are different - did you mean to make them the same? (case sensitive)

pycytominer/tests/test_annotate.py
```diff
@@ -37,34 +37,63 @@ def test_annotate():
...
+    )
+
+    pd.testing.assert_frame_equal(result, expected_result)
+
 def test_annotate_output():
+
```
gwaybio, line 81: I am confused about why `test_annotate_output` needed to be modified.

pycytominer/tests/test_annotate.py
```diff
@@ -74,20 +103,24 @@ def test_annotate_output():
 
 
 def test_annotate_output_compress():
+
```
gwaybio, line 106: Also, why did this need to be modified? (`test_annotate_output_compress`)

The tests were passing previously

gwaybio (changes requested):
A couple discussion questions and required changes prior to review. Thanks again for your contribution!

pycytominer/tests/test_annotate.py
kenibrewer: In the live version of the code, tests are passing inappropriately because PLATEMAP_DF columns being renamed during the execution of `test_annotate()` is cancelled out by the inclusion of incorrect arguments in `test_annotate_output()` and `test_annotate_output_compress()`

### PLATEMAP_DF column names are modified during`test_annotate()`

Because `test_annotate()` uses the default argument `add_metadata_id_to_platemap=True`, the platemap column names are being modified during execution. This modification occurs [here](https://github.com/cytomining/pycytominer/blob/ea9873ddd3b7d9b7072cf7fc55d582012cf6d315/pycytominer/annotate.py#L78) in the load_platemap function. Calling `platemap_df = PLATEMAP_DF.copy()` and using the copied dataframe is necessary to prevent the column renaming from influencing other tests. 

### Incorrect Inputs
 Both `test_annotate_output()` and `test_annotate_output_compress()` contain the argument `add_metadata_id_to_platemap=False`. This means that for both of them, the lines `join_on=["Metadata_well_position", "Metadata_Well"]` should be changed to `join_on=["well_position", "Metadata_Well"]`  because "well_position" is the appropriate column name. Correcting these incorrect arguments will cause those tests to fail unless we modify `test_annotate()` with the line of code being discussed.. 





The reason that these tests were

pycytominer/tests/test_annotate.py
kenibrewer: Because I am only modifying column names of PLATEMAP_DF for this new test and am not changing any of the underlying data, `copy(deep=True)` is not necessary. 

Here is chart that I believe summarizes when which type of copy is required: 
|                        | No changes to Columns | Changes to Columns |
|------------------------|-----------------------|--------------------|
| **No changes to Data** | No copy needed        | .copy(deep=False)  |
| **Changes to Data**    | .copy(deep=True)      | .copy(deep=True)   |

pycytominer/tests/test_annotate.py
kenibrewer: More detailed explanation above, but in short `test_annotate_output` is operating on a modified version of `PLATEMAP_DF` and is only passing because it is being given the wrong arguments.

pycytominer/tests/test_annotate.py
kenibrewer: More detailed explanation above, but in short `test_annotate_output_compress` is operating on a version of `PLATEMAP_DF` that has renamed columns and is only passing because it is being given the wrong arguments.

pycytominer/tests/test_annotate.py
kenibrewer: The use of `platemap_df = PLATEMAP_DF.copy()` is not strictly necessary, but I think it maybe worth adopting that structure in all of the tests for consistency sake.

pycytominer/tests/test_annotate.py
kenibrewer: As mentioned above, the use of `platemap_df = PLATEMAP_DF.copy()` is not strictly necessary for this test, but I think it maybe worth adopting that structure in all of the tests for consistency sake.

pycytominer/tests/test_annotate.py
kenibrewer: There may be some value to using `copy(deep=True)` even though it isn't necessary in this case as a model for future test writers who want to need to operate on modified versions of `PLATEMAP_DF`.

pycytominer/tests/test_annotate.py
gwaybio: Amazing find @kenibrewer ! Thank you for bringing this to our attention.

I think this issue is worth resolving at the source. Can you add a data frame `.copy(deep=True)` to the `load_platemap()` function [here](https://github.com/cytomining/pycytominer/blob/ea9873ddd3b7d9b7072cf7fc55d582012cf6d315/pycytominer/cyto_utils/load.py#L78)?

Maybe something like:

```python
    if add_metadata_id:
        platemap_df = platemap.copy(deep=True)
        platemap_df.columns = [
            f"Metadata_{x}" if not x.startswith("Metadata_") else x
            for x in platemap_df.columns
        ]
    return platemap_df
```

(you'll need to modify `platemap_df = pd.read_csv(platemap, sep=delim)` a couple lines above as well.

I think that it's worth fixing this bug than fixing tests that work around the bug. (you'll also likely have to update tests that were inappropriately passing!)

Thanks again!

gwaybio:
@kenibrewer - please correct me if I'm wrong, but I think we're waiting on your reply to [this comment](https://github.com/cytomining/pycytominer/pull/241#discussion_r1028571680) prior to moving forward with merging this PR. Thanks again!

pycytominer/tests/test_annotate.py
kenibrewer: Hi @gwaybio ,
I think you are correct about it being better to solve the problem by addressing the root cause of the bug which is the failure to use `.copy(deep=True)`.  I'll make the appropriate adjustments.

kenibrewer:
Hi @gwaybio, 
You are correct that you are waiting on me to submit some additional changes prior to merging. I had some other priorities the past few weeks but have some time to work on this now. 
Related to this bug. I've think I've discovered some other bugs related to missing `.copy(deep=True)` in other pycytominer functions. I'll submit those as a separate issue though.

kenibrewer:
Revisiting your [suggestion about adding suffixes](https://github.com/cytomining/pycytominer/pull/241#issuecomment-1315459163), I think it would be good to add `suffixes = ["", "platemap"]` to the platemap merge and `suffixes = ["", "external"] to the external metadata merge. Both of those changes would result in much more readable and predictable outputs than the default "_x" and "_y" suffixes. 

The one problem with this change, however, is that it would likely break pipelines/code (including a few of my own projects), that already include logic around renaming/handling the "_x" and "_y" columns that are generated by the current codebase. I still think the change is worthwhile, but I'm not sure how you handle potentially breaking changes.

pycytominer/tests/test_annotate.py
kenibrewer: Thanks. Fixed.

pycytominer/tests/test_annotate.py
kenibrewer: Done

gwaybio:
Thanks @kenibrewer - thanks for following up and I'm excited you're able to work on this PR again. I'll respond to a couple of your thoughts and then provide a code review.

> I've think I've discovered some other bugs related to missing .copy(deep=True) in other pycytominer functions. I'll submit those as a separate issue though.

Fantastic - thanks! Once you create a new issue, please link this PR

> I think it would be good to add suffixes = ["", "platemap"] to the platemap merge and suffixes = ["", "external"] to the external metadata merge.

Sounds good. My sense is that this change belongs in a separate PR. Let's focus on dropping the well column (and fixing any immediate bugs that arise from this closer investigation) here, and save this potentially breaking change for a subsequent PR.

> I'm not sure how you handle potentially breaking changes.

We haven't had a hard and fast policy, but we typically mint a new release version which will kickoff a new pypi and conda build. If you're using a specific pycytominer version (or github hash) in your previous pipelines, then this should have no impact on past work other than needing to modify code for future projects if new pycytominer versions are good enough to warrant a version upgrade.

pycytominer/cyto_utils/load.py
```diff
@@ -74,6 +74,9 @@ def load_platemap(platemap, add_metadata_id=True):
             platemap = pd.read_csv(platemap, sep=delim)
         except FileNotFoundError:
             raise FileNotFoundError(f"{platemap} platemap file not found")
+    else:
+        # Setting platemap to a copy to prevent column name changes from back-propogating
```
gwaybio, line 78:
```suggestion
        # Setting platemap to a copy to prevent column name changes from back-propagating
```

pycytominer/tests/test_annotate.py
```diff
@@ -52,19 +52,38 @@ def test_annotate():
     pd.testing.assert_frame_equal(result, expected_result)
 
 
+def test_annotate_platemap_naming():
+
+    # Test annotate with the same column name in platemap and data.
+    platemap_df = PLATEMAP_DF.copy().rename(columns={"well_position": "Metadata_Well"})
```
gwaybio, line 58: this is really minor, but can you name `platemap_df` something else more descriptive? (just to increase readability)

gwaybio (commented):
Looks good! I'll kick off tests now. Please review my minor suggestions and then, once the tests pass, I'll check in with you and merge.

Thanks again for your contribution!

pycytominer/cyto_utils/load.py
kenibrewer: Fixed

pycytominer/tests/test_annotate.py
kenibrewer: Fixed.

gwaybio:
Looks like the tests will pass - are we set to merge this in @kenibrewer ?

kenibrewer:
@gwaybio Yep. Ready for merge. 🚀

gwaybio (approved):
🚀

gwaybio:
Thanks again @kenibrewer 

When you get a chance, please create new issues in https://github.com/cytomining/pycytominer/issues. I believe they involved 1) `copy()` based bugs and 2) specific merge suffixes. I don't want to lose track of these

PR #243, opened 2022-11-25
avoid re-validation for excluded features

johnarevalo:
If an `operation` excluded a column, such column should not be validated by further `operation`s. This will reduce the computation time. It will benefit list of operations sorted by computational cost (ascending).

- [X] Enhancement (speed).
- [ ] This change requires a documentation update. Maybe explain the benefits of sorting operations by computational cost.

# Checklist

### DIFF
--- a/pycytominer/tests/test_feature_select.py
+++ b/pycytominer/tests/test_feature_select.py
@@ -326,7 +326,7 @@ def test_feature_select_correlation_threshold():
         features=data_cor_thresh_na_df.columns.tolist(),
         operation=["drop_na_columns", "correlation_threshold"],
     )
-    expected_result = data_df.drop(["z", "x"], axis="columns")
+    expected_result = data_df.drop(["z", "y"], axis="columns")
     pd.testing.assert_frame_equal(result, expected_result)
 
 
--- a/pycytominer/feature_select.py
+++ b/pycytominer/feature_select.py
@@ -176,6 +176,7 @@ def feature_select(
                 noise_removal_stdev_cutoff=noise_removal_stdev_cutoff,
             )
         excluded_features += exclude
+        features = [feat for feat in features if feat not in excluded_features]
 
     excluded_features = list(set(excluded_features))
 
--- a/pycytominer/feature_select.py
+++ b/pycytominer/feature_select.py
@@ -1,10 +1,6 @@
 """
 Select features to use in downstream analysis based on specified selection method
 """
-
-import os
-import pandas as pd
-
 from pycytominer.operations import (
     correlation_threshold,
     variance_threshold,


gwaybio:
@johnarevalo - is this ready for me to take a look? If so, please tag me

johnarevalo:
Thanks for your reply @gwaybio, yes this is ready for your input.

pycytominer/tests/test_feature_select.py
```diff
@@ -326,7 +326,7 @@ def test_feature_select_correlation_threshold():
         features=data_cor_thresh_na_df.columns.tolist(),
         operation=["drop_na_columns", "correlation_threshold"],
     )
-    expected_result = data_df.drop(["z", "x"], axis="columns")
+    expected_result = data_df.drop(["z", "y"], axis="columns")
```
gwaybio, line 329: This is a bit concerning - why did this change?

gwaybio (changes requested):
I am excited about this change! 

One comment worth discussion in the test but then I think this is straightforward to merge

pycytominer/tests/test_feature_select.py
johnarevalo: After `"drop_na_columns"` operation removes the `z` column, `x` is no longer the feature with the largest sum of `corr`:
```python
In [1]: data_df = pd.DataFrame(
    ...:     {
    ...:         "x": [1, 3, 8, 5, 2, 2],
    ...:         "y": [1, 2, 8, 5, 2, 1],
    ...:         "z": [9, 3, 8, 9, 2, 9],
    ...:         "zz": [0, -3, 8, 9, 6, 9],
    ...:     }
    ...: ).reset_index(drop=True)

In [2]: data_df.iloc[0, 2] = np.nan

In [3]: data_df.corr(method='pearson').sum().sort_values()
Out[3]:
z     2.584062
zz    2.591094
y     2.851580
x     2.873417
dtype: float64

In [45]: data_df.dropna(axis=1).corr(method='pearson').sum().sort_values()
Out[45]:
zz    1.881769
x     2.414736
y     2.435524
dtype: float64
```

I believe this behavior is consistent. It's worth mentioning the `feature_select` method is now sensitive to the order of the `operation` list. i.e. `feature_select(df, operation=["drop_na_columns", "correlation_threshold"]) != feature_select(df, operation=["correlation_threshold", "drop_na_columns"])`

gwaybio: Got it! Sounds good

> I believe this behavior is consistent. It's worth mentioning the feature_select method is now sensitive to the order of the operation list

This is interesting! I wonder if it's worth adding a function argument which defaults to existing (but slower) behavior. Until we understand the impact of the function ordering on profiles, it's probably best to not surprise a user in a later version that's used to using the old approach.

Once we do understand the impact (likely minimal) then we can make the default this speed-up.

What do you think? (If you think this is unnecessarily complicating things, then we can merge this PR as is!)

johnarevalo: I agree the impact of this change is minimal for users. The main concern is **exact** reproducibility, this PR may select very similar but not identical feature sets as before. 

If exact reproducibility is not a issue, then I'd suggest to merge it as is.

gwaybio: I think exact reproducibility is a big issue. However, those that would agree with me likely have procedures in place to confirm exact reproducibility external to the pycytominer codebase.

Alright, I'll merge. Thanks again @johnarevalo

gwaybio (approved):

PR #246, opened 2022-12-07
Allow column subsetting at the time of loading

### LINKED ISSUE #245: Load only some columns during compartment loading in SingleCells
Right now, it is possible to subsample the data afterward to certain rows, and after doing something like merge_single_cells you can subset to certain columns, but there is no way to subset before loading in everything. This is a hassle when data is large - @rsenft1 is running into it in some of her projects. If we were to add this functionality, would a PR be welcomed?

bethac07:

# Description

Resolves #245 

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -432,7 +432,7 @@ class SingleCells(object):
 
         return meta_cols, feat_cols
 
-    def load_compartment(self, compartment, features):
+    def load_compartment(self, compartment):
         """Creates the compartment dataframe.
 
         Parameters
@@ -449,8 +449,8 @@ class SingleCells(object):
         # Get data useful to pre-alloc memory
         num_cells = self.count_sql_table_rows(compartment)
         col_names = self.get_sql_table_col_names(compartment)
-        if features != "infer":  # allow to get only some features
-            col_names = [x for x in col_names if x in features]
+        if self.features != "infer":  # allow to get only some features
+            col_names = [x for x in col_names if x in self.features]
         meta_cols, feat_cols = self.split_columns_into_classes(col_names)
         num_meta, num_feats = len(meta_cols), len(feat_cols)
 
@@ -475,7 +475,6 @@ class SingleCells(object):
     def aggregate_compartment(
         self,
         compartment,
-        features,
         compute_subsample=False,
         compute_counts=False,
         add_image_features=False,
@@ -520,7 +519,6 @@ class SingleCells(object):
         object_dfs = []
         for compartment_df in self._compartment_df_generator(
             compartment=compartment,
-            features=features,
             n_aggregation_memory_strata=n_aggregation_memory_strata,
         ):
 
@@ -586,7 +584,6 @@ class SingleCells(object):
     def _compartment_df_generator(
         self,
         compartment,
-        features,
         n_aggregation_memory_strata=1,
     ):
         """A generator function that returns chunks of the entire compartment
@@ -623,9 +620,8 @@ class SingleCells(object):
             con=self.conn,
         )
         all_columns = compartment_row1.columns
-        if features != "infer":
-            all_columns = [x for x in all_columns if x in features]
-            cols = ", ".join(all_columns)
+        if self.features != "infer":
+            all_columns = [x for x in all_columns if x in self.features]
 
         typeof_str = ", ".join([f"typeof({x})" for x in all_columns])
         compartment_dtypes = pd.read_sql(
@@ -728,9 +724,7 @@ class SingleCells(object):
                 ]
 
                 if isinstance(sc_df, str):
-                    sc_df = self.load_compartment(
-                        compartment=left_compartment, features=self.features
-                    )
+                    sc_df = self.load_compartment(compartment=left_compartment)
 
                     if compute_subsample:
                         # Sample cells proportionally by self.strata
@@ -745,9 +739,7 @@ class SingleCells(object):
                         ).reindex(sc_df.columns, axis="columns")
 
                     sc_df = sc_df.merge(
-                        self.load_compartment(
-                            compartment=right_compartment, features=self.features
-                        ),
+                        self.load_compartment(compartment=right_compartment),
                         left_on=self.merge_cols + [left_link_col],
                         right_on=self.merge_cols + [right_link_col],
                         suffixes=merge_suffix,
@@ -755,9 +747,7 @@ class SingleCells(object):
 
                 else:
                     sc_df = sc_df.merge(
-                        self.load_compartment(
-                            compartment=right_compartment, features=self.features
-                        ),
+                        self.load_compartment(compartment=right_compartment),
                         left_on=self.merge_cols + [left_link_col],
                         right_on=self.merge_cols + [right_link_col],
                         suffixes=merge_suffix,
@@ -872,7 +862,6 @@ class SingleCells(object):
             if compartment_idx == 0:
                 aggregated = self.aggregate_compartment(
                     compartment=compartment,
-                    features=self.features,
                     compute_subsample=compute_subsample,
                     compute_counts=True,
                     add_image_features=self.add_image_features,
@@ -882,7 +871,6 @@ class SingleCells(object):
                 aggregated = aggregated.merge(
                     self.aggregate_compartment(
                         compartment=compartment,
-                        features=self.features,
                         n_aggregation_memory_strata=n_aggregation_memory_strata,
                     ),
                     on=self.strata,
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -625,7 +625,7 @@ class SingleCells(object):
         all_columns = compartment_row1.columns
         if features != "infer":
             all_columns = [x for x in all_columns if x in features]
-            col = ", ".join(all_columns)
+            cols = ", ".join(all_columns)
 
         typeof_str = ", ".join([f"typeof({x})" for x in all_columns])
         compartment_dtypes = pd.read_sql(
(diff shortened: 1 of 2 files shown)


bethac07:
Before I go nuts changing the function definition strings for the functions where features is now passed explicitly, please let me know if that is indeed the way it would be preferred to do it - my current thinking was to keep high-level functionality as is by using self.features, but to make the functionality more explicit/obvious in lower level functions by passing `features` explicitly. Obviously self.features can be used everywhere if we want minimum-breaky-ness, but that was my thought process on why some things are done one way and others another; happy to have input on preferred way to pass that info around.

pycytominer/cyto_utils/cells.py
```diff
@@ -613,7 +622,12 @@ def _compartment_df_generator(
             sql=f"select {cols} from {compartment} limit 1",
             con=self.conn,
         )
-        typeof_str = ", ".join([f"typeof({x})" for x in compartment_row1.columns])
+        all_columns = compartment_row1.columns
+        if features != "infer":
+            all_columns = [x for x in all_columns if x in features]
+            col = ", ".join(all_columns)
```
gwaybio, line 628: I am not sure where this variable is used.

pycytominer/cyto_utils/cells.py
```diff
@@ -513,6 +520,7 @@ def aggregate_compartment(
         object_dfs = []
         for compartment_df in self._compartment_df_generator(
             compartment=compartment,
+            features=features,
```
gwaybio, line 523: I'm not sure where this is used?

pycytominer/cyto_utils/cells.py
```diff
@@ -714,7 +728,9 @@ def merge_single_cells(
                 ]
 
                 if isinstance(sc_df, str):
-                    sc_df = self.load_compartment(compartment=left_compartment)
+                    sc_df = self.load_compartment(
+                        compartment=left_compartment, features=self.features
```
gwaybio, line 732: should this be `features=features` instead of `features=self.features`?

Otherwise this doesn't seem like your changes have an effect in `merge_single_cells` - is this intended?

gwaybio: The same comment goes for the changes added below in this function as well.

pycytominer/cyto_utils/cells.py
```diff
@@ -469,6 +475,7 @@ def load_compartment(self, compartment):
     def aggregate_compartment(
         self,
         compartment,
+        features,
```
gwaybio, line 478: maybe worth making `features='infer'` as default to retain previous behavior and not require additional arguments if someone uses a new version?

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -265,15 +278,15 @@ def test_SingleCells_count():
 
 
 def test_load_compartment():
-    loaded_compartment_df = AP.load_compartment(compartment="cells")
+    loaded_compartment_df = AP.load_compartment(compartment="cells", features="infer")
```
gwaybio, line 281: if you add `features="infer"` as default, you can probably delete this explicit variable naming to retain default testing configuration

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -417,6 +432,13 @@ def test_merge_single_cells():
     )
 
 
+def test_merge_single_cells_subset():
+    sc_merged_df = AP_SUBSET.merge_single_cells()
+    assert (sc_merged_df.shape[1]) == 13
+    non_meta_cols = [x for x in sc_merged_df.columns if "Metadata" not in x]
+    assert len(non_meta_cols) == len([x for x in non_meta_cols if x in SUBSET_FEATURES])
```
gwaybio, line 439: 👍

gwaybio (changes requested):
I think there are a couple comments in here that need addressing, some I think are optional

pycytominer/cyto_utils/cells.py
bethac07: Should be clear now after the fix, lmk if not

pycytominer/cyto_utils/cells.py
bethac07: Good catch! Should be clear now what's happening

bethac07:
@gwaybio Thanks for taking a look!

Re: where `features` and where `self.features` - as I said, my thinking was to use `self.features` on higher level functions that people are more likely to be calling the function directly, and then be using `features` on lower level functions to make it more obvious that features is a thing that is being read and acted on there. Do you want it `self.features` everywhere? If so, that's an easy fix.

pycytominer/cyto_utils/cells.py
bethac07: My assumption was that `features` was intended to be passed when defining `SingleCells` (ie `database = SingleCells(f"sqlite:///{backend_file}",features=desired_features)`; in that case, yes, merge_single_cells DOES pick up these changes). 

Do we ever expect/want the user to change what the desired features are AFTER instantiating the class object? I was thinking no, but you tell me.

pycytominer/cyto_utils/cells.py
bethac07: See other comment, but my assumption when writing this was that we don't want the user changing the features after initializing SingleCells; if so, we don't want a default here.

pycytominer/cyto_utils/cells.py
```diff
@@ -613,7 +622,12 @@ def _compartment_df_generator(
             sql=f"select {cols} from {compartment} limit 1",
             con=self.conn,
         )
-        typeof_str = ", ".join([f"typeof({x})" for x in compartment_row1.columns])
+        all_columns = compartment_row1.columns
+        if features != "infer":
+            all_columns = [x for x in all_columns if x in features]
+            cols = ", ".join(all_columns)
```
gwaybio, line 628: I still am not sure where `cols` is used. Do you mean to replace what happens sooner in line 620? (`cols = "*"`)?

Sorry if I'm being dense!

gwaybio (commented):
I'll add other comments in-line

pycytominer/cyto_utils/cells.py
gwaybio: > Do we ever expect/want the user to change what the desired features are AFTER instantiating the class object?

Hmm, I'm still a bit confused. What is your entry point to allow column subsetting?

If the entry point is upon class instantiation, then why do you need to add a `features` argument to `load_compartment()` ([here](https://github.com/cytomining/pycytominer/pull/246/files#diff-ed9097fc5b9747e8287afce4ba63d18724e3e88d7725fabda9b28be2e6ab81f2R452))?

(I agree that a user shouldn't need to change the features arg after instantiation)

pycytominer/cyto_utils/cells.py
gwaybio: I think it was clear even before the fix - sorry for overlooking!

pycytominer/cyto_utils/cells.py
gwaybio: > we don't want the user changing the features after initializing SingleCells

I agree, but then why do we need this argument in `aggregate_compartment` at all?

gwaybio:
> to make it more obvious that features is a thing that is being read and acted on there.

Ah! Maybe this is the reason why you're adding function arguments?

If so, then I encourage an explicit comment (or comments) rather than a code change.

pycytominer/cyto_utils/cells.py
bethac07: Nope, I am, I swear it made sense in my brain when I wrote it...

bethac07:
Great, that was the feedback I was looking for! Changes pushed

bethac07:
@gwaybio if you think sufficiently documented, I think it's ready to pull!

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -291,7 +304,9 @@ def test_sc_count_sql_table():
 def test_get_sql_table_col_names():
     # Iterate over initialized compartments
     for compartment in AP.compartments:
-        meta_cols, feat_cols = AP.get_sql_table_col_names(table=compartment)
+        meta_cols, feat_cols = AP.split_columns_into_classes(
```
gwaybio, line 307: rather than chaining these two function calls together, can you explicitly test both of them separately?

This will enable us to zero in on code changes that might break one but not the other.

e.g.

```python
col_name_result = AP.get_sql_table_col_names(table=compartment)
assert col_name_result == expected_result

meta_cols, feat_cols = AP.split_columns_into_classes(col_name_result)
# Then pass these vars into existing assertions
```

pycytominer/cyto_utils/cells.py
```diff
@@ -445,7 +449,10 @@ def load_compartment(self, compartment):
 
         # Get data useful to pre-alloc memory
         num_cells = self.count_sql_table_rows(compartment)
-        meta_cols, feat_cols = self.get_sql_table_col_names(compartment)
+        col_names = self.get_sql_table_col_names(compartment)
+        if self.features != "infer":  # allow to get only some features
+            col_names = [x for x in col_names if x in self.features]
+        meta_cols, feat_cols = self.split_columns_into_classes(col_names)
```
gwaybio, line 455:
```suggestion
        meta_cols, feat_cols = self.split_column_categories(col_names)
```

To avoid the name collision with pythonic "classes". (Maybe a different name is better than my proposal)

gwaybio (approved):
This looks great, elegant.

Two minor comments for you to address then I will merge if tests pass. Thanks!

pycytominer/tests/test_cyto_utils/test_cells.py
bethac07: Done!

gwaybio:
Alrighty, here we go!

Issue #227, opened 2022-08-19
Locate other performance bottlenecks, document, and add testing constraints


gwaybio:
Discussed in https://github.com/cytomining/pycytominer/issues/195#issuecomment-1217265110

Add Memory/Time Testing Constraints for .merge_single_cells() (or others! 🙂 ) (Something like [pytest-memray](https://github.com/bloomberg/pytest-memray) or similar): automate targeted performance tests to ensure operations within this project are usable by others as appropriate. Maybe this one spot was an outlier, but if there are other performance-heavy functions we could expand there as well.

jenna-tomkinson:
@d33bs Is this issue resolved?

Issue #198, opened 2022-05-21
Address SQLite Read Performance Issues


d33bs:
SQLite files used by this repo create performance challenges during runtime that may hinder or completely stall progress. This issue is dedicated to targeting SQLite read performance issues, including solutions which convert away from the format or may not use SQLite at all.

Issues which may be related or tied to this: 
- #195 
- #142 
- #117 
- #112
- #199

d33bs:
Hi @gwaygenomics - I'm noticing that within `SQ00014613.sqlite` ([from here](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672)), there are numeric columns which specify not-null for column attributes and which sometimes include 'nan' (textual) actual values. Examples may be found in that file within the `cytoplasm` table and column `Cytoplasm_Correlation_Costes_AGP_DNA` (there are many others with this issue). The following SQL shows this as well:

```sql
select ObjectNumber, 
    Cytoplasm_Correlation_Costes_AGP_DNA,
    typeof(Cytoplasm_Correlation_Costes_AGP_DNA) from Cytoplasm
where typeof(Cytoplasm_Correlation_Costes_AGP_DNA) != 'real';
```

SQLite may include column values that don't match the column specification. This could be one cause of the performance issues using Pandas and SQLAlchemy, but it also can impact alternatives. Because these columns are sometimes indicated as not-null, it means we may not use `NULL` unless we create a new column, copying the original.

Do you know of a way we could interpret these 'nan' values correctly for the purposes of pycytominer (for example, replace with 0, filter from results in database, etc)?

d33bs:
No worries and thank you @gwaygenomics - I'll take a look. Enjoy the conference!

d33bs:
Hi @gwaygenomics - I just wanted to follow up with some findings, open to any guidance or input you have along the way. I've spent some time investigating ways to speed up sqlite reads, taking note of the background you mentioned a little earlier. I don't yet have thoughts on solutions for the cytominer-database issue, as a brainstorm here it feels like `pandas.dataframe.to_sql()` or SQLAlchemy may have had a hand in this with built-in assumptions or type inferences.

The file I've focused on for this has so far been only SQ00014613.sqlite. For the below, I have the table `cytoplasm` column `Cytoplasm_Correlation_Costes_AGP_DNA` within SQ00014613.sqlite in mind, which includes mostly floating point number values in addition to a handful of 'nan' values.

[...]

PR #248, opened 2022-12-16
Add float_datatype parameter for SingleCells compartment load and merge performance flexibility

d33bs:
# Description

This change adds a float_datatype parameter for use with SingleCells compartment load and merge operations to assist with performance considerations (as cited within wayscience/cell-health-data#10 and other areas). In testing this locally using `np.float32` instead of `np.float64` I found ~45% reduction in peak memory consumption while running `merge_single_cells` on small testing datasets (this reduction may vary contingent on dataset).

This change is also related to #198 and #227 (if only to notate datatype considerations).

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -429,13 +429,17 @@ class SingleCells(object):
 
         return meta_cols, feat_cols
 
-    def load_compartment(self, compartment):
+    def load_compartment(self, compartment, float_datatype: type = np.float64):
         """Creates the compartment dataframe.
 
         Parameters
         ----------
         compartment : str
             The compartment to process.
+        float_datatype: type, default np.float64
+            Numpy floating point datatype to use for load_compartment and resulting
+            dataframes. Please note: using any besides np.float64 are experimentally
+            unverified.
 
         Returns
         -------
@@ -448,8 +452,8 @@ class SingleCells(object):
         meta_cols, feat_cols = self.get_sql_table_col_names(compartment)
         num_meta, num_feats = len(meta_cols), len(feat_cols)
 
-        # Use pre-allocated np.array for data
-        feats = np.empty(shape=(num_cells, num_feats), dtype=np.float64)
+        # Use pre-allocated np.array for feature data
+        feats = np.empty(shape=(num_cells, num_feats), dtype=float_datatype)
         # Use pre-allocated pd.DataFrame for metadata
         metas = pd.DataFrame(columns=meta_cols, index=range(num_cells))
 
@@ -661,6 +665,7 @@ class SingleCells(object):
         single_cell_normalize: bool = False,
         normalize_args: Optional[Dict] = None,
         platemap: Optional[Union[str, pd.DataFrame]] = None,
+        float_datatype: type = np.float64,
         **kwargs,
     ):
         """Given the linking columns, merge single cell data. Normalization is also supported.
@@ -681,6 +686,10 @@ class SingleCells(object):
             Additional arguments passed as input to pycytominer.normalize().
         platemap: str or pd.DataFrame, default None
             optional platemap filepath str or pd.DataFrame to be used with results via annotate
+        float_datatype: type, default np.float64
+            Numpy floating point datatype to use for load_compartment and resulting
+            dataframes. Please note: using any besides np.float64 are experimentally
+            unverified.
 
         Returns
         -------
@@ -714,7 +723,9 @@ class SingleCells(object):
                 ]
 
                 if isinstance(sc_df, str):
-                    sc_df = self.load_compartment(compartment=left_compartment)
+                    sc_df = self.load_compartment(
+                        compartment=left_compartment, float_datatype=float_datatype
+                    )
 
                     if compute_subsample:
                         # Sample cells proportionally by self.strata
@@ -729,7 +740,9 @@ class SingleCells(object):
                         ).reindex(sc_df.columns, axis="columns")
 
                     sc_df = sc_df.merge(
-                        self.load_compartment(compartment=right_compartment),
+                        self.load_compartment(
+                            compartment=right_compartment, float_datatype=float_datatype
+                        ),
                         left_on=self.merge_cols + [left_link_col],
                         right_on=self.merge_cols + [right_link_col],
                         suffixes=merge_suffix,
@@ -737,7 +750,9 @@ class SingleCells(object):
 
                 else:
                     sc_df = sc_df.merge(
-                        self.load_compartment(compartment=right_compartment),
+                        self.load_compartment(
+                            compartment=right_compartment, float_datatype=float_datatype
+                        ),
                         left_on=self.merge_cols + [left_link_col],
                         right_on=self.merge_cols + [right_link_col],
                         suffixes=merge_suffix,
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -753,7 +753,7 @@ class SingleCells(object):
                 else:
                     sc_df = sc_df.merge(
                         self.load_compartment(
-                            compartment=right_compartment, float_datatype=float_datatype
+                            compartment=right_compartment
                         ),
                         left_on=self.merge_cols + [left_link_col],
                         right_on=self.merge_cols + [right_link_col],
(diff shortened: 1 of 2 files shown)


pycytominer/cyto_utils/cells.py
```diff
@@ -429,13 +429,17 @@ def get_sql_table_col_names(self, table):
...
+    def load_compartment(self, compartment, float_datatype: type = np.float64):
         """Creates the compartment dataframe.
 
         Parameters
         ----------
         compartment : str
             The compartment to process.
+        float_datatype: type, default np.float64
+            Numpy floating point datatype to use for load_compartment and resulting
+            dataframes. Please note: using any besides np.float64 are experimentally
```
gwaybio, line 441: It might be worth noting other options one might use in this docstring and including what the anticipated benefit might be. (i.e. - why is this worth adding as an argument and why anyone would ever change it?)

pycytominer/cyto_utils/cells.py
```diff
@@ -429,13 +429,17 @@ def get_sql_table_col_names(self, table):
 
         return meta_cols, feat_cols
 
-    def load_compartment(self, compartment):
+    def load_compartment(self, compartment, float_datatype: type = np.float64):
```
gwaybio, line 432: I think most folks will use this after initializing the SingleCells class. So it's probably worth making this an init argument and using `self.single_cell_merge_float_datatype` throughout (note the rename - maybe there is something better to rename it to?)

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -280,6 +281,23 @@ def test_load_compartment():
         check_dtype=False,
     )
 
+    # test using non-default float_datatype
+    loaded_compartment_df = AP.load_compartment(
+        compartment="cells", float_datatype=np.float32
+    )
+    pd.testing.assert_frame_equal(
+        loaded_compartment_df,
+        CELLS_DF.astype(
```
gwaybio, line 290: This chaining is a bit hard to follow. Can you create a new df (`CELLS_float32_DF` or something), and make sure it is a deep copy? (see https://github.com/cytomining/pycytominer/pull/241#discussion_r1028571680)

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -280,6 +281,23 @@ def test_load_compartment():
...
+            # cast any float type columns to float32 for expected comparison
+            {
+                colname: np.float32
+                for colname in CELLS_DF.columns
+                if pd.api.types.is_float(CELLS_DF[colname].dtype)
+            }
+        ).reindex(columns=loaded_compartment_df.columns),
+        check_dtype=False,
```
gwaybio, line 298: Shouldn't `check_dtype=True`? Or maybe i'm not following the chaining.

pycytominer/tests/test_cyto_utils/test_cells.py
gwaybio: I have the same comment for the second test

pycytominer/tests/test_cyto_utils/test_cells.py
gwaybio: I have the same comment for the second test

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -416,6 +434,48 @@ def test_merge_single_cells():
         traditional_norm_df.loc[:, new_compartment_cols].abs().describe(),
     )
 
+    # use non-default float_datatype
+    sc_merged_df = AP.merge_single_cells(float_datatype=np.float32)
```
gwaybio, line 438: I'm a bit confused why this test is necessary? What are we testing here?

gwaybio (changes requested):
Looks good @d33bs ! This should help @axiomcura in WayScience/cell-health-data#10 (I think this was the right PR to tag here, but it will also likely help many others as well!)

pycytominer/cyto_utils/cells.py
d33bs: Great point, thank you - I'll update this to include a mention about the relationship to performance and an example.

pycytominer/cyto_utils/cells.py
d33bs: This makes a lot of sense, thank you for this feedback. Pondering on this makes me wonder if perhaps more than floating point data (for example, int's) may need or benefit from a default. I'll add in an attribute for `SingleCells` float datatypes with this in mind.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Understood and will do, thank you for the comment to address this. I'll add an update to this test to make certain it's addressed.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Thank you for the check here - I added this in to make certain the `float_datatype` work could cascade correctly from `merge_single_cells` in addition to `load_compartment`. With the changes you suggested for the class attribute I think this may no longer be relevant, so I'll remove it.

pycytominer/cyto_utils/cells.py
gwaybio: > Pondering on this makes me wonder if perhaps more than floating point data (for example, int's) may need or benefit from a default.

Yeah, it likely would! IIRC we had discussed this previously (creating some sort of feature dictionary) but decided to leave this for a future research project. 

My sense is to focus on `float_datatype` here, since we've seen such huge performance boosts and then worry about this dictionary in a future PR

pycytominer/tests/test_cyto_utils/test_cells.py
gwaybio: got it!

It's probably sufficient to assert `self.float_datatype` once you make the class attribute change

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -1050,3 +1079,12 @@ def test_load_non_canonical_image_table():
         result.sort_index(axis="columns").drop("Metadata_Site_Count", axis="columns"),
         sc_aggregated_df,
     )
+
+def test_singlecells_default_datatype():
```
gwaybio, line 1083: this can probably be deleted, right?

pycytominer/tests/test_cyto_utils/test_cells.py
```diff
@@ -293,6 +294,42 @@ def test_load_compartment():
...
+    ).load_compartment(compartment="cells")
+
+    # for uniformly handling metadata types for both dataframes
+    metadata_types = {"ObjectNumber": "int64"}
+
+    # create deep copy of CELLS_DF with manually re-typed float columns as float32
+    cells_df_for_compare = CELLS_DF.copy(deep=True).astype(
+        # cast any float type columns to float32 for expected comparison
+        {
+            colname: np.float32
```
gwaybio, line 310: is this blacked? I've never seen this format before, but I can read it!

gwaybio (approved):
Looks good @d33bs ! Only two minor comments but then I think we're good to merge.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: Thank you, yes, apologies. I'll remove this block.

pycytominer/tests/test_cyto_utils/test_cells.py
d33bs: I did apply black styling to this, but admittedly the dictionary comprehension mid-block makes things look maybe a bit unusual. Alternatively we could move the dictionary before the `astype` to make this a bit less terse or better matching the reading style of the testing suite. Thoughts?

gwaybio: My preference is to split out the dictionary compensation to increase readability. Love the comments though, so please retain these!

d33bs: Thanks, sounds good, will do.

gwaybio:
Looks good! I'm going to merge. Thanks @d33bs

PR #251, opened 2023-02-06
Pin SQLAlchemy below 2

### LINKED ISSUE #250: Sqlalchemy v2.0.0 bug causes `cells.SingleCells` to not work: ObjectNotExecutableError with read_sql
I noticed that when I pip installed Pycytominer to use in a project yesterday that I was getting this error when trying to run `cells.SingleCells` from the [cells.py file](https://github.com/cytomining/pycytominer/blob/master/pycytominer/cyto_utils/cells.py). 

```python
ObjectNotExecutableError: Not an executable object: 'select * from Per_Image'
```

This error is caused by a bug where between Pandas and the newest release of Sqlalchemy from a few days ago. It is described further in [this issue](https://github.com/pandas-dev/pandas/issues/51061) from the Pandas GitHub.

I noticed that in the [requirements.txt](https://github.com/cytomining/pycytominer/blob/master/requirements.txt) that it states:

```txt
sqlalchemy>=1.3.6
```
This means it is installing the latest version of Sqlalchemy when pip installing. This means a new user if they wanted to use the `cells.SingleCells` function or any function in Pycytominer that uses Sqlalchemy `read_sql` function using a string, then it will not work.

The only work around I have that works is a conda environment that installs Sqlalchemy v1.4.42 (which is the last version that works) and pip installs Pycytominer. The conda install override the pip install. 

Once this bug is fixed, then this won't be a worry, but I wanted to make this issue for others to be aware in case they get this same error.

bethac07:

# Description

Resolves #250

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ numpy>=1.16.5
 scipy>=1.3.0
 pandas>=1.2.0
 scikit-learn>=0.21.2
-sqlalchemy==1.4.42
+sqlalchemy>=1.3.6,<2
 pyarrow>=8.0.0
 pytest>=5.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ numpy>=1.16.5
 scipy>=1.3.0
 pandas>=1.2.0
 scikit-learn>=0.21.2
-sqlalchemy>=1.3.6
+sqlalchemy==1.4.42
 pyarrow>=8.0.0
 pytest>=5.0.1


requirements.txt
```diff
@@ -2,6 +2,6 @@ numpy>=1.16.5
 scipy>=1.3.0
 pandas>=1.2.0
 scikit-learn>=0.21.2
-sqlalchemy>=1.3.6
+sqlalchemy==1.4.42
```
gwaybio, line 5:
```suggestion
sqlalchemy<=1.4.42
```

Do you think it makes sense to be less restrictive? If you disagree, I will merge with a pinned version

gwaybio (approved):

requirements.txt
bethac07: I'd recommend going with `sqlalchemy>=1.3.6,<2`, or `sqlalchemy~=1.4.42` if you don't want an exact pin - I assume the previous version minimum was for a reason. But your call.

gwaybio: @d33bs - what do you recommend?

d33bs: Thanks @gwaybio - I would recommend what @bethac07 mentioned with a version range here, respecting that any broken aspects of the SQLAlchemy API in v2 may incur separate updates for a time. Specifically, I'd vote for `sqlalchemy>=1.3.6,<2` to help make sure we account for any reasons why 1.3.6 was preferred at one point or another. Dependencies could be updated and tested otherwise (for ex. updating >= 1.4.42) outside of the focus of this PR and #250 .

bethac07: Cool, have updated to @d33bs 's preference, since already approved will pull when tests are done!

gwaybio:
Thanks @bethac07 and @d33bs !

Issue #211, opened 2022-06-30
Add functionality to read parquet data sources


d33bs:
As discussed in #202 and #205 (and complementing #117), parquet is desired as an input file format for data read by Pycytominer. Reading parquet as input is an addition to other formats (such as CSV, NPZ, etc).

This work may benefit from documenting data schema across various file- or in-memory based formats.

PR #262, opened 2023-03-16
Implementing `parquet` loading in `load_profiles` function

axiomcura:
Edit by @gwaybio 

If you happen to stumble upon this PR, note the we elected to perform a simpler implementation than what is described immediately below. See files changed for our implemented solution.

Below is @axiomcura original post

---------------------------

## Implementing parquet loading

This PR focuses on solving #211  

This introduces a new function known as `is_path_a_parquet_file` that allows the `load_profile` function to identify whether the input file is either a `parquet` file. 

By identifying what file format the `profiles` are, it will use the appropriate instructions to load it into memory via `pandas`’s API. 

## Implementation approach

File types contain a unique signature within the header of each file. For example, `sqlite` files contains `SQlite format` signature within the first 100 bytes of contents. 

Below is a code example:

```python
buffer_size = 100 
sqlite_file = "pycytominer/tests/test_data/cytominer_database_example_data/test_SQ00014613.sqlite"
with open(sqlite_file, "rb") as stream:
    header_conts = stream.read(buffer_size)

print(header_conts)
```

Here is the output:

```
b'SQLite format 3\x00\x04\x00\x01\x01\x00@  \x00\x005\xfd\x01T\x0b\xd7\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\t\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x005\xfd\x00-\xe6\x02'
```

Similarly, `parquet` files also contain a unique format known as `PAR` within the first 100 bytes of data of the files. 

```python
buffer_size = 100 
sqlite_file = "pycytominer/tests/test_data/cytominer_database_example_data/test_SQ00014613.parquet""
with open(sqlite_file, "rb") as stream:
    header_conts = stream.read(buffer_size)

print(header_conts)
```

Here is the 

```
b'PAR1\x15\x04\x15\x90\xaf\x03\x15\xf0\xd7\x01L\x15\xf25\x15\x04\x12\x00\x00\xc8\xd7\x01\x04\x01\x00\t\x01\x00\x02\t\x07\x04\x00\x03\r\x08\x00\x04\r\x08\x00\x05\r\x08\x00\x06\r\x08\x00\x07\r\x08\x00\x08\r\x08\x00\t\r\x08\x00\n\r\x08\x00\x0b\r\x08\x00\x0c\r\x08\x00\r\r\x08\x00\x0e\r\x08\x00\x0f\r\x08\x00\x10\r\x08\x00\x11\r\x08\x00\x12\r\x08'
```

This PR introduces a new function known as `infer_profile_file_type`that leverages these unique signatures from `parquet` and `sqlite` types. 

Because of this, the `load_profile` function is able to identify which files to load  

Example below: 

loading `csv` file:

csv_loading

loading `parquet` files:

parquet_loading

---

# Description

# Checklist


gwaybio:
Is there a reason not to simply use Pathlib.Path.suffix? https://docs.python.org/3/library/pathlib.html?highlight=suffix#pathlib.PurePath.suffix

This would easily support csv as well

d33bs:
> Is there a reason not to simply use Pathlib.Path.suffix? https://docs.python.org/3/library/pathlib.html?highlight=suffix#pathlib.PurePath.suffix
> 
> This would easily support csv as well

Agreeing with @gwaybio here, it may be simpler to use the file extension via `pathlib.Path.suffix` if available. If extensions are not available for specific datasets, inference might still be useful. Is there a specific dataset you've seen which does not include the extensions? Referencing this may help with the discussion. Maybe also a combined approach could make sense (first pathlib.path.suffix, then inference)?

For filetype inference in specific: Leveraging existing work in this area might be a good idea due to the complexities involved. [`python-magic`](https://github.com/ahupp/python-magic) in particular could be useful (possibly using `python-magic-bin` for dependencies, though this too is not without possible challenges across various OS environments). Especially if/when other formats are considered, providing a scaffold which is widely accepted can help with maintenance over time.

pycytominer/cyto_utils/load.py
```diff
@@ -1,8 +1,66 @@
...
+    headers.
+
+    Parameters
+    ----------
+    file : str
+        file name
+
+    Return
+    -------
+    str
```
d33bs, line 23: The care you provided through type hints within the docstring here made me wonder if it might be helpful to include within the function definition as well. Doing this can be a matter of ["gradual typing"](https://peps.python.org/pep-0483/#summary-of-gradual-typing).

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -196,3 +197,24 @@ def test_load_npz():
...
             example_npz_file_locations, location_x_col_index=0, location_y_col_index=2
         )
+
+def test_infer_plate_files():
+
+    # inferring parquet files
+    parquet_file = "../test_data/cytominer_database_example_data/test_SQ00014613.parquet"
+    expected_file_type_1 = "parquet"
+    test_file_type_infer_1 = infer_profile_file_type(parquet_file)
+    assert(expected_file_type_1 == test_file_type_infer_1)
```
d33bs, line 207: This is mostly up to your preference and I wanted to make a suggestion here for simplicity. Consider making direct references instead of declaring variables for items which are used only once (this stands for some other lines below as well). For example, see this suggestion:
```suggestion
    assert(infer_profile_file_type(parquet_file) == "parquet")
```

d33bs (changes requested):
Nice work - excited to see parquet compatibility additions! I made a few comments throughout. Please don't hesitate to let me know if you have any questions.

I feel the inference discussion and also the failing test ([see here](https://github.com/cytomining/pycytominer/actions/runs/4442165365/jobs/7798101653?pr=262)) should be resolved before we move forward.

axiomcura:
@gwaybio @d33bs 

My initial thought is that using extension based file validation can result in errors. For example, users can submit extension-less files that may "confuse" `pycytominer` loading functions. Therefore resorting to file signatures can help avoid this issue. 

I do agree that the implementation is much more involved and using only extensions is much easier to implement. I was looking at [`python-magic`](https://github.com/ahupp/python-magic) that @d33bs  has recommended and it's pretty handy. 

Here are some examples below:
```python
import pathlib
import magic

# magic object
magic_reader = magic.Magic(uncompress=True, extension=False)
print(magic_reader.from_file("./test_SQ00014613.parquet"))
print()
print(magic_reader.from_file("./test_SQ00014613.sqlite"))
print()
print(magic_reader.from_file("./test_SQ00014613.csv.gz"))
```
Here is the output:
```text
Apache Parquet

SQLite 3.x database, last written using SQLite version 3037002, page size 1024, file counter 13832, database pages 192, cookie 0xa, schema 4, UTF-8, version-valid-for 13832

CSV text (gzip compressed data, was "test_SQ00014613.csv", last modified: Sat Sep  3 15:32:39 2022, max compression)
```

In cases with symlink files, we need to extract the soruce path in order for it to work. Here's an example:
```python 
import pathlib
import magic

# magic object
magic_reader = magic.Magic(uncompress=True, extension=False)

# now lets try with filepaths with symbolic links
sym_link_data = "./SQ00014617.parquet"
path = pathlib.Path(sym_link_data)

# check if the path is a sym_link
# -- if so, get true path
if path.is_symlink():
    path = path.resolve()

# with true path, identify file
print(magic_reader.from_file(path))
```
output:
```text
Apache Parquet
```

If this is too much at this moment, then resorting to only file extensions is an ideal plan! Let me know what ya'll think :)

gwaybio:
> If this is too much at this moment, then resorting to only file extensions is an ideal plan! Let me know what ya'll think :)

My opinion is that it is too much - I think keeping things simple is the way to go for readability and maintainability. I generally get a bit queazy when adding a new dependency as well.

I think all we need is the following:

1. function to ask whether or not the input string or path is a parquet extension (maybe called `is_path_a_parquet_file()` 🤷 
2. An update to `load_profiles(profiles)` that calls `is_path_a_parquet_file()` and uses `pandas.DataFrame.read_parquet()` if true
3. corresponding tests

axiomcura:
@gwaybio and @d33bs 

I have replace my previous implementation with extension based inference of parquet files by using `pathlib.Path.suffix`

pycytominer/cyto_utils/load.py
```diff
@@ -1,8 +1,48 @@
...
+
+    Raises
+    ------
+    TypeError
+        Raised if a non str object is passed in the `file` parameter
+    FileNotFoundError
+        Raised if the provided path in the `file` does not exist 
+    """
+    # type checking
+    if not isinstance(file, str):
```
gwaybio, line 32: We probably should allow pathlib.Path as input as well.

pycytominer/cyto_utils/load.py
```diff
@@ -1,8 +1,48 @@
 import csv
 import gzip
+from pathlib import Path
```
gwaybio, line 3: is `pathlib` a large package? If not, then I think it's best to `import pathlib` and use `pathlib.Path` instead of just `Path` (to increase readability)

pycytominer/cyto_utils/load.py
```diff
@@ -1,8 +1,48 @@
...
 import numpy as np
 import pandas as pd
 
+def is_path_a_parquet_file(file: str) -> bool:
+    """Checks if the provided file path is a parquet file.
+
+    Identification of parquet files are done by inspecting the file extensions.
```
gwaybio, line 10:
```suggestion
    Identify parquet files by inspecting the file extensions. 
```

to avoid passive voice

pycytominer/cyto_utils/load.py
```diff
@@ -41,20 +81,25 @@ def load_profiles(profiles):
     Return
     ------
     pandas DataFrame of profiles
+
+    Raises
+    ------
+    FileNotFoundError
+        Raised if the provided
```
gwaybio, line 88: This looks like an incomplete sentence

pycytominer/cyto_utils/load.py
```diff
@@ -41,20 +81,25 @@ def load_profiles(profiles):
...
     """
     if not isinstance(profiles, pd.DataFrame):
-        try:
-            delim = infer_delim(profiles)
-            profiles = pd.read_csv(profiles, sep=delim)
-        except FileNotFoundError:
```
gwaybio, line 49: Why did you delete this block?

pycytominer/cyto_utils/load.py
```diff
@@ -41,20 +81,25 @@ def load_profiles(profiles):
...
+        
+        delim = infer_delim(profiles)
+        profiles = pd.read_csv(profiles, sep=delim)
+
     return profiles
 
 
 def load_platemap(platemap, add_metadata_id=True):
     """
-    Unless a dataframe is provided, load the given platemap dataframe from path or string
```
gwaybio, line 56: Why delete this docstring?

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -10,7 +10,8 @@
     load_npz_features,
     load_npz_locations,
 )
-from pycytominer.cyto_utils.load import infer_delim
+from pycytominer.cyto_utils.load import (infer_delim,
```
gwaybio, line 13: I think this needs `black`

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -196,3 +197,22 @@ def test_load_npz():
         load_npz_locations(
             example_npz_file_locations, location_x_col_index=0, location_y_col_index=2
         )
+
+def test_infer_plate_files():
```
gwaybio, line 201:
```suggestion
def test_is_path_a_parquet_file():
```

To align the name with the function you're testing

pycytominer/tests/test_cyto_utils/test_load.py
gwaybio: yep! see the red symbol at the bottom :)

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -196,3 +197,22 @@ def test_load_npz():
...
+
+    # checking parquet file
+    check_pass = is_path_a_parquet_file(parquet_file)
+    check_fail = is_path_a_parquet_file(csv_file)
+
+    # checking if the correct booleans are returned
+    assert(check_pass, True)
+    assert(check_fail, False)
+
+    # loading in pandas dataframe from parquet file
```
gwaybio, line 215: Is it worth also loading the csv file with `load_profiles()` and `pd.testing.assert_frame_equal(csv_df, parquet_df)`?

gwaybio (changes requested):
A couple comments that need to be addressed prior to merging - looking good!

pycytominer/tests/test_cyto_utils/test_load.py
gwaybio: It also looks like you need to add the test parquet file https://github.com/cytomining/pycytominer/actions/runs/4470828258/jobs/7854910373?pr=262#step:5:91

pycytominer/cyto_utils/load.py
axiomcura: hmmm... It doesn't appear deleted on my end :thinking:

pycytominer/cyto_utils/load.py
```diff
@@ -1,10 +1,54 @@
...
+    Raises
+    ------
+    TypeError
+        Raised if a non str object is passed in the `file` parameter
+    FileNotFoundError
+        Raised if the provided path in the `file` does not exist
+    """
+    # type checking
+    accepted_type = (str, pathlib.Path)
+    if not isinstance(file, accepted_type):
```
d33bs, line 34: Consider using the tuple directly within `isinstance` to avoid creating a variable which is only used once.

```suggestion
    if not isinstance(file, (str, pathlib.Path)):
```

pycytominer/cyto_utils/load.py
```diff
@@ -1,10 +1,54 @@
...
+        raise TypeError(f"file must be a str not {type(file)}")
+
+    # convert str to pathlib.Path object
+    if isinstance(file, str):
+        file = pathlib.Path(file).absolute()
```
d33bs, line 39: Would it make sense to use the absolute path if `file` is of `pathlib.Path` type too?

pycytominer/cyto_utils/load.py
```diff
@@ -124,9 +176,12 @@ def load_npz_features(npz_file, fallback_feature_prefix="DP", metadata=True):
     # Load metadata
     if "metadata" in files:
         metadata = npz["metadata"].item()
-        metadata_df = pd.DataFrame(metadata, index=range(0, df.shape[0]), dtype=str)
+        metadata_df = pd.DataFrame(
+            metadata, index=range(0, df.shape[0]), dtype=str
```
d33bs, line 180: How does the index specified here impact the DataFrame? If it's not needed, consider removing.

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -196,3 +198,28 @@ def test_load_npz():
...
+    parquet_df = pd.read_parquet(parquet_file)
+    parquet_profile_test = load_profiles(parquet_file)
+    pd.testing.assert_frame_equal(parquet_profile_test, parquet_df)
+
+    # loading csv file with new load_profile()
+    csv_df = pd.read_csv(csv_file)
+    csv_profile_test = load_profiles(csv_file)
+    pd.testing.assert_frame_equal(csv_profile_test, csv_df)
```
d33bs, line 225: Consider adding a newline at the end of this file.

pycytominer/tests/test_cyto_utils/test_load.py
d33bs, line 13: Would `infer_delim` benefit from increased test coverage?

d33bs (changes requested):
Nice job! I left a few comments and suggestions with this review. Please don't hesitate to let me know if you have any questions.

I'd request your feedback on partitioned parquet-based directories (those containing multiple files) with `is_path_a_parquet_file` before approval.

pycytominer/cyto_utils/load.py
```diff
@@ -1,10 +1,58 @@
...
+        raise TypeError(f"file must be a str not {type(file)}")
+
+    # convert str to pathlib.Path object
+    if isinstance(file, str):
+        file = pathlib.Path(file).absolute()
+
+    # if a pathlib.Path object is passed check if it is an absolute path
+    # -- if not, get absolute path
+    if not file.is_absolute():
+        file = file.absolute()
```
gwaybio, line 44:
```suggestion
    # Convert str to pathlib.Path object
    if isinstance(file, str):
        file = pathlib.Path(file)

    # Convert to absolute path if necessary
    if not file.is_absolute():
        file = file.absolute()
```

This feels simpler and more pythonic to me. What do you think?

Also, is `if not file.is_absolute():` necessary? Will an already absolute path fail if `.absolute()` is called?

gwaybio: Please also note the simpler and properly formatted comments

pycytominer/cyto_utils/load.py
```diff
@@ -1,10 +1,58 @@
...
+
+    # checking if the file exists
+    if not file.exists():
+        raise FileNotFoundError(f"{str(file.absolute())} does not exist")
```
gwaybio, line 48:
```suggestion
        raise FileNotFoundError(f"{str(file)} does not exist")
```

No need to call `.absolute()` again, right?

pycytominer/cyto_utils/load.py
```diff
@@ -1,10 +1,58 @@
...
+    # checking if file path is a parquet file
+    if not file.suffix.lower() == ".parquet":
+        return False
+    return True
```
gwaybio, line 52:
```suggestion
    # Check if file path is a parquet file
    if file.suffix.lower() == ".parquet":
        return True
    
    return False
```

Avoiding the double negative makes things only marginally easier to read, and might be preferable.

pycytominer/tests/test_cyto_utils/test_load.py
gwaybio: Still here

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -188,11 +215,35 @@ def test_load_npz():
...
+    check_fail = is_path_a_parquet_file(output_data_file)
+
+    # checking if the correct booleans are returned
+    assert (check_pass, True)
+    assert (check_fail, False)
+
+    # loading in pandas dataframe from parquet file
+    parquet_df = pd.read_parquet(output_data_parquet)
+    parquet_profile_test = load_profiles(output_data_parquet)
+    pd.testing.assert_frame_equal(parquet_profile_test, parquet_df)
```
gwaybio, line 244: I asked this in my previous review, but maybe I was not clear.

You're testing parquet vs. parquet and csv vs. csv (which is great!) but what about parquet vs. csv? 

Maybe at the very end of this test, add the following:

```python
pd.testing.assert_frame_equal(parquet_profile_test, csv_profile_test)
```

gwaybio (commented):
Nearly there! 🎉 

I made some comments which probably should be addressed prior to merging. I will let @d33bs give the final 👍

pycytominer/tests/test_cyto_utils/test_load.py
```diff
@@ -188,11 +215,38 @@ def test_load_npz():
...
+
+    # loading csv file with new load_profile()
+    csv_df = pd.read_csv(output_data_comma_file)
+    csv_profile_test = load_profiles(output_data_comma_file)
+    pd.testing.assert_frame_equal(csv_profile_test, csv_df)
+
+    # checking if the same df is produced from parquet and csv files
+    pd.testing.assert_frame_equal(parquet_profile_test, csv_profile_test)
```
d33bs, line 252: Consider adding a newline at the end of this file.

d33bs (changes requested):
Thanks Erik, great job on the new additions! I'm requesting some changes and your thoughts on a few items, especially on `pathlib.Path.absolute()` usage. Please don't hesitate to let me know if you have any questions.

axiomcura:
Helllo @d33bs 

I have attended your comments. Hopefully they were answered!

pycytominer/cyto_utils/load.py
```diff
@@ -32,23 +32,20 @@ def is_path_a_parquet_file(file: Union[str, pathlib.Path]) -> bool:
...
     if isinstance(file, str):
-        file = pathlib.Path(file)
-
-    # Convert to absolute path if necessary
-    if not file.is_absolute():
-        file = file.absolute()
+        try:
+            file = pathlib.Path(file).resolve(strict=True)
+        except FileExistsError as e:
+            raise FileNotFoundError(f"{str(file)} does not exist") from e
```
d33bs, line 43: A `FileExistsError` exception might not be triggered with the `try` step above (which attempts to resolve a path object but does not create it). Consider relying on `pathlib.Path.resolve(strict=True)` without this try/except block, relying on its built-in exception handling instead.

d33bs (changes requested):

d33bs (approved):
Nice work!

gwaybio (approved):
This looks great @axiomcura - congrats on this contribution! 🎉

axiomcura:
Thank you! Merging :)!

PR #298, opened 2023-06-19
Parse CellProfiler features

shntnu:
# Description

This PR introduces a function to parse a CellProfiler feature string into its semantic components -- we are missing a standardized method for doing this.

I've focused on [features](https://github.com/cytomining/pycytominer/blob/78e1503196b697a100464f6a56d8b816e2d39055/pycytominer/tests/test_data/parse_cp_features_example_data/cp_features.txt) generated in the JUMP Cell Painting datasets.

# Checklist


ErinWeisbart:
One further thought, I think your code breaks if there are `_` in the channel name, but this is something we do consistently with `Brightfield` features e.g. `Brightfield_high` and `Brightfield_low` and occasionally with other channels to represent high/low focal planes (for Brightfield) or exposures (for fluorescence channels).

shntnu:
> My only quibble is that I would like the `Mito` channel to be returned as `Mito` and not `MITO`. (It is our standard styling, unlike all other channels that are usually ALLCAPS).

Now fixed in 26591d7 (but later superseded by 36b5d50)

> One thing to consider - the features measured in JUMP (and our subsequent default CP pipeline) have both `Mito` and `mito` features which this script handles and parses as `MITO` (though if I get my way that will change to `Mito` :) I don't know the intended use of this script, but there is actually a difference between `Mito` and `mito` features. `Mito` features are captured from the `Mito` channel images, just like any other channel's features. `mito` features are on heavily processed images that did initially derive from `Mito` images but are heavily manipulated to generate additional features that were targeted for a specific project/screen. We do not manipulate/process any other channels in a similar manner. So, that leads me to the question of do we want to distinguish between `Mito` and `mito` because `mito` features are derived differently than all other features, or do we not care?

Darn – that's definitely annoying, and we would want to capture this ideally (and we'd want to call that channel `mito_tubeness`, not `mito` IIUC)

 I wanted to keep this as general as possible, so maybe the right way to address this issue (and the issue of allowing underscores) is to specify the list of channels as a parameter. 

See 36b5d50

d33bs:
Hello, this is a courtesy notice that Pcytominer's `master` branch has been renamed to `main` as part of changes merged in #303. While changes here should be seamless, GitHub is unable to change your local git environment and offers the following guidance when it comes to this. Thank you for your understanding and please don't hesitate to reach out with any questions or concerns.

image

shntnu:
@d33bs @gwaybio – this was already reviewed by @ErinWeisbart and is ready to be merged. Would you like to review next or should we proceed with merging?

gwaybio:
Yes, we'll need to review. @kenibrewer maybe you can take a look?

I'll make two guiding notes here:

- this touches upon two items from our roadmap (@shntnu , i still need to share the document with you, we will make public too, just haven't had the chance). 1) institution specific code. We will work to refactor this out of pycytominer in subsequent versions, aiming for greater generalizability. 2) cli. We will implement Fire in subsequent versions.
- cyto_utils has less but still rigorous merge guidelines, and this PR, upon first glance, might take some time to merge.

Thanks!

shntnu:
@gwaybio – I'm in no hurry to merge this, so certainly let's follow the roadmap all through (exciting!)

kenibrewer (changes requested):
@shntnu Thanks for this contribution. As @gwaybio mentioned, part of our technical roadmap for pycytominer is moving away from hard-coded uses of CellProfiler feature names towards more generalizable approaches that work different software. A way we'd like to accomplish this is by a greater reliance on abstractions and object-oriented programming. Having cell_profiler features be simply one of many solutions supported. 

Ideally, I think we would implement something similar to what is described in #360 first and then make this PR a method of that class. Is that an approach that you think would be interested in implementing? Alternately, this script could also be refactored to accept a regex and set of feature groups to extract.

shntnu:
> @shntnu Thanks for this contribution. As @gwaybio mentioned, part of our technical roadmap for pycytominer is moving away from hard-coded uses of CellProfiler feature names towards more generalizable approaches that work different software. A way we'd like to accomplish this is by a greater reliance on abstractions and object-oriented programming. Having cell_profiler features be simply one

I like this plan!

> Ideally, I think we would implement something similar to what is described in #360 first and then make this PR a method of that class. Is that an approach that you think would be interested in implementing? 

I am happy for us to take this approach but I would not have the capacity to think through what's needed here. 

IIUC the plan is to introduce a class like [_SummarizedExperiment_](https://bioconductor.org/packages/release/bioc/html/SummarizedExperiment.html) or [_SingleCellExperiment_](https://bioconductor.org/packages/release/bioc/html/SingleCellExperiment.html) but probably not as general as [_AnnData_](https://anndata.readthedocs.io/en/latest/)?

That's quite a foundational effort if so. 

Is that what you have in mind? 

I'm happy to have this PR sit around for a bit until this upstream issue is sorted out.

> Alternately, this script could also be refactored to accept a regex and set of feature groups to extract.

Hm – I don't think this would be practical; it would need a really complex regex to replace all those conditional statements.

gwaybio:
> I'm happy to have this PR sit around for a bit until this upstream issue is sorted out.

The upstream issue is complex, and we have no timetable for addressing. So closing for staleness - please feel free to reopen if changes are needed!

Issue #278, opened 2023-05-03
[Refactor] Migrate from setuptools to pyproject.toml


kenibrewer:
A pyproject.toml file allows more standardized definition of dev, test, build, and run dependencies. It would also allow more granular configuration of tools such as black with configurable line width and other factors. 

Here is a good guide to setting up a pyproject.toml that I would recommend to anyone working on this issue: https://betterprogramming.pub/a-pyproject-toml-developers-cheat-sheet-5782801fb3ed

gwaybio:
@kenibrewer - are you able to take this on? We'd like to have this for version 1

kenibrewer:
@gwaybio What's the timeline for the version 1 release? I'll be pretty busy until June 24th, but will have some additional time to work on open-source projects after that.

gwaybio:
We are aiming to release version 1 prior to paper submission (which we're planning concurrently, all pycytominer contributors will be authors). We're aiming paper submission in 4 months.

So after June 24 sounds great! (we'll likely want to release several months before paper sub, as a soft deadline)

kenibrewer:
Sure. I can take this on. Feel free to assign this issue to me so I can keep track.

I've been looking at Python Poetry lately. Any objections if I use that as the build engine and for dependency management?

gwaybio:
No objections! @d33bs has experience and will likely be happy to see this. Thanks!

d33bs:
Cheers @kenibrewer - Poetry is a pleasure to work with and could offer enhancements to this project. Do you have any thoughts surrounding how core python `pyproject.toml` expectations may align with that of Poetry's format? My last recollection was that there may be inconsistencies to be aware of when moving towards Poetry (potentially minor, likely depending on the scenario), but maybe this has changed.

gwaybio:
Checking in on this @kenibrewer - have you been able to find the time? This is the final feature prior to version 1 release! :)

(no worries if you haven't found time, we can prioritize this if need be - there is also no real rush, just eagerness on our part)

PR #303, opened 2023-07-04
Switch build from setuptools to Poetry and move tests path

kenibrewer:
# Description

This pull request implements #278 and switches the project build approach from setuptools to pyproject.toml using poetry for dependency management.

Additional changes made during conversion:

- [x] `tests` directory moved from `pycytominer/tests` to `<root>/tests`
- [x] Refactored `tests/test_cyto_utils/test_DeepProfiler_processing.py` to move process heavy parts into test_functions
- [x] Changed workflows to use poetry for test / building / publishing
- [x] Removed .vscode from gitignore and added pytest, pyenv, and poetry configuration relevant to devcontainers
- [x] Added poetry precommit hooks to keep the `poetry.lock` up to date and `requirements{-dev}.txt` in sync with `pyproject.toml`
- [x] Change `cyto_utils.collate.py` to use python functions instead of cli commands.
- [x] Documentation update to `CONTRIBUTING.md` about poetry

# Checklist


kenibrewer:
@gwaybio What are your thoughts about switching from `master` to the more inclusive `main` default branch for the project as a part of this PR?

gwaybio:
I am in favor @kenibrewer , please make the switch if it's straightforward. Thanks!

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+codecov = ">=2.1.12"
+pre-commit = ">=3.3.2"
+
+[build-system]
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.black]
+line-length = 88
+target-version = ['py38']
```
d33bs, line 44: Consider aligning this version with what's found in `pre-commit-ci.yml` (Python 3.9). If there's a reason not to do this, consider adding a comment here to describe the version selection.

.pre-commit-config.yaml
```diff
@@ -8,3 +8,12 @@ repos:
...
       - id: prettier
+  - repo: https://github.com/python-poetry/poetry
+    rev: "1.5.1"
+    hooks:
+      - id: poetry-check
+      - id: poetry-lock
+      - id: poetry-export
+        args: ["-f", "requirements.txt", "-o", "requirements.txt"]
+      - id: poetry-export
+        args: ["--dev", "-f", "requirements.txt", "-o", "requirements-dev.txt"]
```
d33bs, line 19: With the transition to a `pyproject.toml` file and Poetry these requirements files might not be necessary. Consider removing these lines and the `requirements*.txt` files to help simplify and unify setup maintenance or implementation. If they are still needed as a backup or for compatibility of some kind, consider adding comments here and within the docs to help describe this.

pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[tool.poetry]
+name = "pycytominer"
+version = "1.0.0"
+description = "Processing perturbation profiling readouts."
```
d33bs, line 4: Consider aligning the value here with that of the Github repository description.

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+fsspec  =   {version = ">=2023.1.0",    extras=["cell_locations"]}
+s3fs    =   {version = ">=0.4.2",       extras=["cell_locations"]}
+boto3   =   {version = ">=1.26.79",     extras=["cell_locations"]}
+fire    =   {version = ">=0.5.0",       extras=["cell_locations"]}
+
+cytominer-database = {version = "0.3.4", extras=["collate"]}
+
+[tool.poetry.group.dev.dependencies]
+pytest = ">=5.0.1"
+pytest-cov = "^4.1.0"
```
d33bs, line 34: This looks like a new addition to the dev dependencies. Would it make sense to prefer one coverage tool over the other? If there are reasons for including both this and `codecov` consider adding a comment to help describe why.

tests/test_cyto_utils/test_DeepProfiler_processing.py
```diff
@@ -19,92 +20,87 @@
...
-df_well = well_class.aggregate_deep()
 
-plate_class = AggregateDeepProfiler(
-    deep_data=deep_data,
-    aggregate_operation="median",
-    aggregate_on="plate",
-    output_file=output_folder,
-)
-df_plate = plate_class.aggregate_deep()
+def test_DeepProfiler():
```
d33bs, line 24: Seeing the changes here and below made me wonder what might be lost from more modular testing given the move one larger test (I may be missing context here on design decisions surrounding these changes, and if so, apologies!). Consider retaining the modularity from the existing test functions instead of removing them and using a [Pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) to help ensure the data used by these functions is provided from one common declarative spot.

tests/test_cyto_utils/test_DeepProfiler_processing.py
d33bs, line 1: Given the changes to tests here, consider updating the title of the PR to also indicate changes to tests.

tests/test_cyto_utils/test_collate.py
```diff
@@ -9,22 +9,31 @@
 
 plate = "BR00121431"
 
-test_config_location = pathlib.Path(
-    f"{os.path.dirname(__file__)}/../../cyto_utils/database_config/ingest_config.ini"
+test_dir = pathlib.Path(__file__).parents[1].absolute()
```
d33bs, line 12: Consider using all-caps for global variable names in this file to help distinguish these. Alternatively, consider using a [Pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) which would help here and possibly elsewhere along with reducing repetition. For example:
```suggestion
TEST_DIR = pathlib.Path(__file__).parents[1].absolute()
```

tests/test_cyto_utils/test_collate.py
```diff
@@ -9,22 +9,31 @@
...
 
-test_csv_location = pathlib.Path(
-    f"{os.path.dirname(__file__)}/../test_data/collate/backend/{batch}/{plate}/{plate}.csv"
+test_csv_location = (
+    test_dir / "test_data" / "collate" / "backend" / batch / plate / f"{plate}.csv"
 )
 
-master_csv_location = pathlib.Path(
-    f"{os.path.dirname(__file__)}/../test_data/collate/backend/{batch}/{plate}/{plate}_master.csv"
+master_csv_location = (
```
d33bs, line 29: Consider updating the name of this variable to be in alignment with the potential changes from the `master` branch to the `main` branch inclusive naming scheme.
```suggestion
main_csv_location = (
```

tests/test_cyto_utils/test_collate.py
```diff
@@ -9,22 +9,31 @@
...
+    test_dir
+    / "test_data"
+    / "collate"
+    / "backend"
+    / batch
+    / plate
+    / f"{plate}_master.csv"
```
d33bs, line 36: Similar comment to the above, consider using `main` over `master`.
```suggestion
    / f"{plate}_main.csv"
```

d33bs (changes requested):
Nice work! Excited to see the transition into a `pyproject.toml` file and Poetry. I left a few comments regarding the changes and also wanted to follow up here with additional questions. Generally I'm requesting changes due to these more wide-reaching questions and hoping to better understand the development story behind your design considerations.

- Given how different Poetry and `pyproject.toml` development configuration can be, I highly recommend making documentation additions to the `CONTRIBUTING.md` content or perhaps in the Sphinx docs. This would help support the current and future developer community surrounding Pycytominer. Without this and the current state of retaining the `requirements*.txt` means developers may misinterpret how environment management is handled.
- Regarding the changes to `tests/test_cyto_utils/test_DeepProfiler_processing.py` (and perhaps others): were these changes necessary in order to maintain passing tests in transitioning to the new setup? I appreciate the work towards changes in those files. At the same I felt that those changes might be different enough to become their own PR's to help distinguish the changes going on with the setup process. Reducing the number of additional changes alongside setup and environment management would decrease the chance for unintended or complex consequences and increase the overall assurance that the setup works like it did prior to the changes in this PR. Just the same, I'm seeking to understand and learn from you - so open to your thoughts and interested to hear more!

.devcontainer/postCreateCommand.sh
```diff
@@ -1,5 +1,5 @@
 #! /usr/bin/env bash
-
-python -m pip install -r requirements-dev.txt
-python -m pip install -e .
-pre-commit install --install-hooks
+python -m pip install poetry
+poetry config virtualenvs.in-project true --local
```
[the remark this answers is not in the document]
kenibrewer: I agree that it would be good to put a more complete `.gitignore` into place. I also think it makes sense to do that in the context of this PR where we are changing substantial aspects of the build, & development workflow. I've added the additional content for the Python template.

kenibrewer:
> * Given how different Poetry and `pyproject.toml` development configuration can be, I highly recommend making documentation additions to the `CONTRIBUTING.md` content or perhaps in the Sphinx docs. This would help support the current and future developer community surrounding Pycytominer. Without this and the current state of retaining the `requirements*.txt` means developers may misinterpret how environment management is handled.

I agree. I added an additional todo item to this PR for documentation that I left unchecked. I wanted to get feedback on the initial set of changes that I proposed before doing the documentation changes. 

> * Regarding the changes to `tests/test_cyto_utils/test_DeepProfiler_processing.py` (and perhaps others): were these changes necessary in order to maintain passing tests in transitioning to the new setup? I appreciate the work towards changes in those files. At the same I felt that those changes might be different enough to become their own PR's to help distinguish the changes going on with the setup process. 

The changes to the other test files were necessary as a part of changing their paths. `test_DeepProfiler_processing.py` required some path changes too, but I ended up doing more work on that file because it took me a several hours to realize the root cause of  #304. Having pytest crash during test collection with no informative error message was a very frustrating experience. Without these changes, I am unable to run tests locally in a Github Codespace and can only discover failed tests when I run the github pipelines.

In general I agree with the strategy of trying to limit the "bast radius" of PRs. However, in the case of this PR with changes to core aspects of the code is built I think it is worth being a bit more flexible in allowing a broader set of related changes required to allow the program to function as a whole.

.github/workflows/codecov.yml
```diff
@@ -10,17 +10,48 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v3
       - name: Setup Python
-        uses: actions/setup-python@master
+        uses: actions/setup-python@v4.6.1
```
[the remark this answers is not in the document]
kenibrewer: Done

.gitignore
```diff
@@ -7,8 +7,8 @@ build
 .idea/*
 /dist
 *.sqlite
-.vscode/
-pycytominer/tests/test_data/collate/backend/**/*.csv
-!pycytominer/tests/test_data/collate/backend/**/*master.csv
-!pycytominer/tests/test_data/cell_locations_example_data/*.sqlite
+poetry.toml
```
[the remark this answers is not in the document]
kenibrewer: Done

tests/test_cyto_utils/test_DeepProfiler_processing.py
kenibrewer: @d33bs I agree that using a pytest fixture is the correct approach here. I spent some time trying to implement that until I realized that it will require a complete restructuring of the entire set of tests in this file. I think we should add a note about the fixture approach in #304.  

The way an entire data package was generated outside of functions and functions only included assert statements about that data package means that these changes did not actually decrease real modular testing.

tests/test_cyto_utils/test_DeepProfiler_processing.py
kenibrewer: Done.

.pre-commit-config.yaml
kenibrewer: Done.

pyproject.toml
kenibrewer: Done

pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[tool.poetry]
+name = "pycytominer"
+version = "1.0.0"
+description = "Processing perturbation profiling readouts."
+authors = [
+    "Gregory Way <gregory.way@gmail.com>",
+]
+maintainers = [
```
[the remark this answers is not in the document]
kenibrewer: I like this idea, but it would probably be best to include in a different PR. Would you like to create that issue?

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[tool.poetry.dependencies]
+python          = ">=3.8"
+numpy           = ">=1.16.5"
+scipy           = ">=1.5"
+pandas          = ">=1.2.0"
+scikit-learn    = ">=0.21.2"
+sqlalchemy      = ">=1.3.6,<2"
+pyarrow         = ">=8.0.0"
+
```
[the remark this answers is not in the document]
kenibrewer: Done

pyproject.toml
kenibrewer: This was an oversight and has been removed.

pyproject.toml
kenibrewer: Done

.pre-commit-config.yaml
kenibrewer: I also removed the poetry lock pre-commit hook because I think it would be best if the `poetry.lock` file only changes after making changes in response to changes to the `pyproject.toml` file.

tests/test_cyto_utils/test_collate.py
kenibrewer: Done.

.github/workflows/codecov.yml
```diff
@@ -10,17 +10,48 @@ jobs:
...
-          pytest --cov=./ --cov-report=xml
+          python-version: 3.11
+      #----------------------------------------------
+      #  -----  install & configure poetry  -----
+      #----------------------------------------------
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
```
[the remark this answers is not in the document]
kenibrewer: Agreed. Done

.github/workflows/python-app.yml
```diff
@@ -28,12 +28,37 @@ jobs:
...
+      #----------------------------------------------
+      # install dependencies if cache does not exist
+      #----------------------------------------------
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install pytest
-          pip install .[collate,cell_locations,dev]
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root  --all-extras
```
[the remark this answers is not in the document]
kenibrewer: The `--no-root` option installs all dependences, but not the current project. Those dependencies are stored in the `.venv` folder which can be loaded via caching (and gets invalidated by changing `poetry.lock`).

The next install is intended to be a  `--root-only` install (which I've added for clarity) executes any auxiliary install instructions related to the package. (e.g. adding a script file to path). These changes can be different even when the `poetry.lock` file doesn't change. 

I added some additionaly clarity via comments too.

tests/test_cyto_utils/test_collate.py
kenibrewer: Done

kenibrewer:
@d33bs I still have some outstanding changes related to versioning and documentation, but it would be good to get another review from you at this point for the changes so far.

pyproject.toml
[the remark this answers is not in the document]
d33bs: Agreed, thanks! I've created #310 in reference to this.

tests/test_cyto_utils/test_DeepProfiler_processing.py
d33bs: Thank you for the details here; I better understand the context here. We do lose pytest test modularity here; there is now one large test performed with many asserts as opposed to smaller tests with fewer asserts. One drawback of this is the reduction of testing specificity during development (where we only want to test one failing test to conserve time, for example). This may make maintenance and test troubleshooting more difficult in the future.

I also wasn't sure and might have missed it: have we addressed or documented the root cause of the reason the tests were failing for `test_DeepProfiler_processing.py` to help justify the reasons for switching to the new approach found in this PR?

Generally, I'd still recommend a fixture approach here if possible to help retain the modularity and resolve the challenges mentioned in #304. It might be that [fixture scope](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) and multi-value / tuple returns from a fixture could help.

gwaybio:
@kenibrewer - are you able to take a look at this again? I'm wondering if there are any pending items for @d33bs or I to address prior to your response, but I don't see any.

kenibrewer:
Thanks for checking in @gwaybio. I'll be able to work on this in earnest again early next week.

.github/workflows/python-app.yml
```diff
@@ -5,9 +5,15 @@ name: Python build
...
+    branches:
+      - main
+      # Temporary inclusion of master branch during default branch name transition
+      - master
   pull_request:
-    branches: [master]
+    branches:
+      - main
+      # Temporary inclusion of master branch during default branch name transition
+      - master
```
kenibrewer, line 16: @gwaybio @d33bs  I've made preliminary changes to allow the tests for this PR to run against both `master` and `main` branches. However, I think the best way to complete the transition would be for you to create a new `main` branch now and for me to change the merge target of this PR. Then after the merge to main is completed, changing the default branch. 

Does that sound like a good plan to you two?

tests/test_cyto_utils/test_DeepProfiler_processing.py
kenibrewer: I have changed the description of #304 to represent the primary bug which is being addressed here, which is the crashing of pytest during test collection. 

I created a separate [issue ](https://github.com/cytomining/pycytominer/issues/313) for the crashes that occur on smaller machines which will require more substantial re-factoring of `test_DeepProfiler_processing.py`.

kenibrewer: This re-factoring is completed done.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[tool.poetry]
+name = "pycytominer"
+version = "0.3.0"
+description = "Python package for processing image-based profiling data"
+authors = [
+    "Gregory Way <gregory.way@gmail.com>",
+]
+maintainers = [
+    "Gregory Way <gregory.way@gmail.com>",
+]
```
kenibrewer, line 10: @gwaybio Could you provide the names of the authors and maintainers you would like to include here in the pyproject.toml? Long-term, @d33bs suggested finding a way to auto-populate this from the CITATION.cff, but in the meantime we just want the current, up-to-date list.

tests/test_cyto_utils/test_DeepProfiler_processing.py
d33bs: Thank you @kenibrewer for the diligence on these items!

.github/workflows/python-app.yml
d33bs: @kenibrewer great thoughts! Given your changes here I propose we merge as-is with `master` as the target (and remaining in the actions spec as a branch reference) then afterwards rename `master` to `main` referencing [GitHub's guidance on renaming branches](https://github.com/github/renaming#renaming-existing-branches). After the rename at some point a small PR could be issued to remove `master` references within for GitHub Actions or elsewhere. Thoughts?

Either way, doing this may incur confusion for in-flight PR's so we should be sure to notify those with open or draft state PR's about the change just in case anything unusual happens. Once you merge I can take responsibility for renaming the branch and sending along notifications to the open or draft PR's (granted this all sounds good).

CC @gwaybio

d33bs (changes requested):
Nice work @kenibrewer ! I left a couple comments about the additional changes added. I'm marking this as requesting changes due to what I felt like might be an unclosed database connection (which could result in unexpected behavior) - otherwise things looked good to me. Given the changes to `collate.py` I recommend including those details upfront in the PR title or description to help observe and reason through what historically changed in this PR.

.github/workflows/python-app.yml
kenibrewer: @d33bs I like your plan a lot better than my suggestion. Let's proceed with that approach.

pycytominer/cyto_utils/collate.py
```diff
@@ -118,43 +119,29 @@ def collate(
...
-            "CREATE INDEX IF NOT EXISTS plate_well_image_idx ON Image(Metadata_Plate, Metadata_Well);",
-        ]
-        run_check_errors(index_cmd_metadata)
+            cursor.execute(
+                f"""CREATE INDEX IF NOT EXISTS table_image_object_{eachcompartment.lower()}_idx 
+                            ON {eachcompartment}(TableNumber, ImageNumber, ObjectNumber);"""
+            )
+        cursor.execute(
+            "CREATE INDEX IF NOT EXISTS plate_well_image_idx ON Image(Metadata_Plate, Metadata_Well);"
+        )
```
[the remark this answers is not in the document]
kenibrewer: The main purpose for the transition from `subprocess.run(['sqlite3', ...])` to the `cursor.execute()` approach is because the `sqlite3` cli command isn't installed for some reason when using the poetry dependency manager. Instead of troubleshooting that issue, I figured it would be better to adopt a more pythonic approach. 

Despite the change in structure, I deliberately limited the changes execute the exact same SQL commands. I would prefer to address performance/storage considerations in the context of a different PR.

pycytominer/cyto_utils/collate.py
```diff
@@ -118,43 +119,29 @@ def collate(
             print(f"Ingesting {input_dir}")
         run_check_errors(ingest_cmd)
 
+        # Create a sqlite3 connection
+        connection = sqlite3.connect(cache_backend_file, isolation_level=None)
```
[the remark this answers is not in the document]
kenibrewer: Done.

pycytominer/cyto_utils/collate.py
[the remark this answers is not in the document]
kenibrewer: To clarify the problems with the `subprocess.run()` access to command line tools, it seems to be a corner case that involves the use of VSCode's pytest integration, poetry, and cli commands from extras packages. I confirmed that all the required tools are still properly installed in the actual environment created by poetry. 

I also changed the `cytominer-database` integration from the cli to import form for the same reason.

pyproject.toml
kenibrewer: I've copied over the authors from the CITATION.cff file. We can merge the updated author list in a separate PR. I've created [this issue](https://github.com/cytomining/pycytominer/issues/316) to track completion prior to 1.0 release.

pycytominer/cyto_utils/collate.py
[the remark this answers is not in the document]
kenibrewer: > Given the changes to collate.py I recommend including those details upfront in the PR title or description to help observe and reason through what historically changed in this PR.

This is done.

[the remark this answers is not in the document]
d33bs: Cheers, thanks for the added effort with implementing and documenting this here. Seeing these changes made an impression on me how much database management is happening in this area. To this effect, creating indexes and defragmentation is sometimes an unannounced expectation of audiences receiving data (often this is very much warranted to provide them the best possible chance of utility, among other things). All this said, I'm good with leaving things as is and greatly appreciate the movement towards the import-based methods.

d33bs (approved):

kenibrewer:
@d33bs @gwaybio This PR is ready to merge.

gwaybio:
Thanks for this contribution @kenibrewer !

I have one question regarding the collate changes. Is it still possible to use collate from the command line? 

We have collaborators who use collate this way, and if we update collate usage, then they will not benefit from future pycytominer improvements. Is there a way to keep the collate CLI for now? FWIW, we believe CytoTable will supercede collate in the very near term, so we can stop supporting collate in future pycytominer versions.

kenibrewer:
> I have one question regarding the collate changes. Is it still possible to use collate from the command line?

@gwaybio Nothing about the changes to `collate.py` make `collate_cmd.py` unable to be used from the command line. Collate simply invokes fewer command line programs via `subprocess.run` and uses python functions instead when it gets invoked.

kenibrewer:
Perhaps I should clarify. The problem the `collate.py` changes were meant to address was only encountered when using the vscode pytest integration to run tests. I believe the root cause is that vscode doesn't fully initialize the poetry virtual environment. During this special vscode pytest, all the python modules are properly available via imports, but external command line tools are not added to the path. This causes issues when the python code executes a `subprocess.run()` command expecting to find an external cli that is only present in the poetry venv. 

I also went ahead and double checked that all the CLI tools that vscode pytest couldn't find, were actually properly installed and available in the PATH when the poetry venv was properly initialized from the command line or when the built package was installed via pip.

gwaybio (approved):
Thanks for this explanation @kenibrewer ! And thank you for this important contribution :)

@d33bs - please feel free to merge when you are ready.

tests/test_cyto_utils/test_collate.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        with pytest.raises(AssertionError):
+            pd.testing.assert_frame_equal(test, main)
+    return test, main
+
+
+def test_base_case():
+    cleanup()
+
+    collate(
```
d33bs, line 46: @kenibrewer - apologies for not seeing this sooner, but I wanted to raise an issue I noticed which seems to pertain to the changes with `collate.py` and how it is used by `collate_cmd.py`. In attempting to replicate this test using the command line through the poetry environment I'm seeing an error. I tested this same command with the master branch and things seem to work without error. Apologies in advance if I've somehow confused how this is implemented.

- **command**: (from within the pycytominer dir) `poetry run python pycytominer/cyto_utils/collate_cmd.py "2021_04_20_Target2" "pycytominer/cyto_utils/database_config/ingest_config.ini" "BR00121431" --base pycytominer/tests/test_data/collate --tmp-dir pycytominer/tests/test_data/collate --dont-add-image-features --dont-print`
- **error**: `sqlite3.OperationalError: no such table: main.Image`. When I inspected the SQLite output I found it was a seemingly 0 table file. Maybe this means cytominer-database was unable to complete the ingest?

kenibrewer: @d33bs There error you are experiencing is due to tests being moved from `/pycytominer/tests` to `/tests` . If you run the command below everything works as expected.
```
poetry run python pycytominer/cyto_utils/collate_cmd.py "2021_04_20_Target2" "pycytominer/cyto_utils/database_config/ingest_config.ini" "BR00121431" --base tests/test_data/collate --tmp-dir tests/test_data/collate --dont-add-image-features --dont-print
```

In the process of researching this possible bug, I did discover an [important detail](https://docs.python.org/3/library/sqlite3.html#how-to-use-the-connection-context-manager) about the sqlite3 connection context manager. Specifically, it does not automatically close connections after it leaves the with statement. Therefore, I added an explicit `connection.close()` to the code.

d33bs: Thanks and great find! I tested your command and things worked properly on my end.

d33bs:
Thank you so much @kenibrewer for the tremendous contributions in this PR! I feel good about how things look and would like to merge.

An aside: in the (I hope) kindest possible way I wanted to offer a suggestion for future work here or elsewhere (I have no qualms as is and these comments are intended as continuous improvement). There were a number of ideas in this PR which formed a larger and less modular group of code "chunks" to review (and historically understand). From a development standpoint, this is completely expected; when working on one thing we often can see other related things that need to get fixed.

When a "big enough" (totally subjective to the project and developer community) related but somewhat isolated chunk of work like this comes up it can be a good idea to spur a "fix / related" branch and PR, updating the original feature branch after the "fix" is merged to help keep things modular and isolated. While this can take longer, it can help isolate the changes for the developer(s), reviewers, and those looking at the code in the future to understand why things changed. It also might effect how other in-flight PR's and feature branches grow. When we isolate the discussion around singular and smaller chunks of code this way we offer ourselves a chance to be more careful, deliberate, and descriptive about the changes taking place.

A visual of how this might look would be the following:
```mermaid
    %%{init: { 'logLevel': 'debug', 'theme': 'default' , 'themeVariables': {
      'git0': '#6366F1',
      'git1': '#10B981',
      'git2': '#F59E0B',
      'gitBranchLabel1': '#ffffff',
      'gitBranchLabel2': '#ffffff'
} } }%%
    gitGraph
       commit id: "..."
       commit id: "original state"
       branch feature
       checkout feature
       commit id: "discover new needs"
       checkout main
       branch related_fix
       checkout related_fix
       commit id: "fix those needs"
       checkout main
       merge related_fix id: "merged fix (main)"
       checkout feature
       merge main id: "merged fix (feature)"
       checkout main
       merge feature id: "merge feature"
```

Only offering this up as a suggestion for future consideration. Thank you again so much for your thoughtful and thorough contributions to Pycytominer!

kenibrewer:
> When a "big enough" (totally subjective to the project and developer community) related but somewhat isolated chunk of work like this comes up it can be a good idea to spur a "fix / related" branch and PR, updating the original feature branch after the "fix" is merged to help keep things modular and isolated.

Thanks for this suggestion! This is definitely one of the more complicated software PR's I've worked on, so it is helpful to know about the best practices I can apply for future work.

PR #311, opened 2023-07-17
Make sample argument uniform for normalize and feature_select

gwaybio:
# Description

Currently, the `samples` argument in the `normalize` function expects a string to be used in a `pandas.DataFrame().query()` call. However ,the `samples` argument in `feature_select()` (and all associated operations) expects the `samples` argument to be a list to be used in a `pandas.DataFrame().loc[]` call.

Here, I update all `feature_select()` operations to expect the `pandas.DataFrame.query()` call. I also update the documentation and tests.

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -152,7 +152,7 @@ def drop_outlier_features(
         DataFrame that includes metadata and observation features.
     features : list of str or str, default "infer"
         Features present in the population dataframe. If "infer", then assume Cell Painting features are those that start with "Cells_", "Nuclei_", or "Cytoplasm_"
-    samples : list or str, default "all"
+    samples : str, default "all"
         List of samples to perform operation on. The function uses a pd.query()
         function, so you should  structure samples in this fashion. An example is
         "Metadata_treatment == 'control'" (include all quotes).
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -153,7 +153,7 @@ def drop_outlier_features(
     features : list of str or str, default "infer"
         Features present in the population dataframe. If "infer", then assume Cell Painting features are those that start with "Cells_", "Nuclei_", or "Cytoplasm_"
     samples : str, default "all"
-        List of samples to perform operation on. The function uses a pd.query()
+        List of samples to perform operation on. The function uses a pd.DataFrame.query()
         function, so you should  structure samples in this fashion. An example is
         "Metadata_treatment == 'control'" (include all quotes).
         If "all", use all samples to calculate.
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -169,7 +169,7 @@ def drop_outlier_features(
 
     # Subset dataframe
     if samples != "all":
-        population_df = population_df.query(samples)
+        population_df.query(samples, inplace=True)
 
     if features == "infer":
         features = infer_cp_features(population_df)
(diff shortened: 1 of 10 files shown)


pycytominer/cyto_utils/features.py
d33bs, line 1: Seeing how the patterns with `pd.DataFrame.query` repeat in this and the other functions, consider using a utility or decorator function to decrease the amount of repeated code and documentation. A decorator might work nicely to enable both documentation and functionality for these and any other functions which may need these capabilities in the future. See the following example for how this could work:

```python
import pandas as pd


def enable_pd_df_query(func):
    """
    A decorator for enabling pd.DataFrame.query()
    """

    def wrapper(df: pd.DataFrame, samples: str = "all", *args, **kwargs):
        df = df.query(samples) if samples != "all" else df
        return func(df, samples, *args, **kwargs)

    return wrapper


@enable_pd_df_query
def some_function(df: pd.DataFrame, samples: str = "all"):
    """
    A function to demonstrate decorator functionality
    """

    return df


df = pd.DataFrame({"A": range(1, 6), "B": range(10, 0, -2), "C C": range(10, 5, -1)})
some_function(df=df, samples="A > 1")
```

pycytominer/cyto_utils/features.py
```diff
@@ -166,7 +169,7 @@ def drop_outlier_features(
 
     # Subset dataframe
     if samples != "all":
-        population_df = population_df.loc[samples, :]
+        population_df = population_df.query(samples)
```
d33bs, line 172: Would it make sense to use `population_df.query(...,inplace=True)` here to help conserve memory usage (avoiding the production of a copied dataframe)?

d33bs (approved):
Nice work! Kudos also on the increased test coverage percentage. I thought things looked good with this PR and only had suggestions for enhancements. Generally I left comments in `pycytominer/cyto_utils/features.py` which apply to multiple files with the `pd.DataFrame.query()` repeated blocks. Please don't hesitate to let me know if you have any questions.

pycytominer/cyto_utils/features.py
```diff
@@ -152,8 +152,11 @@ def drop_outlier_features(
         DataFrame that includes metadata and observation features.
     features : list of str or str, default "infer"
         Features present in the population dataframe. If "infer", then assume Cell Painting features are those that start with "Cells_", "Nuclei_", or "Cytoplasm_"
-    samples : list of str or str, default "all"
-        Samples to perform the operation on
+    samples : str, default "all"
```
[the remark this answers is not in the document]
gwaybio: I think for our community, "samples" makes more sense than "query". Although, "samples_query" is probably the best :)

I think for now, we should keep the samples argument for historical reasons. We can explore a deprecation warning or alias in the future, but I think for now it is worth more trouble than benefit.

pycytominer/cyto_utils/features.py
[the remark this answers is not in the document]
gwaybio: While it reduces a line, it seems harder to read (for me at least).

Thank you for the suggestion, but I think i will skip this one! Happy to change it in the future, if there is consensus in the contrary direction.

gwaybio: I will make this change in the next commit! Thanks! (I hadn't realized `inplace` conserves memory, and historically, I haven't liked using `inplace` in pandas, probably for the wrong reasons)

pycytominer/cyto_utils/features.py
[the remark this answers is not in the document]
gwaybio: added in d9b06fe46f6f3b91bad9541224e7d67f6a1b9326

pycytominer/cyto_utils/features.py
gwaybio: added in d9b06fe46f6f3b91bad9541224e7d67f6a1b9326

pycytominer/cyto_utils/features.py
gwaybio: this is a really elegant solution... I like this a lot, but I think this is best left for a future re-haul of the entire code base. It would be great to get all instances of repeated code into this format. I do think this is beyond scope here, given our timeframe for version 1 release. I'll create an issue to make sure we document this for future work

Issue #179, opened 2021-12-03
Consider Using SVD over eigh in ZCA for numerical stability


johnarevalo:
https://github.com/cytomining/pycytominer/blob/9a55470be3de1b2bf54ebf78c892e2929bc4203f/pycytominer/operations/transform.py#L80

There is some evidence that `eigh` implementation could be less stable than `svd` algorithms in LAPACK.

https://stats.stackexchange.com/a/314062

gwaygenomics:
have you observed instability in this function in your work? Also, do you know if there are any speed tradeoffs?

PR #320, opened 2023-09-09
Fix Spherize – use SVD, simplify calculations

shntnu:
# Description

Make several updates to `Spherize`

- Use SVD for sphering instead of first computing covariance matrix followed by eigenvalue decomposition. This notebook demonstrates the equivalence [svd-eig.pdf](https://github.com/cytomining/pycytominer/files/12566840/FireShot.Capture.017.-.test.-.pdf).
- Use `epsilon` to regularize, not clip (the former is more standard; unclear why we were doing the latter)
- `PCA-cor` and `ZCA-cor` necessary center the data, so raise an error if centering is `False`
- Handle wide matrices 

# Checklist


pycytominer/normalize.py
```diff
@@ -162,14 +162,26 @@ def normalize(
         fitted_scaler = scaler.fit(profiles.query(samples).loc[:, features])
 
     # Scale the feature dataframe
+    # See https://github.com/cytomining/pycytominer/issues/319
```
gwaybio, line 165: Amazing to catch this Shantanu! Thanks for this proposed fix.

I wonder if a simpler solution is to output the numpy array from `spherize()` directly. i.e., the return line at https://github.com/shntnu/pycytominer/blob/33cbb30f8030fddc252d3f470358df289035f977/pycytominer/operations/transform.py#L165

should simply be `return X @ self.W` (or what you have edited to include, I haven't looked yet; basically, we should return the values instead of a DataFrame, which might be a simpler fix. Oh, but maybe this still breaks b/c the columns will be principal components and not `feature_df.columns`?

pycytominer/normalize.py
```diff
@@ -162,14 +162,26 @@ def normalize(
...
 
     normalized = meta_df.merge(feature_df, left_index=True, right_index=True)
 
+    assert (
+        feature_df.shape == profiles.loc[:, features].shape
+    ), "Feature dataframe shape does not match original dataframe shape"
+
+    assert (normalized.shape[0] == profiles.shape[0]) and (
+        normalized.shape[1] == len(features) + len(meta_features)
+    ), "Normalized dataframe shape does not match original dataframe shape"
```
gwaybio, line 183: I like these checks, but they could be frustrating errors for our users. Is there a way to be more specific about what to check if this assertion fails?

pycytominer/operations/transform.py
```diff
@@ -54,6 +55,11 @@ def __init__(self, epsilon=1e-6, center=True, method="ZCA"):
         ), f"Error {method} not supported. Select one of {avail_methods}"
         self.method = method
 
+        # PCA-cor and ZCA-cor require center=True
+        assert (
+            self.method not in ["PCA-cor", "ZCA-cor"] or self.center
+        ), "PCA-cor and ZCA-cor require center=True"
```
gwaybio, line 61:
```suggestion
        if self.method in ["PCA-cor", "ZCA-cor"]:
            assert(self.center), "PCA-cor and ZCA-cor require center=True"
```

This feels more readable to me

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
...
-            # process the covariance diagonal matrix and fix rounding errors
-            v = np.diag(1.0 / np.sqrt(np.diag(C).clip(self.epsilon)))
+        self.W = (Vt / Sigma[:, np.newaxis]).transpose() * np.sqrt(n - 1)
 
-            # Calculate the sphering matrix
-            self.W = np.dot(np.dot(D, G.transpose()), v)
+        # If ZCA, perform additional rotation
+        if self.method in ["ZCA", "ZCA-cor"]:
+            # If rank is not d then ZCA is not possible
+            # TODO: Explain this better
```
gwaybio, line 124: I fear that this would take a while to revisit after we merge, would it take too long to explain?

pycytominer/normalize.py
gwaybio: Oh, I think I understand. We want to use `Spherize()` standalone. I wonder then if it is more pythonic to add an argument like `Spherize(return_numpy=False)` and then set `normalize()` internals to use `return_numpy=True` in `Spherize` internals. What do you think?

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
...
 
-            # Process the eigenvalues into a diagonal matrix and fix rounding errors
-            D = np.diag(1.0 / np.sqrt(s.clip(self.epsilon)))
+            X = self.standard_scaler.transform(X)
+        else:
+            if self.center:
+                self.mean_centerer = StandardScaler(with_mean=True, with_std=False).fit(
+                    X
+                )
+                X = self.mean_centerer.transform(X)
```
gwaybio, line 97: ls there a reason to not set X as `X = StandardScaler(with_mean=True, with_std=False).fit_transform(X)`?

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
...
-            # If ZCA, perform additional rotation
-            if self.method == "ZCA":
-                self.W = np.dot(U, self.W)
+        # compute the rank of the matrix X
+        r = np.linalg.matrix_rank(X)
 
-        if self.method in ["PCA-cor", "ZCA-cor"]:
-            # Get the correlation matrix
-            R = np.corrcoef(X.transpose())
+        assert (r == n - 1) | (r == d), "X is not full rank"
```
gwaybio, line 105: This error message could be bolstered as well to reduce user frustration. Is there somethign we could recommend a user do to make it work?

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
...
+        # Get the eigenvalues and eigenvectors of the covariance matrix using SVD
+        _, Sigma, Vt = np.linalg.svd(X, full_matrices=True)
 
-            # Fix sign ambiguity of eigenvectors
-            G = pd.DataFrame(G * np.sign(np.diag(G)))
+        # if n <= d then Sigma has shape (n,) so it will need to be expanded to
+        # d filled with the value r'th element of Sigma
+        if n <= d:
+            assert Sigma.shape[0] == n, "Unexpected shape of Sigma"
+            assert r == n - 1, "Unexpected rank"
```
gwaybio, line 114: Same comment about these error messages

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
...
+            # If rank is not d then ZCA is not possible
+            # TODO: Explain this better
+            assert r == d, "ZCA is not possible if rank is not d"
+            self.W = self.W @ Vt
 
-            # If ZCA-cor, perform additional rotation
-            if self.method == "ZCA-cor":
-                self.W = np.dot(G, self.W)
+        # number of columns of self.W should be equal to that of X
+        assert (
```
gwaybio, line 129: will this assertion ever be triggered? In other words, can we test it?

tests/test_normalize.py
```diff
@@ -445,6 +445,8 @@ def test_normalize_standardize_allsamples_compress():
 def test_normalize_spherize():
     for spherize_method in ["ZCA", "PCA", "ZCA-cor", "PCA-cor"]:
         for spherize_center in [True, False]:
+            if spherize_method in ["PCA-cor", "ZCA-cor"] and not spherize_center:
+                continue
```
gwaybio, line 449: Can you add a test that enters the assertion? This is a great spot for it!

tests/test_operations/test_transform.py
```diff
@@ -22,6 +22,8 @@ def test_spherize():
     spherize_methods = ["PCA", "ZCA", "PCA-cor", "ZCA-cor"]
     for method in spherize_methods:
         for center in [True, False]:
+            if ["PCA-cor", "ZCA-cor"] and not center:
+                continue
```
gwaybio, line 26: Can you add a test that enters the assertion? This is a great spot for it!

gwaybio (changes requested):
Glad to see this fix, especially since the tests pass!

I made several inline comments, which we should discuss prior to merging. My general comments include:

1. Please add tests for entry into the assertion that centering must be performed
2. We really should also add testing conditions for the other assertions
3. The other assertion statement errors need more user-friendly (read, less potential for frustration) text. Can we prompt a user on what to fix if they run into one?
4. I'm wondering if an easier fix is to optionally return the numpy array rather than the DataFrame in `Spherize()` itself. 

I made more specific comments for each of these four general comments in line below.

Thanks again!

gwaybio:
Oops! @shntnu - I just saw the "removed the request" and WIP tag. Feel free to consider my comments as preliminary or to ignore them

johnarevalo:
This PR fixes #179

pycytominer/operations/transform.py
```diff
@@ -67,58 +73,62 @@ def fit(self, X, y=None):
         self
             With computed weights attribute
         """
-        # Get the mean of the features (columns) and center if specified
-        self.mu = X.mean()
```
johnarevalo, line 71: This should be X.mean(axis=0) or assert X is pd.DataFrame object. If X is numpy array, then X.mean() will be a scalar(global mean)

johnarevalo: nvm 😅, X.values line below is implicitly assuming X is a dataframe.

pycytominer/normalize.py
shntnu: The purpose of this assertion is really to identify new test fixtures, because this indicates a bug in software. I now prompt the user to file an issue.

pycytominer/operations/transform.py
shntnu: Thanks. I've made these more readable and also changed them to exceptions

pycytominer/operations/transform.py
shntnu: Ah because I use `self.mean_centerer` later in `Spherize.transform`

```
if self.center:
    X = self.mean_centerer.transform(X)
```

Makes sense?

pycytominer/operations/transform.py
shntnu: More explanation added

pycytominer/normalize.py
shntnu: What a great idea! This should work

shntnu: Works!

pycytominer/operations/transform.py
shntnu: Right, I do this within `fit`

```
        # Get Numpy representation of the DataFrame
        X = X.values
```

Maybe I should have renamed the variance but keeping it `X` makes it cleaner

shntnu: The new changes have made this outdated so I will resolve

tests/test_normalize.py
shntnu: nice

pycytominer/operations/transform.py
shntnu: Now explained in more detail

tests/test_operations/test_transform.py
shntnu: I do this in `test_normalize_spherize` instead

pycytominer/operations/transform.py
shntnu: It may be triggered if `np.linalg.svd` ever changes behavior. We cannot test it (easily) because it will involve mocking a version of `np.linalg.svd` that has this behavior.  I use `create_bug_report_message` here, which asks the user to report a bug.

pycytominer/operations/transform.py
shntnu: Turns out this was very likely an incorrect requirement 
I've added this comment in the code for our future selves
```
            # Note: There was previously a requirement r==d otherwise the
            # ZCA transform would not be well defined. However, it later appeared
            # that this requirement was not necessary.
```

pycytominer/normalize.py
shntnu: I now use `create_bug_report_message`

gwaybio (commented):
Thanks for updating this code Shantanu!

I'm going to propose the following to get this merged in ASAP

- @mattsoncam - can you focus on the math side of things? The current tests pass so we can be confident about accuracy, but it would be great for you to read for understanding and provide suggestions on where the documentation can be upgraded.
- @kenibrewer - can you focus on the integration aspects? The added bug report function in `util.py` is neat, but you have your finger on the pulse of the pycytominer roadmap. Can you provide your thoughts here?

I think this can be a relatively quick PR merge, which I will do once we have these two approvals. Sorry about the delays Shantanu, we're right in the middle of me phasing out of PR review in favor of much more knowledgeable/qualified maintainers :)

kenibrewer (dismissed):
Would it be possible to move the new `create_bug_report_message` function into a separate branch and pull request to be reviewed separately? There are a few aspects about the new pattern that merit discussion and I would prefer to keep this PR focused specifically on the spherize/SVD changes. 

EDIT: To clarify. I agree with Greg that the feature has value, but it would be good to have a plan (and some documentation) about when pycytominer wants to "own" the error message as a bug and when we want to consider it data/user error and instead provide instructions on how to fix the problem.

pycytominer/operations/transform.py
```diff
@@ -43,17 +45,25 @@ def __init__(self, epsilon=1e-6, center=True, method="ZCA"):
...
-            method in avail_methods
-        ), f"Error {method} not supported. Select one of {avail_methods}"
+        if method not in avail_methods:
+            raise ValueError(
+                f"Error {method} not supported. Select one of {avail_methods}"
+            )
         self.method = method
 
+        # PCA-cor and ZCA-cor require center=True
+        if self.method in ["PCA-cor", "ZCA-cor"] and not self.center:
```
MattsonCam, line 64: This is a good approach. Could also just change self.center, but this may be better for informing the user.

pycytominer/operations/transform.py
```diff
@@ -67,58 +77,94 @@ def fit(self, X, y=None):
...
 
-        if self.method in ["PCA-cor", "ZCA-cor"]:
-            # Get the correlation matrix
-            R = np.corrcoef(X.transpose())
+        # If n < d, then rank should be equal to n - 1 (if centered) or n (if not centered)
+        # If n >= d, then rank should be equal to d
+        if not ((r == d) | (self.center & (r == n - 1)) | (not self.center & (r == n))):
+            raise ValueError(
+                (
+                    "Sphering is not supported when the data matrix X is not full rank."
```
MattsonCam, line 114: Consider mentioning why sphering is not supported in this case.

pycytominer/operations/transform.py
```diff
@@ -67,58 +77,94 @@ def fit(self, X, y=None):
...
-            # If ZCA, perform additional rotation
-            if self.method == "ZCA":
-                self.W = np.dot(U, self.W)
+        # compute the rank of the matrix X
+        r = np.linalg.matrix_rank(X)
 
-        if self.method in ["PCA-cor", "ZCA-cor"]:
-            # Get the correlation matrix
-            R = np.corrcoef(X.transpose())
+        # If n < d, then rank should be equal to n - 1 (if centered) or n (if not centered)
```
MattsonCam, line 109: Consider mentioning why the rank should follow these properties

pycytominer/operations/transform.py
```diff
@@ -43,17 +45,25 @@ def __init__(self, epsilon=1e-6, center=True, method="ZCA"):
...
-        assert (
-            method in avail_methods
-        ), f"Error {method} not supported. Select one of {avail_methods}"
+        if method not in avail_methods:
+            raise ValueError(
+                f"Error {method} not supported. Select one of {avail_methods}"
+            )
         self.method = method
 
+        # PCA-cor and ZCA-cor require center=True
```
MattsonCam, line 63: Consider adding a source/documentation for why this is required.

MattsonCam (commented):
Very detailed pr. I left a few comments to consider. Great job!

pycytominer/cyto_utils/util.py
```diff
@@ -358,3 +358,22 @@ def get_pairwise_correlation(population_df, method="pearson"):
...
 
     return data_cor_df, pairwise_df
+
+
+def create_bug_report_message(error_detail, context):
+    """
+    Constructs an error message with a bug report prompt.
+
+    Parameters:
+    - error_detail (str): Description of the error.
```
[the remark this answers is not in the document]
shntnu: @kenibrewer has requested moving this into a new PR, so will do that there

pycytominer/normalize.py
```diff
@@ -161,15 +160,34 @@ def normalize(
...
+            "Feature dataframe shape does not match original dataframe shape."
+        )
+        context = f"the `{method}` method in `pycytominer.normalize`"
+        raise ValueError(util.create_bug_report_message(error_detail, context))
+
+    if (normalized.shape[0] != profiles.shape[0]) or (
+        normalized.shape[1] != len(features) + len(meta_features)
+    ):
+        error_detail = (
+            "Normalized dataframe shape does not match original dataframe shape."
```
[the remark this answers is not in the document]
shntnu: I have changed it to `The number of rows and columns in the normalized dataframe does not match the original dataframe` because it needn't be cells (could be well, for example), and we also check the number of columns (not just rows)

pycytominer/operations/transform.py
shntnu: Thanks for the nudge

pycytominer/operations/transform.py
shntnu: Ah yes, good point. I'd favor erroring out in this case.

pycytominer/operations/transform.py
```diff
@@ -67,58 +77,94 @@ def fit(self, X, y=None):
...
+        if not ((r == d) | (self.center & (r == n - 1)) | (not self.center & (r == n))):
+            raise ValueError(
+                (
+                    "Sphering is not supported when the data matrix X is not full rank."
+                    "Check for linear dependencies in the data and remove them."
+                )
+            )
+
+        # Get the eigenvalues and eigenvectors of the covariance matrix using SVD
+        _, Sigma, Vt = np.linalg.svd(X, full_matrices=True)
```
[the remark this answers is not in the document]
shntnu: Good catch; `X` is the data matrix. I now cite https://stats.stackexchange.com/q/134282/8494

pycytominer/operations/transform.py
```diff
@@ -67,58 +77,94 @@ def fit(self, X, y=None):
...
+
+        # if n <= d then Sigma has shape (n,) so it will need to be expanded to
+        # d filled with the value r'th element of Sigma
+        if n <= d:
+            # Do some error checking
+            if Sigma.shape[0] != n:
```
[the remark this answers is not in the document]
shntnu: If n > d (long matrix) then Sigma would be dxd -- is that your question?

pycytominer/operations/transform.py
shntnu: Great catch!

Sphering is indeed supported in this case, but it's just highly unlikely that we'd end up with a perfect linear dependency, so we should not proceed. I have updated the error message. LMK if this suffices.

pycytominer/operations/transform.py
shntnu: Now added

```
        # Case 1: More features than samples (n < d).
        # If centered (mean of each feature subtracted), one dimension becomes dependent, reducing rank to n - 1.
        # If not centered, the max rank is limited by n, as there can't be more independent vectors than samples.
        # Case 2: More samples than features or equal (n >= d).
        # Here, the max rank is limited by d (number of features), assuming each provides unique information.
```

pycytominer/operations/transform.py
```diff
@@ -67,58 +77,94 @@ def fit(self, X, y=None):
...
-            G = pd.DataFrame(G * np.sign(np.diag(G)))
+            Sigma = np.concatenate((Sigma[0:r], np.repeat(Sigma[r - 1], d - r)))
 
-            # Process the eigenvalues into a diagonal matrix and fix rounding errors
-            D = np.diag(1.0 / np.sqrt(t.clip(self.epsilon)))
+        Sigma = Sigma + self.epsilon
 
-            # process the covariance diagonal matrix and fix rounding errors
-            v = np.diag(1.0 / np.sqrt(np.diag(C).clip(self.epsilon)))
+        self.W = (Vt / Sigma[:, np.newaxis]).transpose() * np.sqrt(n - 1)
```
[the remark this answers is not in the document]
shntnu: Thanks for the nudge; I've now written it out

shntnu:
> Would it be possible to move the new `create_bug_report_message` function into a separate branch and pull request to be reviewed separately?

Makes sense, but instead, I'll just copy the code snippet here because I bet you'd want to do this more systematically later

<details>

```py
def create_bug_report_message(error_detail, context):
    """
    Constructs an error message with a bug report prompt.

    Parameters
    ----------
    error_detail: str
        Description of the error.
    context: str
        A description of the context for the error.
    Returns
    -------
    str
        Error message with a link to file a bug report.
    """

    bug_report_url = "https://github.com/cytomining/pycytominer/issues"
    return (
        f"{error_detail} This is likely a bug in `{context}`. "
        "Please help us improve our software by filing a bug report at "
        f"{bug_report_url}."
    )


def test_create_bug_report_message():
    error_detail = "Division by zero error occurred."
    context = "calculate_statistics function"
    actual_message = create_bug_report_message(error_detail, context)

    # check that the strings `error_detail` and `context` are in the message
    assert error_detail in actual_message
    assert context in actual_message

    # check that the message contains a link to file a bug report
    assert "https://github.com/cytomining/pycytominer/issues" in actual_message

```

shntnu:
> Very detailed pr. I left a few comments to consider. Great job @shntnu !

Thank you for nudging me to clarify the math further @MattsonCam! I understood my code better after explaining it :D I've resolved all your comments but feel free to reopen if necessary.

gwaybio:
It would be great to get this fix merged in. 

I just chatted with @MattsonCam and he will give a 👍 

@kenibrewer - is this ready from your perspective?

pycytominer/operations/transform.py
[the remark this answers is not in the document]
MattsonCam: The documentation looks great; I didn't know numpy implemented computing the whitening matrix. May also consider providing the numpy link

pycytominer/operations/transform.py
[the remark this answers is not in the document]
MattsonCam: Oh, that makes sense. You are making sure there is no bug when implementing numpy's svd function. If you're returning the full matrices, then Sigma should be (nxd), even if it's sparse, from my understanding

pycytominer/operations/transform.py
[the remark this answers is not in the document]
MattsonCam: Ahh, I see, thanks for the resource

pycytominer/operations/transform.py
MattsonCam: This looks like it will suffice!

pycytominer/operations/transform.py
MattsonCam: That looks good!

pycytominer/operations/transform.py
MattsonCam: Saw the comment, well done

MattsonCam:
> > Very detailed pr. I left a few comments to consider. Great job @shntnu !
> 
> Thank you for nudging me to clarify the math further @MattsonCam! I understood my code better after explaining it :D I've resolved all your comments but feel free to reopen if necessary.

Absolutely! Thank you for addressing them @shntnu! I've finished addressing your response comments, but didn't feel it was imperative to reopen any of them. 

> It would be great to get this fix merged in.
> 
> I just chatted with @MattsonCam and he will give a 👍 when he gets a chance
> 
> @kenibrewer - is this ready from your perspective?

I give the 👍

MattsonCam (approved):

pycytominer/operations/transform.py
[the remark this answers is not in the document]
shntnu: Ah, `numpy` does not implement computing the whitening matrix. Maybe you are referring to the comment "In NumPy, this is implemented as"? That's referring to`W = (np.linalg.inv(S) @ Vt) * np.sqrt(n - 1)` but you still need to compute `Vt`, plus computing `np.linalg.inv(S)` explicitly is too slow/memory intensive. I have now updated that comment.

shntnu:
> I give the 👍

Hooray! 

Over to one of you to merge once the tests pass (I made a documentation edit)

gwaybio:
this looks good to me @shntnu - I am going to merge this in now. Thanks for this contribution and for your patience as we reviewed! 🎉

gwaybio (approved):

gwaybio:
@kenibrewer - thanks for your initial review. Given @MattsonCam 's approval (and your initial review focused on the bug report document now described in #365 ) - I will merge this in now to avoid additional delays. Thanks!

kenibrewer:
Sorry I missed the pings here. I had modified my github notifications settings incorrectly and didn't get the usual notification emails I rely on. 

Everything looks good to me.

gwaybio:
No worries!

shntnu:
Hooray! Thank you all.

PR #373, opened 2024-02-20
ci: add versioned artifact build action

### LINKED ISSUE #369: FeatureRequest: Add build artifacts for CI/CD pipeline
### Feature type

- [X] Add new functionality

- [ ] Change existing functionality

Add additional steps to Github Actions workflow that will build a properly-versioned code artifact using `poetry build` instead of the "version-less" copy of the source code currently generated. These code artifacts can be used in the following ways:

1. Include new build artifacts in the Github release alongside the existing source code tarball. 
2. Drops the need for `poetry-dynamic-versioning` and a full git history in:
- Docker images 
- Dependent Github actions workflows. 

### Feature example

```yaml 
- step: build wheel/sdist
  run: poetry build
- step: upload-artifact 
  with: https://github.com/actions/upload-artifact

- step: build docker image
  with: download-artifact
- step: unpack artifact
- step: build docker image
```

### Alternative Solutions

_No response_

### Additional information

We should also add a section to `pyproject.toml` that includes files we want/need in our build artifact such as the `tests` directory. 

See: https://python-poetry.org/docs/pyproject/#include-and-exclude

kenibrewer:
# Description

This adds a versioned code artifact to the ci pipeline that can be used during other actions and in Github releases.

Closes #369 

# Checklist

- [ ] ~I have made corresponding changes to the documentation.~

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -136,9 +136,11 @@ jobs:
           name: wheel
           path: dist/*.whl
           if-no-files-found: error
+          retention-days: 90
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
           name: sdist
           path: dist/*.tar.gz
           if-no-files-found: error
+          retention-days: 90
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -38,8 +38,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
-          # Full history likely required for planned commitizen checks
-          # See #345
+          # Full history required for branch-based pre-commit checks
           fetch-depth: 0
       - name: Setup python, and check pre-commit cache
         uses: ./.github/actions/setup-env
@@ -105,3 +104,41 @@ jobs:
         # For every other version and/or OS, run pytest without coverage
         if: (matrix.os != 'ubuntu-latest') || (matrix.python-version != env.TARGET_PYTHON_VERSION )
         run: poetry run pytest ${{ github.event.inputs.pytest_addopts }}
+  build:
+    name: Build versioned package
+    # This job is used to build the package and upload the artifacts to GitHub Actions workflow results.
+    # See https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Setup python, and load cache
+        uses: ./.github/actions/setup-env
+        with:
+          python-version: ${{ env.TARGET_PYTHON_VERSION }}
+          cache-pre-commit: false
+          cache-venv: true
+          setup-poetry: true
+          install-deps: true
+      - name: Build
+        run: poetry build
+      - name: Get pycytominer version
+        id: get_version
+        run: |
+          echo "version=$(poetry version | cut -d' ' -f2 )" >> "$GITHUB_OUTPUT"
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheel
+          path: dist/*.whl
+          if-no-files-found: error
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: sdist
+          path: dist/*.tar.gz
+          if-no-files-found: error


.github/workflows/integration-test.yml
```diff
@@ -105,3 +104,41 @@ jobs:
...
+          install-deps: true
+      - name: Build
+        run: poetry build
+      - name: Get pycytominer version
+        id: get_version
+        run: |
+          echo "version=$(poetry version | cut -d' ' -f2 )"
+          echo "version=$(poetry version | cut -d' ' -f2 )" >> "$GITHUB_OUTPUT"
+      - name: Upload artifacts
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
```
d33bs, line 134: Consider using the semantic version specification for this action (instead of the hash-based reference) (here and below). If there's a reason for using the hash instead, consider adding a comment as to why.

.github/workflows/integration-test.yml
```diff
@@ -105,3 +104,41 @@ jobs:
...
+          python-version: ${{ env.TARGET_PYTHON_VERSION }}
+          cache-pre-commit: false
+          cache-venv: true
+          setup-poetry: true
+          install-deps: true
+      - name: Build
+        run: poetry build
+      - name: Get pycytominer version
+        id: get_version
+        run: |
```
d33bs, line 130: Is the duplicative echo focused command below used to help log the output in addition to using it with `$GITHUB_OUTPUT`? If so, consider adding a minor code comment somewhere near here about that usage pattern.

.github/workflows/integration-test.yml
```diff
@@ -105,3 +104,41 @@ jobs:
...
+          echo "version=$(poetry version | cut -d' ' -f2 )"
+          echo "version=$(poetry version | cut -d' ' -f2 )" >> "$GITHUB_OUTPUT"
+      - name: Upload artifacts
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
+        with:
```
d33bs, line 135: Consider setting an explicit retention period via `retention-days` (here and below) to help indicate how long the artifacts might be available. Depending on the usage intention, I wonder if non-default, shorter periods of time might be helpful to avoid exceeding the storage quota. Open to other thoughts here too.

.github/workflows/integration-test.yml
```diff
@@ -105,3 +104,41 @@ jobs:
         # For every other version and/or OS, run pytest without coverage
         if: (matrix.os != 'ubuntu-latest') || (matrix.python-version != env.TARGET_PYTHON_VERSION )
         run: poetry run pytest ${{ github.event.inputs.pytest_addopts }}
+  build:
+    name: Build versioned package
+    # This job is used to build the package and upload the artifacts
```
d33bs, line 109: Consider adding an explicit mention of the upload target to the comment here. It might also be useful to link to this anchor (which has a helpful screenshot): https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go

```suggestion
    # This job is used to build the package and upload the artifacts to GitHub Actions workflow results.
```

.github/workflows/integration-test.yml
```diff
@@ -38,8 +38,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
-          # Full history likely required for planned commitizen checks
-          # See #345
+          # Full history required for branch-based pre-commit checks
```
d33bs, line 41: Would it make sense to retain the mention about commitizen or dynamic versioning here (as a way of being explicit about why the line below is necessary)?

d33bs (commented):
Neat to see this, nice work! I left a few comments up for your consideration. Please don't hesitate to let me know if you have any questions / concerns.

.github/workflows/integration-test.yml
kenibrewer: Best I can tell from the Github documentation, there is no artifact storage limit for public repos. The storage limits in the table [here](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) for Github Free accounts only apply to private repos. 

I suggest we hold off on adding a retention policy until if/when we encounter issues.

.github/workflows/integration-test.yml
kenibrewer: Done.

.github/workflows/integration-test.yml
d33bs: Mostly here I suggest being as specific as we can when it comes to artifact retention and storage considerations. This is an element we can help reduce the [uncertainty of GitHub's mysteriousness in this area](https://github.com/actions/upload-artifact/issues/9) through explicit definition. The [default retention policy is 90 days](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy), and informing Pycytominer developers upfront I feel would be helpful if it became an issue of total storage considerations or planning for other tooling.

.github/workflows/integration-test.yml
d33bs: Good point, I can see how this might be helpful for security considerations! Candidly, I feel that GitHub Actions as an ecosystem are complex enough to warrant a "package manager" of some kind where the versions and reference locations are explicitly defined somewhere besides inline (especially when/if they are used many times). Would love to see if we could achieve something like this in the future for improved maintenance, security, and usability.

.github/workflows/integration-test.yml
d33bs: Thanks, makes sense!

.github/workflows/integration-test.yml
kenibrewer: Dependabot can also be configured to check your github action versions. That's probably the best cool that I'm aware of for doing that.

Issue #214, opened 2022-07-15
Docker container


bethac07:
I made a proof of principle one [here](https://github.com/cytomining/pycytominer-Docker) for my own purposes, but if you want, it would be < an hour or two of work to set this up as a GH action inside this repo so that `tag:latest` is built and pushed every time there is a commit to master and `tag:version` is pushed with every new official pycytominer version. You also may decide you want the Python version more (or less!) pinned or even matrixed.  

Probably also makes sense to make a cytomining Docker org, if Dockers are a thing you want.

d33bs:
Related notes and thoughts here: cytomining/pycytominer-Docker#2

d33bs:
A quick update here: I'm going to begin working towards addressing this issue through the creation of a Cytomining organization on Docker Hub (as @bethac07 originally suggested). We can seek support through the [Docker-Sponsored Open Source (DSOS) program](https://www.docker.com/community/open-source/application/) which will help avoid monetary costs normally associated with Docker Hub organizations. Creating this organization would be a first step here, providing a "place" for Cytomining images to be uploaded and used by others. Once we have an image registry location in place, we could add a Dockerfile and related code within this repository to create a Cytomining/Pycytominer image there.

Please don't hesitate to let me know if you have any questions / concerns!

CC @gwaybio, @kenibrewer

kenibrewer:
@d33bs Another option that we might consider is using biocontainers organization and tooling. There are some advantages for that route beyond just the fact that we wouldn't need to stand up our own organization. AWS, for example, caches all the biocontainer images locally within their Elastic Container Registry to provide fast access to those images to people running workflows on AWS.

d33bs:
Thanks @kenibrewer ! I took a look at Biocontainers and thought it looked promising. As a general advantage it seemed like the Biocontainers path might be more directly recognized by the bioinformatics community and timeline-wise faster in terms of finding a place for a Pycytominer image to land (rather than waiting for sponsorship / approval from Docker, which is an unknown). I'm generally unsure of where Pycytominer audiences obtain or implement their containerized work, so would trust you or others on guidance concerning the most valuable way to share Cytomining images.

[...]

PR #377, opened 2024-03-20
build(docker): add docker hub push capabilities

d33bs:
# Description

This PR continues the efforts surrounding #214 to push Docker images to Docker Hub using various triggers from GitHub Actions. I tried to abide the spec mentioned by @bethac07 (thanks Beth!) with slight adjustment to add the dynamic version to each tag along with the commit hash (which includes a commit hash for non-release-based versions). Please don't hesitate to let me know if I should change the tagging or other aspects to meet the goals here.

Open question I had while creating this: when would be best to create a scheduled push for Docker images? I chose Wednesdays as a time where challenges (if there were any) would be quite visible and could be resolved before week-end. I can also see how Mondays, Fridays, or sometime over the weekend would be good too.

Once this is merged, I believe we should see a new push triggered, and will likely be able to reach back out to Docker for more support surrounding their DSOS program.

Thank you in advance for any thoughts and suggestions you may have on this!

# Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches: [main]
     tags: ["*"]
+  schedule:
+    # runs every Wednesday at 7 AM UTC
+    - cron: "0 7 * * 3"
   workflow_dispatch:
     inputs:
       pytest_addopts:
@@ -34,6 +37,7 @@ jobs:
     strategy:
       fail-fast: false
     runs-on: ubuntu-latest
+    if: github.event_name != 'schedule'
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
@@ -69,6 +73,7 @@ jobs:
       # https://sqlalche.me/e/b8d9
       # We can remove this once we upgrade to SQLAlchemy >= 2.0
       SQLALCHEMY_SILENCE_UBER_WARNING: "1"
+    if: github.event_name != 'schedule'
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
@@ -144,7 +149,7 @@ jobs:
           path: dist/*.tar.gz
           if-no-files-found: error
           retention-days: 90
-  docker-image-test:
+  docker-image-test-and-push:
     runs-on: ubuntu-latest
     # set a dependency sequence to occur after build job completion
     needs: build
@@ -193,3 +198,48 @@ jobs:
       - name: Run tests through docker image
         run: |
           docker run pycytominer:testing pytest
+      # Gather and prepare the dynamic version and other vars for use with work below.
+      # For example, 1.1.0.post6.dev0+bc093ef becomes 1.1.0.post6.dev0_bc093ef
+      - name: Gather dynamic version and set env var
+        id: env-vars
+        run: >-
+          echo "CLEAN_VERSION=$(echo ${{ env.sdist_filename }} | tr '+' '_' )" >> "$GITHUB_OUTPUT" &&
+          echo "DATE_TAG=$(date +'%y%m%d')" >> "$GITHUB_OUTPUT"
+      - name: Login to Docker Hub
+        # only run this task if we have a pull request merged or the schedule triggers from main
+        if: github.event.pull_request.merged || ( github.event_name == 'schedule' && github.ref == 'refs/heads/main')
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      # Docker image push based on weekly schedule
+      - name: Schedule-based docker image push
+        # only run this task if we have a schedule trigger from main on cytomining/pycytominer
+        if: github.event_name == 'schedule' &&  github.ref == 'refs/heads/main' && github.repository == 'cytomining/pycytominer'
+        uses: docker/build-push-action@v5
+        with:
+          context: "${{ env.sdist_extracted_name }}"
+          push: true
+          file: ../build/docker/Dockerfile
+          target: production
+          tags: cytomining/pycytominer:${{ steps.env-vars.outputs.CLEAN_VERSION }}_${{ steps.env-vars.outputs.DATE_TAG }}
+      # Docker image push based on GitHub pushes to main
+      - name: Push-based docker image push
+        # only run this task if we have a pull request which is merged
+        if: github.event.pull_request.merged
+        uses: docker/build-push-action@v5
+        with:
+          context: "${{ env.sdist_extracted_name }}"
+          push: true
+          file: ../build/docker/Dockerfile
+          target: production
+          tags: cytomining/pycytominer:latest,cytomining/pycytominer:${{ steps.env-vars.outputs.CLEAN_VERSION }}
+      # update docker hub description using readme on merge
+      - name: Docker Hub
+        # only run this task if we have a pull request which is merged
+        if: github.event.pull_request.merged
+        uses: peter-evans/dockerhub-description@v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+          repository: cytomining/pycytominer
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -207,7 +207,7 @@ jobs:
           echo "DATE_TAG=$(date +'%y%m%d')" >> "$GITHUB_OUTPUT"
       - name: Login to Docker Hub
         # only run this task if we have a pull request merged or the schedule triggers from main
-        if: github.event.pull_request.merged || ( github.event_name == 'schedule' && github.ref == 'refs/heads/main')
+        if: (github.event.pull_request.merged ||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USER }}
@@ -226,7 +226,7 @@ jobs:
       # Docker image push based on GitHub pushes to main
       - name: Push-based docker image push
         # only run this task if we have a pull request which is merged
-        if: github.event.pull_request.merged
+        if: github.event.pull_request.merged && github.ref == 'refs/heads/main' && github.repository == 'cytomining/pycytominer'
         uses: docker/build-push-action@v5
         with:
           context: "${{ env.sdist_extracted_name }}"
@@ -237,7 +237,7 @@ jobs:
       # update docker hub description using readme on merge
       - name: Docker Hub
         # only run this task if we have a pull request which is merged
-        if: github.event.pull_request.merged
+        if: github.event.pull_request.merged && github.ref == 'refs/heads/main' && github.repository == 'cytomining/pycytominer'
         uses: peter-evans/dockerhub-description@v3
         with:
           username: ${{ secrets.DOCKER_USER }}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,6 +191,16 @@ Creating a new release includes the following steps:
 11. The release will be automatically published to [PyPI](https://pypi.org/project/pycytominer/) via Github Actions.
 12. Manually create the release at [conda-forge](https://anaconda.org/conda-forge/pycytominer).
 
+#### Docker Hub Image Releases
+
+We automate image pushes for `pycytominer` under the `cytomining` organization on [Docker Hub](https://hub.docker.com/) using GitHub Actions workflows.
+These pushes are defined within [.github/workflows/integration-test.yml](.github/workflows/integration-test.yml).
+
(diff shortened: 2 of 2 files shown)


.github/workflows/docker-hub-push.yml
kenibrewer, line 1: Having two separate actions for the "build and test" workflow and the "build-push" workflow requires us to maintain two files (and keep them in sync) and introduces the possibility of pushing an image to production that hasn't been tested. This is a concern, because perhaps an underlying system dependency update might cause a particular image to be broken. 

Perhaps we may wish to adapt the pattern here: https://docs.docker.com/build/ci/github-actions/test-before-push/

Then on the "push" action you add a selector such as `if: github.event.pull_request.merged ` so the push only happens during a merge to main.

.github/workflows/docker-hub-push.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Checkout selected files
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            build/docker/Dockerfile
+          sparse-checkout-cone-mode: false
+      - name: Download sdist artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: sdist
```
kenibrewer, line 36: I don't think this will work as intended because I don't think there will be an artifact attached to this workflow run. This would be another argument in favor of housing this action within the overall ci workflow because then we could get the artifact from a rerun of `build``.

https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories

.github/workflows/docker-hub-push.yml
```diff
@@ -0,0 +1,108 @@
...
+        id: env-vars
+        run: >-
+          VERSION=$(
+            find ./*.tar.gz |
+            sed 's/pycytominer-\(.*\)\.tar\.gz/\1/' |
+            tr '+' '_'
+          ) &&
+          echo "VERSION=$VERSION" >> "$GITHUB_OUTPUT" &&
+          echo "SDIST_FILENAME=$(find ./*.tar.gz)" >> "$GITHUB_OUTPUT" &&
+          echo "SDIST_EXTRACTED_NAME=\"pycytominer-$VERSION\"" >> "$GITHUB_OUTPUT" &&
```
kenibrewer, line 49: If we switch to making this part of the ci overall ci workflow, I would prefer we get these vars from `needs.upload.output` instead of using string parsing. String parsing is hard to ... parse ... 🥁 and is a common source of bugs in my view. 

If we get these variables from an appropriate output, this particular step could then be limited to generating a `CLEAN_VERSION` variable (without the `+`) and the `DATE_TAG`.

.github/workflows/docker-hub-push.yml
```diff
@@ -0,0 +1,108 @@
...
+      # unzips the sdist
+      - name: Extract sdist
+        run: |
+          tar -xzvf "${{ steps.env-vars.outputs.SDIST_FILENAME }}"
+      # note: roughly follows Docker documentation on GitHub Actions usage
+      # found on the following https://github.com/docker/build-push-action
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
kenibrewer, line 61: This appears to be duplicated code.

CONTRIBUTING.md
kenibrewer, line 1: Appreciate the clear documentation here!

kenibrewer (changes requested):
Things are looking very clean, but there's a use of download_artifact I think will break and I have an additional suggestion around merging the docker "build-test" action and the "build-push" action into a unified "build-test-push" action.

CONTRIBUTING.md
```diff
@@ -191,6 +191,16 @@ Creating a new release includes the following steps:
 11. The release will be automatically published to [PyPI](https://pypi.org/project/pycytominer/) via Github Actions.
 12. Manually create the release at [conda-forge](https://anaconda.org/conda-forge/pycytominer).
 
+#### Docker Hub Image Releases
```
gwaybio, line 194: Dropping by with a quick request to make sure to use active voice (instead of passive voice) when possible.

.github/workflows/docker-hub-push.yml
d33bs: Thanks, this is a great suggestion! I didn't realize we could use if-branching on trigger types. This makes a ton of sense and I'll update with related changes.

.github/workflows/docker-hub-push.yml
d33bs: Thanks for the feedback here, will plan on moving towards the "build-test-push" model you mention in other comments.

.github/workflows/docker-hub-push.yml
d33bs: Thanks, agreed on parsing 😆 ! I'll move towards limiting the creation of these vars as you described.

CONTRIBUTING.md
d33bs: Thanks!

.github/workflows/docker-hub-push.yml
d33bs: Thanks good point, will remove.

CONTRIBUTING.md
d33bs: Thanks! Will get this updated with active voice in mind.

d33bs:
Thanks @gwaybio and @kenibrewer for your reviews! I've worked towards addressing these and believe things are now ready for a re-review when there's a moment.

.github/workflows/integration-test.yml
```diff
@@ -193,3 +199,48 @@ jobs:
...
+        # only run this task if we have a pull request merged or the schedule triggers from main
+        if: github.event.pull_request.merged || ( github.event_name == 'schedule' && github.ref == 'refs/heads/main')
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      # Docker image push based on weekly schedule
+      - name: Schedule-based docker image push
+        # only run this task if we have a schedule trigger from main
+        if: github.event_name == 'schedule' &&  github.ref == 'refs/heads/main'
```
bethac07, line 219: Is there a case when this would not be `refs/heads/main`? I don't necessarily think so

.github/workflows/integration-test.yml
```diff
@@ -111,6 +116,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.get_version.outputs.version }}
+    if: github.event_name != 'schedule'
```
kenibrewer, line 119:
```suggestion
```
Won't this line mean there won't be a built package available to download during scheduled docker builds?

.github/workflows/integration-test.yml
kenibrewer: As currently written, we don't have any scheduled action runs that don't target refs/heads/main. However, I think keeping the explicit check for `refs/heads/main` check is good defensive posture to guard against a future scenario where someone working in a branch in a new github actions automation that inadvertently triggers a build and push out to dockerhub.

CONTRIBUTING.md
```diff
@@ -191,6 +191,16 @@ Creating a new release includes the following steps:
...
+
+We automate image pushes for `pycytominer` under the `cytomining` organization on [Docker Hub](https://hub.docker.com/) using GitHub Actions workflows.
+These pushes are defined within [.github/workflows/integration-test.yml](.github/workflows/integration-test.yml).
+
+- **Scheduled**: We create new Docker image releases on a weekly basis to incorporate the latest updates from external dependencies (such as OS updates, Python versions, etc.).
+  An image tag published this way may appear as `cytomining/pycytominer:1.1.0.post2.dev0_892dee2_240320`, where the dynamic version of `pycytominer` is referenced alongside a date in the format `YYMMDD`.
+- **Push (to `main`)**: We generate new Docker image releases on pushes or merges to the `main` branch.
+  An image tag published this way might appear as `cytomining/pycytominer:1.1.0.post2.dev0_892dee2`.
+
```
kenibrewer, line 203: Could we clarify that this scenario also applies for releases for clean version tags?

.github/workflows/integration-test.yml
d33bs: Thanks for the comments here! This was declarative for the reasons Ken mentioned; to avoid any ambiguity on where / when this should run. Thinking about this more, I'm going to add an additional constraint to avoid potential scheduled runs from a forked `main`.

.github/workflows/integration-test.yml
d33bs: Now that you mention this, I'm not certain what would occur. My intent was to use pre-built packages which were available during the latest push to `main`. This would largely be an effort towards reusing an unchanged package (if we have it) and reducing the need to consume resources (if we're able to).

CONTRIBUTING.md
d33bs: Thanks, good point! I'll add that in.

.github/workflows/integration-test.yml
d33bs: Reading on this more, you're correct, I'll get this updated.

.github/workflows/integration-test.yml
d33bs: I've added in greater specificity here to target only the `cytomining/pycytominer` repository and only `main` for scheduled pushes. Recognize this might be overdoing it but want to avoid any circumstances where we have challenges when it comes to multiple scheduled pushes coming from various locations (potentially overwriting one another without much observability).

d33bs:
Hi @bethac07 and @kenibrewer , I've added some changes based on your comments and have re-requested your review. Please don't hesitate to let me know if I should make any further changes to help ready this for pycytominer.

.github/workflows/integration-test.yml
```diff
@@ -193,3 +198,48 @@ jobs:
...
+      # Gather and prepare the dynamic version and other vars for use with work below.
+      # For example, 1.1.0.post6.dev0+bc093ef becomes 1.1.0.post6.dev0_bc093ef
+      - name: Gather dynamic version and set env var
+        id: env-vars
+        run: >-
+          echo "CLEAN_VERSION=$(echo ${{ env.sdist_filename }} | tr '+' '_' )" >> "$GITHUB_OUTPUT" &&
+          echo "DATE_TAG=$(date +'%y%m%d')" >> "$GITHUB_OUTPUT"
+      - name: Login to Docker Hub
+        # only run this task if we have a pull request merged or the schedule triggers from main
+        if: github.event.pull_request.merged || ( github.event_name == 'schedule' && github.ref == 'refs/heads/main')
```
kenibrewer, line 210: I think we want this if statement to avoid triggering off pull requests merged to non-main branches. 
```suggestion
        if: (github.event.pull_request.merged ||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
```

.github/workflows/integration-test.yml
```diff
@@ -193,3 +198,48 @@ jobs:
...
+        with:
+          context: "${{ env.sdist_extracted_name }}"
+          push: true
+          file: ../build/docker/Dockerfile
+          target: production
+          tags: cytomining/pycytominer:${{ steps.env-vars.outputs.CLEAN_VERSION }}_${{ steps.env-vars.outputs.DATE_TAG }}
+      # Docker image push based on GitHub pushes to main
+      - name: Push-based docker image push
+        # only run this task if we have a pull request which is merged
+        if: github.event.pull_request.merged
```
kenibrewer, line 229:
```suggestion
        if: github.event.pull_request.merged && github.ref == 'refs/heads/main' && github.repository == 'cytomining/pycytominer'
```

.github/workflows/integration-test.yml
```diff
@@ -193,3 +198,48 @@ jobs:
...
+        with:
+          context: "${{ env.sdist_extracted_name }}"
+          push: true
+          file: ../build/docker/Dockerfile
+          target: production
+          tags: cytomining/pycytominer:latest,cytomining/pycytominer:${{ steps.env-vars.outputs.CLEAN_VERSION }}
+      # update docker hub description using readme on merge
+      - name: Docker Hub
+        # only run this task if we have a pull request which is merged
+        if: github.event.pull_request.merged
```
kenibrewer, line 240:
```suggestion
        if: github.event.pull_request.merged && github.ref == 'refs/heads/main' && github.repository == 'cytomining/pycytominer'
```

kenibrewer (approved):
A few suggestions to make certain if checks a bit more restrictive. But otherwise this looks good to me and seems like implement's @bethac07 's tagging suggestions well. 

Excited to see this merged so we can hotfix whatever github actions corner case we inevitably missed despite triple checking every line!

bethac07 (approved):
I agree with Ken, I'm sure some hotfixing will be required just due to the laws of coding/the universe, but awesome job!

.github/workflows/integration-test.yml
d33bs: Thanks, great point! Adding this in.

d33bs:
Thank you so much @kenibrewer, @bethac07, and @gwaybio for your reviews and great suggestions throughout! Going to squash and merge this in.

PR #391, opened 2024-04-02
feat(dev): add improved makefile with additional helper commands

kenibrewer:
# Description

This adds an improved Makefile with several additional helper scripts to aid in development efforts. 

# Checklist


Makefile
```diff
@@ -1,25 +1,44 @@
...
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+.PHONY: docker-build
+test_docker_build: ## Build the docker image and run the tests
+	@echo "🐳 Building docker image and running tests"
+	@docker build -f build/docker/Dockerfile -t pycytominer:latest .
+	@docker run pycytominer:latest poetry run pytest
+
+.PHONY: help
+help:
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
```
d33bs, line 42: Consider quoting `$(MAKEFILE_LIST)` to prevent word splitting.
```suggestion
	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' "$(MAKEFILE_LIST)" | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
```

d33bs (approved):
Nice work @kenibrewer ! I left a minor comment about quoting. To help check syntax I used [`checkmake`](https://github.com/mrtazz/checkmake) which might be good to include as a linting step for `Makefile` modifications (it complained about missing phony targets for `all` and `clean` but I don't think these are required). I also used [`shellcheck`](https://github.com/koalaman/shellcheck) as a double check, which I feel is a great addition where any console scripting is used.

The previous content within the `Makefile` made it look similar to a Sphinx project default. As part of this, I think the `make.bat` file might also need an update (or possibly a removal). I believe the `.bat` is there to ensure Windows compatibility on systems which may not have `make` available. Double checking through Read the Docs, it didn't look like either the `Makefile` or `make.bat` are used. How would you feel about removing `make.bat` as part of this change?

As a future-facing thought: should we consider alternative devtooling task managers like [Poe the Poet](https://poethepoet.natn.io/index.html) or others? I've had good luck with it lately in other projects and only bring it up if it serves as inspiration (no strong feelings either way). A nice side-effect of using these is that we may have fewer files / formats to maintain within the project.

Makefile
kenibrewer: Accepted.

PR #395, opened 2024-04-02
fix(ci): docker image push fixes

d33bs:
# Description

This PR fixes a Docker image push trigger which didn't occur after merging in #377 . I believe [this GH Actions run](https://github.com/cytomining/pycytominer/actions/runs/8529106924) should have seen a push to Docker hub, but the if-conditions failed to trigger.

It also fixes incorrect Dockerfile path and tag naming witnessed after a [scheduled job failure](https://github.com/cytomining/pycytominer/actions/runs/8534627467/job/23379422488).

Thanks for any feedback you may have!

References #214 

# Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -201,13 +201,14 @@ jobs:
       # Gather and prepare the dynamic version and other vars for use with work below.
       # For example, 1.1.0.post6.dev0+bc093ef becomes 1.1.0.post6.dev0_bc093ef
       - name: Gather dynamic version and set env var
+        if: (github.event_name == 'push'||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
         id: env-vars
         run: >-
-          echo "CLEAN_VERSION=$(echo ${{ env.sdist_filename }} | tr '+' '_' )" >> "$GITHUB_OUTPUT" &&
+          echo "CLEAN_VERSION=$(echo ${{ env.sdist_extracted_name }} | tr '+' '_' )" >> "$GITHUB_OUTPUT" &&
           echo "DATE_TAG=$(date +'%y%m%d')" >> "$GITHUB_OUTPUT"
       - name: Login to Docker Hub
         # only run this task if we have a pull request merged or the schedule triggers from main
-        if: (github.event.pull_request.merged ||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
+        if: (github.event_name == 'push'||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USER }}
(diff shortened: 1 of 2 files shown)


kenibrewer (dismissed):
Sorry for jinxing this. 😅

d33bs:
Absolutely no worries @kenibrewer ! Thanks for the earlier review. After seeing the [scheduled job failure](https://github.com/cytomining/pycytominer/actions/runs/8534627467/job/23379422488) I decided to apply further fixes for the image tags, Dockerfile location, and removing tests from the build source (I believe the earlier copy for testing would persist into the production builds). I've re-requested a review as the changes are more broad than they initially were.

.github/workflows/integration-test.yml
```diff
@@ -198,16 +198,22 @@ jobs:
       - name: Run tests through docker image
         run: |
           docker run pycytominer:testing pytest
+      # remove tests for production builds below
+      - name: Remove tests from build source
+        if: (github.event_name == 'push'||  github.event_name == 'schedule') && github.ref == 'refs/heads/main'
+        run: |
+          rm -r ./"${{ env.sdist_extracted_name }}/tests"
```
kenibrewer, line 205: This is something that should be done in the Dockerfile rather than in Github Actions. 

Replace [this copy line](https://github.com/cytomining/pycytominer/blob/3286270321f0baac232556f6f16f61870596cb88/build/docker/Dockerfile#L28) with something like this:

```Dockerfile
COPY README.md pyproject.toml poetry.lock ./
COPY pycytominer ./pycytominer
```
The README.md is required because it is referenced in pyproject.toml and otherwise poetry won't install.

And then [here](https://github.com/cytomining/pycytominer/blob/3286270321f0baac232556f6f16f61870596cb88/build/docker/Dockerfile#L35) add:

```Dockerfile
COPY tests ./tests
```

kenibrewer (changes requested):
Thanks for re-requesting review here. There's one change we should make to the Dockerfile if we want a slimmed down production build.

kenibrewer:
Here's another possible change at your discretion. This would make it a bit easier with the context management.

1. Move `./build/docker/Dockerfile` to `./Dockerfile`
2. Add the `tests` directory as an include for the sdist build artifact but not wheel ([instructions](https://python-poetry.org/docs/pyproject/#include-and-exclude))

I think the end result would be that the commands `docker build . --target testing` would work both in a cloned repo root or in an extracted sdist.

d33bs:
Thanks @kenibrewer ! I applied the changes you mentioned but held off for now on the Dockerfile / `pyproject.toml` changes as I wasn't as certain about impacts or implementation here. I'll create a new issue as these feel like great future additions.

kenibrewer (approved):
Looks good! Thanks for capturing the new issue

d33bs:
Thanks @kenibrewer ! Merging this in.

PR #430, opened 2024-09-11
Improve Documentation for Non-CellProfiler Datasets in Pycytominer

axiomcura:
# Description

Thank you for your contribution to pycytominer!
Please _succinctly_ summarize your proposed change.
What motivated you to make this change?

This PR updates the README with instructions on how to handle morphological features extracted from software other than CellProfiler. These changes were conducted due to a reviewer’s comment/suggestion.

>"When I used the function pycytominer.normalize() on a dataset originating from my own work, I encountered an error message saying
“No CP features found. Are you sure this dataframe is from CellProfiler?”. 
I could easily fix the problem because I know the properties of a CellProfiler dataset, and the error message was hinting at some “Metadata_” missing. 
The fix was very easy, I just needed to add the prefix “Metadata_” to my metadata columns. My dataset comes from another commonly used proprietary software called Harmony. I understand that the authors are in favor of using open-source resources, but I think that the pycytominer tool could be useful in a much larger context, and take as input different types of datasets. In the readme file, that I followed to perform the normalization, I couldn’t find anywhere that the dataset has to come from CellProfiler. The author should state clearly, that only cellprofiler datasets can be used. Instead, currently, the sentence in the readme files of the pycytominer github repository says “Image data flow from a microscope to cell segmentation and feature extraction tools (e.g. CellProfiler or DeepProfiler)”. And in the code chunck the sentence says “Each function takes as input a pandas DataFrame or file path”, but in reality, it has to be a specifically formatted pandas dataframe (at least regarding the column names). I think the authors should fix this misunderstanding in the documentation. Alternatively, the better option would be to support the use of different datasets, and explain how the column names should be formatted to be easily processed by the different functions. I can’t exclude that this is described somewhere in the tutorials, but I didn’t encounter it when following the instructions on the readmefile. I would also make the error message more self-explanatory because for now, it implies that I know what CellProfiler is and how its output is formatted. If I don’t have this knowledge, as a user, I might give up on using this tool. This would be unfortunate, as I believe that this tool is a fundamental step in image-based profiling. Considering how easy was to fix this issue for my own dataset, which allowed me to immediately see the outputs of your tool, I recommend reconsidering how to best document the formatting steps necessary to be able to use the different functions of the pycytominer too"

Please also link to any relevant issues that your code is associated with.

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
[...]

PR #439, opened 2024-09-15
ci(mypy): add mypy check and adjust code for types

d33bs:
# Description

This PR adds a check through GitHub Actions for mypy in order to make sure Python types align with certain standards within pycytominer. I've also made some adjustments to the code in order to pass mypy checks.

Closes #437 

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [x] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

----
📚 Documentation preview 📚: https://pycytominer--439.org.readthedocs.build/en/439/

PR #396, opened 2024-04-03
docs: add pydocsstyle linting and switch to mkdocs

kenibrewer:
# Description

I was working on #382 and couldn't figure out how to resolve some sphinx build errors with opaque error messages. I decided to take this opportunity to switch our docs engine to mkdocs instead. 

- Enables pydocstrings linting checks
- Various documentation strings changes to conform with linting checks.
- Switch from Sphinx to Mkdocs
- Added docs build ci check

## What is the nature of your change?

- [x] Bug fix (fixes an issue).
- [x] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [x] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

----
📚 Documentation preview 📚: https://pycytominer--396.org.readthedocs.build/en/396/

PR #448, opened 2024-09-22
Update error message and docs for `features` argument to clarify CellProfiler default expectations and how to handle non-CellProfiler data

axiomcura:
# Description

## Summary

This PR addresses one of the reviewers’ comments:

---
" When I used the function pycytominer.normalize() on a dataset originating from my own work, I encountered an error message saying

“No CP features found. Are you sure this dataframe is from CellProfiler?”. 

I could easily fix the problem because I know the properties of a CellProfiler dataset, and the error message was hinting at some “Metadata_” missing. 

The fix was very easy, I just needed to add the prefix “Metadata_” to my metadata columns. My dataset comes from another commonly used proprietary software called Harmony. I understand that the authors are in favor of using open-source resources, but I think that the pycytominer tool could be useful in a much larger context, and take as input different types of datasets. In the readme file, that I followed to perform the normalization, I couldn’t find anywhere that the dataset has to come from CellProfiler. ..."

---

I replaced the ‘assert’ statement with an exception, as it’s considered bad practice to use assertions during runtime. Assertions can be disabled with the `-o` parameter, which means they aren’t reliable for handling errors in production code.
Here are some useful links:
- https://stackoverflow.com/questions/42837054/are-assert-methods-acceptable-in-production
- https://docs.astral.sh/ruff/rules/assert/
- https://stackoverflow.com/questions/1273211/disable-assertions-in-python

Additionally, I updated the error message to clarify that pycytominer infers CellProfiler features by default and provided guidance on how users can manually specify their feature space when working with non-CellProfiler data.

This code is also an addition to #430 

# Checklist


gwaybio:
@axiomcura - integration tests failed, please fix. Looks like the tests expect a specific error message which also needs updating

axiomcura:
@gwaybio All tests have passed! Please review when you have a moment to confirm if it's safe to merge.

gwaybio:
Thanks Erik! I see that the tests are passing now.

Re-reading the reviewer comment, it seemed that they have issues with the metadata parameter.

I'm thinking that in addition to clarifying the `features` exception, we should also do something to the `metadata` parameter here: https://github.com/cytomining/pycytominer/blob/main/pycytominer/cyto_utils/features.py#L113C1-L116C19

Perhaps we can do two more things in this PR to update error message in the feature inference.

1. Add an exception (and corresponding tests) for if `metadata_features > 0`
2. Add a description to the metadata parameter docstring specifying exactly what `metadata="infer"` actually expects/does

I'll also tag @d33bs - Dave, after Erik addresses these two points, perhaps you can take a final look to see if both Erik and I overlooked something in making this change. Thanks!

axiomcura:
@gwaybio - This is certainly doable, but I have a few clarifying questions. Could you please explain why the exception condition needs to be `metadata_features > 0`?

From my understanding, Pycytominer’s main functions default to `metadata_features="infer"`, which always expects metadata features from CellProfiler. As a result, adding an exception for `metadata_features > 0` in `infer_cp_features()` seems to introduce a contradictory logic. My guess would have been `metadata_features == 0`. 

I might be misunderstanding something where the `metadata_features > 0` exception is implemented. I just want to make sure if I am thinking of this correctly before I implement! Thanks!

gwaybio:
Great catch Erik! We should **not** check metadata features in the way I suggested above. We ask users to provide metadata features within each function (e.g., `normalize()`) call. 

Perhaps the only change should be whereever we see `meta_features="infer"` in the docstring (e.g., here: https://github.com/cytomining/pycytominer/blob/main/pycytominer/normalize.py#L41)

We add some specification to clarify that the cases in which `meta_features="infer"` works as expected is _not only_ when metadata features are prefixed by `Metadata_` but that this occurs when using CellProfiler-derived input.

I think this is a relatively minor and straightforward fix.

1. Find wherever `meta_features="infer"` is used.
2. Modify the docstring slightly to indicate CellProfiler-derived default

axiomcura:
@gwaybio - I have updated the docstrings.

kenibrewer:
Ahh shucks... there's going to be a lot of merge conflicts between this branch and #396. I was going to finish that one off this weekend. Not sure which order we want to merge / resolve merge conflicts in.

gwaybio:
Bummer! My sense is:

1. Merge #448 (clarify CellProfiler defaults)
2. Merge #439 (mypy)
3. Merge #396 (mkdocs)

We'll likely do 1 and 2 before the weekend FWIW

kenibrewer:
Sounds good! Let's do that!

gwaybio:
@d33bs - ready for you to take a final look, thanks!

kenibrewer (approved):
Looks great!

This would have been very helpful for my first attempt at using pycytominer too!

tests/test_operations/test_correlation_threshold.py
```diff
@@ -84,7 +84,12 @@ def test_correlation_threshold_featureinfer():
             method="pearson",
         )
 
-    assert "No CP features found." in str(nocp.value)
+        expected_message = (
+            "No features or metadata found. Pycytominer expects CellProfiler column names by default. "
+            "If you're using non-CellProfiler data, please do not 'infer' features. "
+            "Instead, check if the function has a `features` or `meta_features` parameter, and input column names manually."
+        )
+        assert expected_message in str(nocp.value)
```
d33bs, line 92: Similar comment to earlier.

tests/test_operations/test_get_na_columns.py
```diff
@@ -67,9 +67,14 @@ def test_get_na_columns_sample():
...
             population_df=data_df, samples="all", features="infer", cutoff=0.1
         )
 
-    assert "No CP features found." in str(nocp.value)
+        expected_message = (
+            "No features or metadata found. Pycytominer expects CellProfiler column names by default. "
+            "If you're using non-CellProfiler data, please do not 'infer' features. "
+            "Instead, check if the function has a `features` or `meta_features` parameter, and input column names manually."
+        )
+        assert expected_message in str(nocp.value)
```
d33bs, line 80: Similar comment to earlier.

tests/test_operations/test_variance_threshold.py
```diff
@@ -102,12 +102,17 @@ def test_variance_threshold():
...
             population_df=data_unique_test_df, features="infer", unique_cut=unique_cut
         )
 
-    assert "No CP features found." in str(nocp.value)
+        expected_message = (
+            "No features or metadata found. Pycytominer expects CellProfiler column names by default. "
+            "If you're using non-CellProfiler data, please do not 'infer' features. "
+            "Instead, check if the function has a `features` or `meta_features` parameter, and input column names manually."
+        )
+        assert expected_message in str(nocp.value)
```
d33bs, line 115: Similar comment to earlier.

d33bs (approved):
Nice job, I left a few minor comments, otherwise LGTM!

axiomcura:
@d33bs @gwaybio I’ve finished applying the requested changes. I’m currently waiting for the tests to pass, and once they do, it’s ready for merging. ( I am unable to merge)

gwaybio:
merged!

PR #495, opened 2025-01-21
Fix Bug in `sample` Parameter in `feature_select()`: Prevent Unintended Row Removal and Dataset Modification

### LINKED ISSUE #494: Bug: `samples` parameter in `feature_select` function reduces the rows in output
### Example code with output

Example code:
```python
    # Only for Plate 6, we want to normalize to iNFixion institution and Null and WT cells 
    # to keep consistent with how the other plates are normalized (same cell line)
    if plate == "Plate_6":
        samples = "Metadata_Institution == 'iNFixion' and (Metadata_genotype == 'Null' or Metadata_genotype == 'WT')"

    print(f"Performing normalization for {plate} using samples parameter: {samples}")

    # Step 2: Normalization
    normalized_df = normalize(
        profiles=output_annotated_file,
        method="standardize",
        samples=samples,
    )
    print("Normalized dataframe shape", normalized_df.shape)

    output(
        df=normalized_df,
        output_filename=output_normalized_file,
        output_type="parquet",
    )

    # Step 3: Feature selection
    feature_select_df = feature_select(
        output_normalized_file,
        operation=feature_select_ops,
        na_cutoff=0,
        samples=samples,
    )

    print("Feature selected dataframe shape", feature_select_df.shape)

    output(
        df=feature_select_df,
        output_filename=output_feature_select_file,
        output_type="parquet",
    )
```
Output:
```
Normalized dataframe shape (7383, 2319)
Feature selected dataframe shape (1674, 1148)
```

### Issue description

When using the `samples` parameter in the `feature_select` function, with the help of @gwaybio, we noticed that the number of rows reduced significantly. This did not happen if the `samples` parameter is removed, meaning it is defaulted to "all". 

Through talking with @axiomcura, it was determined that the issue was related to this part of the code:

```python
if samples != "all":
    population_df.query(samples, inplace=True)
```

When using query in this way, it removes all the other samples/rows in the dataframe, which means the output will be incorrect. 

NOTE: This parameter is also seen in the `normalize` function but it does not have this issue.

### Expected behavior

The expected behavior would be that the `feature_select` function will return a data frame with the same number of rows as the input when the `samples` parameter is used but will only determine which columns to keep based on only the samples specified.

### Additional information

- Pycytominer v1.1.0
- Linux operating system; Pop OS!
- Running in a jupyter notebook

axiomcura:
# Description

This pull request closes #494, which describes when using the `sample` parameter unexpectedly removes rows and modifies the original dataset. The changes aim to resolve this issue, ensuring that the `sample` parameter behaves as intended without altering the original data and removing data.

I have also created new tests to verify that the number of rows remains unchanged when applying the `samples` parameter, whether through individual operations and a combination of multiple operations.

# Checklist

- [ ] I have deleted all non-relevant text in this pull request template.

### DIFF
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -175,7 +175,7 @@ def drop_outlier_features(
 
     # Subset dataframe
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
 
     if features == "infer":
         features = infer_cp_features(population_df)
--- a/pycytominer/cyto_utils/features.py
+++ b/pycytominer/cyto_utils/features.py
@@ -173,19 +173,29 @@ def drop_outlier_features(
         Features greater than the threshold.
     """
 
-    # Subset dataframe
+    # Subset the DataFrame if specific samples are specified
+    # If "all", use the entire DataFrame without subsetting
     if samples != "all":
-        population_df = population_df.query(samples)
+        # Using pandas query to filter rows based on the conditions provided in the
+        # samples parameter
+        population_df = population_df.query(expr=samples)
 
+    # Infer  CellProfiler features if 'features' is set to 'infer'
     if features == "infer":
+        # Infer CellProfiler features
         features = infer_cp_features(population_df)
+        # Subset the DataFrame to only include inferred CellProfiler features
         population_df = population_df.loc[:, features]
     else:
+        # Subset the DataFrame to only include the features of interest
+        # this would be more tailored to non-CellProfiler features
         population_df = population_df.loc[:, features]
 
+    # Get the max and min values for each feature
     max_feature_values = population_df.max().abs()
     min_feature_values = population_df.min().abs()
 
+    # Identify features with max or min values greater than the outlier cutoff
     outlier_features = max_feature_values[
         (max_feature_values > outlier_cutoff) | (min_feature_values > outlier_cutoff)
     ].index.tolist()
--- a/pycytominer/operations/correlation_threshold.py
+++ b/pycytominer/operations/correlation_threshold.py
@@ -48,7 +48,7 @@ def correlation_threshold(
 
     # Subset dataframe and calculate correlation matrix across subset features
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
 
     if features == "infer":
         features = infer_cp_features(population_df)
--- a/pycytominer/operations/correlation_threshold.py
+++ b/pycytominer/operations/correlation_threshold.py
@@ -40,19 +40,26 @@ def correlation_threshold(
          List of features to exclude from the population_df.
     """
 
-    # Check that the input method is supported
+    # Checking if the provided correlation method is supported
     method = check_correlation_method(method)
 
+    # Checking if the threshold is between 0 and 1
     if not 0 <= threshold <= 1:
         raise ValueError("threshold variable must be between (0 and 1)")
 
     # Subset dataframe and calculate correlation matrix across subset features
+    # If samples is not 'all', then subset the dataframe
     if samples != "all":
-        population_df = population_df.query(samples)
+        # Using pandas query to filter rows based on the conditions provided in the
+        # samples parameter
+        population_df = population_df.query(expr=samples)
 
+    # Infer CellProfiler features if 'features' is set to 'infer'
     if features == "infer":
+        # Infer CellProfiler features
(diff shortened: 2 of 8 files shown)


gwaybio:
This looks functionally good to me @axiomcura - kudos and thanks! @d33bs could you give a quick, but close look to see if there are any technical things for Erik to consider prior to merging? One thing that I could potentially see is to create a new test (rather than append to an existing one), which would be more explicitly named to test the sample parameter (e.g., `test_feature_select_samples_paramater()`), but I'm not sure how hard or important this would be.

pycytominer/cyto_utils/features.py
```diff
@@ -175,7 +175,7 @@ def drop_outlier_features(
 
     # Subset dataframe
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
```
d33bs, line 178: Consider adding a comment here or replacing the one above this if-branch to better describe what's happening here and why. For example, the comment in `correlation_threshold` provides more information (but may also need to be updated).

pycytominer/operations/correlation_threshold.py
```diff
@@ -48,7 +48,7 @@ def correlation_threshold(
 
     # Subset dataframe and calculate correlation matrix across subset features
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
```
d33bs, line 51: Consider also adding or enhancing the code comment above this if-branch to provide a description of what this is doing.

pycytominer/operations/get_na_columns.py
```diff
@@ -33,7 +33,7 @@ def get_na_columns(population_df, features="infer", samples="all", cutoff=0.05):
     """
 
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
```
d33bs, line 36: Similar comment here: consider adding more detail through a code comment which describes what's happening here and why.

pycytominer/operations/noise_removal.py
```diff
@@ -43,7 +43,7 @@ def noise_removal(
     """
     # Subset dataframe
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
```
d33bs, line 46: Similar comment here: consider adding more detail through a code comment which describes what's happening here and why.

pycytominer/operations/variance_threshold.py
```diff
@@ -51,7 +51,7 @@ def variance_threshold(
 
     # Subset dataframe
     if samples != "all":
-        population_df.query(samples, inplace=True)
+        population_df = population_df.query(samples)
```
d33bs, line 54: For each one of these modified `population_df` lines consider adding in the named parameter within the call.
```suggestion
        population_df = population_df.query(expr=samples)
```

tests/test_feature_select.py
```diff
@@ -169,6 +169,58 @@ def test_feature_select_noise_removal():
     pd.testing.assert_frame_equal(result4b, expected_result4b)
     pd.testing.assert_frame_equal(result5b, expected_result5b)
 
+    # testing samples query leveraging metadata
```
d33bs, line 172: Comment applies to the modifications here within tests. I agree with @gwaybio and would suggest that this becomes a new test here to help limit the lines of code found within this single test.

pycytominer/operations/noise_removal.py
d33bs, line 46: Seeing how these functions which implement a `query` pattern made me wonder if there's an opportunity in future changes to abstract the functionality into a decorator or class which can help provide the capability without the repetition. I mention this as a thought towards continuous improvement and not an expectation within this PR for review.

d33bs (commented):
Nice job @axiomcura! I added some comments with this review and would be interested in getting your response. Please don't hesitate to let me know if you have any questions.

Generally I also wondered:
- Have we covered the change in functionality with new tests? I noticed there was a test for `feature_select` but perhaps not the other functions which were modified. Consider adding tests for these as well to ensure the sampling edge cases are covered (to avoid surprises later).

pycytominer/operations/get_na_columns.py
axiomcura: I have fixed this problem, but I've also noticed that there wasn't a check for the cutoff value. Therefore, I added a check and some tests to reflect this implementation.

d33bs (approved):
Thanks @axiomcura for making these changes! LGTM

PR #514, opened 2025-03-05
update doi for nature methods pub

gwaybio:
# Description

Our paper was published at Nature Methods: https://www.nature.com/articles/s41592-025-02611-8#citeas
I update DOI here to point to peer-reviewed paper.

Please also link to any relevant issues that your code is associated with.

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [x] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

PR #515, opened 2025-03-07
Remove citation info (keep just pointer) in readme

gwaybio:
# Description

I missed updating this in #514 - we shouldn't maintain two citation pointers


d33bs (approved):
LGTM!

PR #573, opened 2025-09-02
Enable anndata input

### LINKED ISSUE #566: FeatureRequest: add anndata as input format
We could add [anndata](https://github.com/scverse/anndata) as an input format for use within Pycytominer.

d33bs:
# Description

This PR enables anndata input through `load_profiles` and related functionality. 

Closes #566 

Thanks for any feedback!

# Checklist


d33bs:
Hello @flying-sheep , this PR follows our work in https://github.com/cytomining/CytoTable/pull/351 to add `anndata` integration with `pycytominer`. Could we ask for thoughts on the implementation? We'd greatly appreciate any guidance you 'd have for us - thank you!

pycytominer/cyto_utils/load.py
```diff
@@ -1,8 +1,13 @@
...
+Module for loading profiles from files or dataframes.
+"""
+
 import csv
 import gzip
 import pathlib
-from typing import Union
+from typing import Any, Optional, Union
 
+import anndata as ad
```
gwaybio: It is a big deal for pycytominer to now require `anndata` as a dependency. I think this is fine to do, but I want to confirm that you've considered the consequences (both good and bad).

A good is supporting interoperability, while a bad is the extra maintenance it may require (I do not the extent of additional maintenance, and it could be minimal). Another potential, maybe "neutral" consequence, is that an astute observer may see `anndata` as a dependency and then assume it is Pycytominer's intended default. It very may well become the default (only benchmarking and time will tell), but I don't think it is currently.

Any other consequences you can think?

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
...
 
-def load_profiles(profiles):
+def is_anndata(
+    path: Union[str, pathlib.Path, ad._core.anndata.AnnData],
+) -> tuple[bool, Optional[str]]:
+    """Return True if ``path`` contains an AnnData dataset (H5AD or Zarr).
+
+    This function prefers using the AnnData readers directly:
+    - in-memory AnnData objects are recognized directly.
+    - H5AD files are opened in backed mode to avoid loading data into memory.
```
gwaybio: what is "backed mode"?

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
...
+
+    Returns:
+        Tuple containing:
+        - Bool: True if the path appears to contain an AnnData dataset; otherwise
+        False.
+        - Str: If the path is an AnnData dataset, the type of store
+    """
+
+    # passthrough for anndata in-memory objects
+    if isinstance(path, ad._core.anndata.AnnData):
```
gwaybio: it would be great to get confirmation from the `anndata` folks if this covers 100% of anndata objects

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
     return dialect.delimiter
 
 
-def load_profiles(profiles):
+def is_anndata(
+    path: Union[str, pathlib.Path, ad._core.anndata.AnnData],
```
gwaybio:
```suggestion
    path_or_anndata_object: Union[str, pathlib.Path, ad._core.anndata.AnnData],
```

More accurate?

pycytominer/cyto_utils/load.py
gwaybio:
```suggestion
    # passthrough check if anndata in-memory object
    if isinstance(path_or_anndata_object, ad._core.anndata.AnnData):
```

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
...
+    """
+
+    # passthrough for anndata in-memory objects
+    if isinstance(path, ad._core.anndata.AnnData):
+        return True, "in-memory"
+
+    # check that the path exists
+    p = pathlib.Path(path)
+    if not p.exists():
+        return False, None
```
gwaybio: I think we should maintain the `strict=True` convention we have elsewhere

```python
try:
      file = pathlib.PurePath(file)
      # strict=true tests if path exists
      file = pathlib.Path(file).resolve(strict=True)
  except FileNotFoundError:
      raise FileNotFoundError("load_profiles() didn't find the path.")
  except TypeError:
      print("Detected a non-str or non-path object in the `file` parameter.")
      return False
```

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
     return dialect.delimiter
 
 
-def load_profiles(profiles):
+def is_anndata(
+    path: Union[str, pathlib.Path, ad._core.anndata.AnnData],
+) -> tuple[bool, Optional[str]]:
+    """Return True if ``path`` contains an AnnData dataset (H5AD or Zarr).
```
gwaybio:
```suggestion
    """
```

This is redundant with below and also not accurate (it returns a tuple, described well below)

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
...
+    # check that the path exists
+    p = pathlib.Path(path)
+    if not p.exists():
+        return False, None
+
+    # Zarr stores can be directories (common) or files (e.g., zipped stores).
+    # Try Zarr first for directories; for files, try H5AD then Zarr.
+    if p.is_dir():
+        try:
+            ad.read_zarr(p)
```
gwaybio: Am I correct that this function will read the entire file into memory only to return true if it is able to do so? Is there a way to avoid expending read energy?

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +77,66 @@ def infer_delim(file: str):
...
+    # File path: first try H5AD (backed)
+    try:
+        ad.read_h5ad(p, backed="r")
+        return True, "h5ad"
+    except Exception:
+        return False, None
+
+
+def load_profiles(
+    profiles: Union[str, pathlib.Path, pd.DataFrame, ad._core.anndata.AnnData],
```
gwaybio: 🎉

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +148,40 @@ def load_profiles(profiles):
     FileNotFoundError
         Raised if the provided profile does not exists
     """
-    if not isinstance(profiles, pd.DataFrame):
-        # Check if path exists and load depending on file type
-        if is_path_a_parquet_file(profiles):
-            return pd.read_parquet(profiles, engine="pyarrow")
 
+    # If already a dataframe, return it
+    if isinstance(profiles, pd.DataFrame):
```
gwaybio, line 101: we prob should also just return the annData object if it's this instead?

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +148,40 @@ def load_profiles(profiles):
...
+    anndata, anndata_type = is_anndata(profiles)
+    if anndata:
+        if anndata_type == "h5ad":
+            adata = ad.read_h5ad(profiles, backed="r")
+        elif anndata_type == "zarr":
+            adata = ad.read_zarr(profiles)
+        elif anndata_type == "in-memory":
+            adata = profiles
+        else:
+            raise ValueError("Unrecognized AnnData type")
```
gwaybio: would we ever reach this line?

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +148,40 @@ def load_profiles(profiles):
...
+
+        # Convert to dataframe
+        if adata.isbacked:
+            # if we're backed by a file, just read it directly
```
gwaybio: this is a curiosity question, but could pycytominer work on backed data? WOuld this have speed up benefits? (probably worth discussing in the context of a `narwhals` upgrade in 2.0 as well)

tests/test_cyto_utils/test_load.py
```diff
@@ -254,5 +283,45 @@ def test_load_profiles_file_path_input():
 
     # Testing non-existing file paths should result in expected behavior
     data_file_not_exist: pathlib.Path = pathlib.Path(tmpdir, "file_not_exist.csv")
-    with pytest.raises(FileNotFoundError, match="No such file or directory"):
+    with pytest.raises(FileNotFoundError, match="didn't find the path"):
         load_profiles(data_file_not_exist)
+
+
+def test_is_anndata(tmp_path: pathlib.Path) -> None:
```
gwaybio: This is an incredibly elegant test, kudos! 🎉

gwaybio (changes requested):
An exciting milestone @d33bs ! Let's discuss this a bit more

pycytominer/cyto_utils/load.py
```diff
@@ -32,19 +37,22 @@ def is_path_a_parquet_file(file: Union[str, pathlib.PurePath]) -> bool:
...
-    file = pathlib.PurePath(file)
     try:
+        file = pathlib.PurePath(file)
         # strict=true tests if path exists
         file = pathlib.Path(file).resolve(strict=True)
-    except FileNotFoundError as e:
-        print("load_profiles() didn't find the path.", e, sep="\n")
+    except FileNotFoundError:
+        raise FileNotFoundError("load_profiles() didn't find the path.")
+    except TypeError:
```
flying-sheep, line 44: if you really want to catch this, you should remove `TypeError` from the `Raises` section of the docstring and also document that passing some random non-path-like object in will make this return `False`.

I’d personally make that configurable with `*, strict: bool = True` (which when set to `False` would turn on the more lenient behavior of accepting anything)

pycytominer/cyto_utils/load.py
flying-sheep: it shouldn’t return a tuple, just `Literal["h5ad", "zarr", "in-memory", None]`, where `None` means that it’s not an AnnData object.

conveniently, that would mean that `if is_anndata(...)` would work as expected (a tuple of two things is always truthy!)

pycytominer/cyto_utils/load.py
flying-sheep: file-backed: as described, parts of the AnnData object are not read into memory, but will lazily be read from disk whenever a slice is accessed. (e.g. only when doing `adata[:5].X`, we will read the first 5 observations from disk)

heads-up: we currently think about deprecating it and recommending [`anndata.experimental.read_lazy`](https://anndata.readthedocs.io/en/stable/generated/anndata.experimental.read_lazy.html) as the way forward once it’s stable.

pycytominer/cyto_utils/load.py
flying-sheep: yes, but under no circumstances should you access private submodules!


```suggestion
    if isinstance(path, ad.AnnData):
```

pycytominer/cyto_utils/load.py
flying-sheep:
```suggestion
    path_or_anndata_object: Union[str, pathlib.Path, ad.AnnData],
```

yes. also don’t access private submodules!

pycytominer/cyto_utils/load.py
flying-sheep: yeah!

```py
from importlib.metadata import version
from packaging.version import Version

zarr_store = p
if path.suffix == ".zip" and Version(version("zarr")) >= Version("3"):
    zarr_store = zarr.storage.ZipStore(path)
try:
    group = zarr.open_group(zarr_store, mode="r")
except ValueError:
else:
    if group.attrs.get("encoding-type") == "anndata":
        return "zarr"
```

A less manual alternative would be to rely on the aforementioned `ad.experimental.read_lazy` as an alternative that validates that it’s a zarr or hdf5 store, but doesn’t read in too much. Unfortunately
1. AFAIK there’s no way to find out if we just opened a HDF5 file or a zarr store!
2. it won’t implicitly open a ZipStore at the time of writing

@ilan-gold will be back from holiday on monday and can tell us if `anndata.read_lazy` will be the future place for this API.

```py
for fn in [
   lambda p: ad.read_lazy(p),  # doesn’t exist yet
   lambda p: ad.experimental.read_lazy(p),
   ad.read_zarr,
]:
    try:
        with warnings.catch_warnings():
            warnings.filterwarnings("ignore", category=FutureWarning)
            fn(p)
    except AttributeError:
        continue
    return ???
```

pycytominer/cyto_utils/load.py
flying-sheep: it’s good to code defensively, but it should be an `AssertionError`, as we define the possible outputs.

tests/test_cyto_utils/test_load.py
flying-sheep: I’d suggest doing `@pytest.mark.parametrize`, that way if a change breaks a subset of these test cases, you’ll see exactly which ones.

pycytominer/cyto_utils/load.py
ilan-gold: Unless I'm off base, my response to the above and https://github.com/cytomining/pycytominer/pull/573#discussion_r2319384516 would be as follows:

1. I do think we will recommend deprecating backed mode or at least advise people not to use it since it is hdf5 specific and confusing _unless you want to be able to write back to the store_ (and even there, meh)
2. Considering the above, if you think you will need truly random, but non-lazy, access to the underlying files, I would recommend manually managing your own sparse array in `X` or elsewhere via a [custom dispatched io function](https://anndata.readthedocs.io/en/stable/generated/anndata.experimental.read_dispatched.html) using https://anndata.readthedocs.io/en/stable/generated/anndata.io.sparse_dataset.html
3. The above `sparse_dataset` function can be unfriendly because it does not allow building "views" fully as @flying-sheep mentioned (i.e., accessing `X` causes data to be read into memory).  It also requires you to write your own io function.  So if you are not too worried about perfect-random performance (i.e., are ok with "good enough"), `read_lazy` uses dask and will give chunked-access to the file in a unified API (including `obs` and `var`).  The resulting types and objects will also be simpler

I need to write up a blog post on this topic, but feel free to reply with questions, comments, or clarifications about your use-case, happy to help

pycytominer/cyto_utils/load.py
d33bs: I concur - good thoughts. Introducing the format means an added layer of support expectation from `pycytominer` users which will have related maintainer time / priority costs (I see this as a different aspect from solely technical maintenance in that we may need to educate and guide people towards solutions).

I'm a 👍 from the standpoint that it would likely be beneficial and we can always evolve the package towards user needs (whichever direction). We could also make it an optional dependency to tuck it away a bit but I like the simplicity of it being there from the get-go with a `pip install pycytominer`. If the dependency maintenance becomes a larger challenge we could move to an optional install.

pycytominer/cyto_utils/load.py
d33bs: Thanks @flying-sheep and @ilan-gold for the comments here! I'll add some notes to the code comments in reference to this. For now I'll plan to keep using backed mode until `read_lazily` matures from experimental status.

pycytominer/cyto_utils/load.py
d33bs: Thanks, I'll get this updated!

pycytominer/cyto_utils/load.py
d33bs: Thanks - sounds good, I'll update here. Sidenote, we should probably address [`expanduser`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.expanduser) scenarios for `~` usage. 

E.g.
```python
import pathlib

print(pathlib.Path("~").resolve())
print(pathlib.Path("~").expanduser().resolve())
```

pycytominer/cyto_utils/load.py
d33bs: Thanks, good points!

pycytominer/cyto_utils/load.py
d33bs: Thanks! I like the approach you highlighted @flying-sheep, I'll move to that instead!

pycytominer/cyto_utils/load.py
d33bs: Thanks good points @gwaybio and @flying-sheep ! I'll see about making this a lazier operation.

pycytominer/cyto_utils/load.py
d33bs: We haven't yet switched to an `anndata` compatible interface for all internal functions within `pycytominer` so I feel we should leave this as-is for now.  That said, we should double check on the other capabilities enabled here (we may still need a Pandas dataframe as the core in-memory format we can work with/from for now).

pycytominer/cyto_utils/load.py
d33bs: Thanks, good points, I'll move to `AssertionError` for this just in case.

pycytominer/cyto_utils/load.py
d33bs: It could work with backed data / lazy approached. However, it will come with a great deal of technical complexity no matter which API we choose (we won't be able to assume the materialization of the data as we do with Pandas currently). For example, `sklearn` methods might not all be "lazy compatible" or allow for partial fits.

It might be that we could bridge from backed `anndata` to a lazy frame of some kind with `polars`. I'd recommend opting for an abstraction layer such as `narwhals` or similar for internal processing so we leave the door open for current and future code development.

tests/test_cyto_utils/test_load.py
d33bs: Thanks @gwaybio and @flying-sheep ! Good call on `parametrize`, I'll go with that in a further revision.

pycytominer/cyto_utils/load.py
d33bs: Opened #576 in reference to this.

pycytominer/cyto_utils/load.py
d33bs: Thinking about this more I wondered if and how much this might become a stronger constraint to installs where folks might not be using anndata in their work (we haven't required it upfront before this point so it could take some time for this to build within the community as a pattern). I opted to create an `anndata` extra for installation following existing conventions within the project and using a TypeVar to help simulate the type where needed for `load`.

d33bs:
Thank you so much @gwaybio for the feedback! I've changed this PR significantly and would like to ask for another review when you have a moment. 

Thanks as well @flying-sheep and @ilan-gold for your thoughtful review comments throughout! I've done my best to address these with existing changes and look forward to exploring further how some of the things you mentioned, such as lazy processing through anndata.

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +190,62 @@ def load_profiles(profiles):
...
+        if anndata_type == "h5ad":
+            adata = ad.read_h5ad(profiles, backed="r")
+        elif anndata_type == "zarr":
+            adata = ad.read_zarr(profiles)
+        elif anndata_type == "in-memory":
+            adata = profiles
         else:
-            delim = infer_delim(profiles)
-            return pd.read_csv(profiles, sep=delim)
+            raise AssertionError("Unrecognized AnnData type")
```
ilan-gold: It looks like `adata` is only created to use `obs` in which case I would highly recommend using https://anndata.readthedocs.io/en/stable/generated/anndata.io.read_elem.html.  Otherwise, you will load everything into memory except `X` in the `h5ad` case, or literally everything for zarr

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +90,95 @@ def infer_delim(file: str):
...
+
+    import anndata as ad
+    import zarr
+    from packaging.version import Version
+
+    # passthrough check if anndata in-memory object
+    if isinstance(path_or_anndata_object, ad.AnnData):
+        return "in-memory"
+
+    # Expand user tilde and environment variables
```
gwaybio: Perhaps relevant to @flying-sheep comment in #575 ? I agree that automatic path expansion is dangerous, and probably too hand-holdy

pycytominer/cyto_utils/load.py
```diff
@@ -69,13 +90,95 @@ def infer_delim(file: str):
...
+                return "zarr"
+
+        # if we run into any error while attempting a read for zarr
+        # return None
+        except Exception:
+            return None
+
+    # File path: first try H5AD (backed)
+    try:
+        ad.read_h5ad(path, backed="r")
```
gwaybio: Will this read the entire file? Is there a way to only partially read to save computation?

pycytominer/cyto_utils/load.py
gwaybio: sounds good!

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +190,62 @@ def load_profiles(profiles):
...
+    if isinstance(profiles, pd.DataFrame):
+        return profiles
+
+    # Check if path exists and load depending on file type
+    if isinstance(
+        profiles, (str, pathlib.Path, pathlib.PurePath)
+    ) and is_path_a_parquet_file(profiles):
+        return pd.read_parquet(profiles, engine="pyarrow")
+
+    # Check if path is an AnnData file or object
```
gwaybio, line 110: shouldn't we use the logic in `is_anndata()` to handle this?

gwaybio: oh, I see, we do, but first we need to check for anndata like object?

IMO, and maybe I'm wrong, but shouldn't this logic go in `is_anndata()` as well? This change proposes a lot of changes to `load_profiles()`, which are kind of temporary, given our stance on AnnData dependency, which I think these changes sidestep

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +190,62 @@ def load_profiles(profiles):
...
+        )
+    ) and (anndata_type := is_anndata(profiles)):
+        # attempt an import of anndata and raise an error if not installed
+        try:
+            import anndata as ad
+        except ImportError:
+            raise ImportError(
+                """Optional dependency `anndata` is not installed.
+                Please install the `collate` optional dependency group:
+                e.g. `pip install pycytominer[anndata]`
```
gwaybio: I'm not a fan of this - if someone wants to use anndata, then they also have to install cytominer-database, a deprecated package? IMO, adding anndata as a dependency is a better option than requiring also a cytominer-database install.

tests/test_cyto_utils/test_load.py
```diff
@@ -254,5 +288,46 @@ def test_load_profiles_file_path_input():
 
     # Testing non-existing file paths should result in expected behavior
     data_file_not_exist: pathlib.Path = pathlib.Path(tmpdir, "file_not_exist.csv")
-    with pytest.raises(FileNotFoundError, match="No such file or directory"):
+    with pytest.raises(FileNotFoundError, match="didn't find the path"):
         load_profiles(data_file_not_exist)
+
+
+@pytest.mark.parametrize(
```
gwaybio: 🎉

pycytominer/cyto_utils/load.py
gwaybio: I agree this represents a significant change for existing workflows, and I am not aware of future workflows that would adopt anndata usage within pycytominer. 

An additional, probably stronger, argument for not making anndata a default dependency is our current functions can't operate on the format - is this true? If so, then merging only partial anndata support is going to alienate early adopters. What would it take to make pycytominer fully anndata compliant? If it's trivial, then maybe we make this part of this PR. I suspect, however, that it will be a moderate lift. In this case, I'm not quite sure what to do. We could merge this, but not release; but I would only do this if we know the scope of fully supporting anndata and have a reasonable timeline (~1 month) to completion.

d33bs: Great thoughts - this implementation is a first and "shallow" step towards integration with the anndata format (solely focusing on input and not general processing). We use `load_profiles` to load data from anndata sources and harmonize the in-memory format to align with existing use inside Pycytominer as a Pandas DataFrame (for ex. similar to how we treat CSV or Parquet input). Nearly every capability in Pycytominer relies on Pandas DataFrame input and as a result it'd be non-trivial to implement anndata use more deeply.

Harking to our existing community and also to future possibilities in the vein of anndata would I feel involve expanding on #576 and is I feel outside the scope of this PR. This would help us take full advantage of what anndata and distributed data formats have to offer.

pycytominer/cyto_utils/load.py
d33bs: Thanks - I disagree and feel that it's a very typical convention to use `~` and `$ENV_VARS` to expand paths as needed within software. That said, I'm outnumbered by vote here and I can live without it so I'll revert!

pycytominer/cyto_utils/load.py
d33bs: This was a typo on my part, sorry! The `anndata` extra just requires `anndata` and `zarr` dependencies. I'll revise here.

pycytominer/cyto_utils/load.py
d33bs: Thanks - I'll check on this and follow up with changes.

pycytominer/cyto_utils/load.py
d33bs: Thanks for checking on this - using the `backed='r'` argument avoids reading the full file.

pycytominer/cyto_utils/load.py
gwaybio: I also wonder if we can anticipate a potentially deeper dependency integration and create a new file `cyto_utils/anndata_utils.py`, where we can abstract anndata away from `load.py` to keep cleaner

d33bs: Thanks - I ended up leaning into this suggestion and created a new module for anndata utilities. With this I also opted to add an anndata writer to keep with our input / output patterns from `load` and `output`.

pycytominer/cyto_utils/load.py
d33bs: Thanks @ilan-gold ! I've added changes which attempt to use `read_elem` for `hda5` and `zarr` options here.

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    An interface for objects that behave like AnnData objects
+    without loading the actual AnnData package.
+    """
+
+    X: Any
+    obs: Any
+    var: Any
+
+
+# create an anntada-like type variable
```
gwaybio:
```suggestion
# create a custom type variable for anndata-like objects
```

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+Type_AnnDataLike = TypeVar("Type_AnnDataLike", bound=AnnDataLike)
+
+
+def is_anndata(
+    path_or_anndata_object: Union[str, pathlib.Path, AnnDataLike],
+) -> Optional[str]:
+    """
+    Return anndata type as str if
```
gwaybio:
```suggestion
    Return AnnData category as str if
```

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    path_or_anndata_object contains an AnnData dataset
+    or object (H5AD, Zarr, or in-memory object).
```
gwaybio:
```suggestion
    path_or_anndata_object contains an AnnData dataset,
    AnnData object (H5AD, Zarr, or in-memory object) or
    path to an AnnData dataset.
```

pycytominer/cyto_utils/anndata_utils.py
gwaybio: to match capitalization and to avoid name collision with type

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    This function prefers using the AnnData readers directly:
+    - in-memory AnnData objects are recognized directly.
+    - H5AD files are opened in backed mode to avoid loading data into memory.
+    (note:  anndata.experimental.read_lazy is likely to be a better option
+    in the future once stable)
+    - Zarr stores (directories or files like ``.zarr`` or ``.zip``) are read
+      via :func:`anndata.read_zarr`.
+
+    The function is conservative: on any read error (or if AnnData is not
+    installed), it returns None.
```
gwaybio: I don't think this is currently true - it will throw an error if AnnData cannot be imported, I believe

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    # File path: first try H5AD (backed)
+    try:
+        ad.read_h5ad(path, backed="r")
+        return "h5ad"
+    except Exception:
+        return None
+
+
+def read_anndata(
+    profiles: Union[str, pathlib.Path, pd.DataFrame, AnnDataLike], anndata_type: str
```
gwaybio:
```suggestion
    profiles: Union[str, pathlib.Path, pd.DataFrame, AnnDataLike], anndata_category: str
```

If you update language in the `is_anndata()` function docstring, it's probably best to update here too. "Type" may be a bit confusing given the Type variable AnnDataLike - A user may not have the same risk of confusion as a developer, so I could see it either way.

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    import zarr
+
+    # conditional import for read_elem because the
+    # location changed between anndata versions
+    try:
+        # 0.12+
+        from anndata.io import read_elem
+    except Exception:
+        # 0.10.x
+        from anndata.experimental import read_elem
```
gwaybio, line 139: I'm wondering if we should be opinionated about anndata version support. It seems we would benefit from opinions in terms of maintenance.

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # conditional import for ZarrGroup because the
+    # location changed between zarr versions
+    # note: used for typing only
+    try:
+        from zarr.hierarchy import Group as ZarrGroup  # zarr < 3
+    except Exception:
+        from zarr import Group as ZarrGroup  # zarr >= 3
```
gwaybio: This is super super minor, but the pattern above is try higher version, exception, try lower version, but here is the opposite (try lower version, exception, try higher version)

pycytominer/cyto_utils/load.py
```diff
@@ -87,16 +96,44 @@ def load_profiles(profiles):
...
+                Please install the `anndata` optional dependency group:
+                e.g. `pip install pycytominer[anndata]`
+                """
+            )
+        if anndata_type := is_anndata(profiles):
+            return read_anndata(profiles, anndata_type)
+
+    # otherwise, assume its a csv/tsv file and infer the delimiter
+    delim = infer_delim(profiles)
+    # also expand user tilde and environment variables in order to load the file
```
gwaybio:
```suggestion
```

I think now true?

pycytominer/aggregate.py
```diff
@@ -20,7 +20,9 @@ def aggregate(
     features: Union[list[str], str] = "infer",
     operation: str = "median",
     output_file: Optional[str] = None,
-    output_type: Optional[str] = "csv",
+    output_type: Literal[
+        "csv", "parquet", "anndata_h5ad", "anndata_zarr", None
+    ] = "csv",
```
gwaybio, line 25: Would we also need to update all other core functions?

README.md
```diff
@@ -68,7 +68,7 @@ docker run -v $PWD:/opt --platform=linux/amd64 cytomining/pycytominer:pycytomine
 
 Pycytominer is primarily built on top of [pandas](https://pandas.pydata.org/docs/index.html), also using aspects of SQLAlchemy, sklearn, and pyarrow.
 
-Pycytominer currently supports [parquet](https://parquet.apache.org/) and compressed text file (e.g. `.csv.gz`) i/o.
+Pycytominer currently supports [parquet](https://parquet.apache.org/), compressed text (e.g. `.csv.gz`), and [`anndata`](https://github.com/scverse/anndata) (through the extra `pycytominer[anndata]` and limited to `h5ad` or `zarr`) input and output data.
```
gwaybio:
```suggestion
Pycytominer currently supports [parquet](https://parquet.apache.org/), compressed text (e.g. `.csv.gz`), and [`anndata`](https://github.com/scverse/anndata) (through the extra `pip install pycytominer[anndata]` and limited to `h5ad` or `zarr`) input and output data.
```

gwaybio (approved):
Looks great @d33bs ! I have only minor comments

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+            # account for zipped zarr stores if zarr >= 3.0.0
+            if path.suffix == ".zip" and Version(version("zarr")) >= Version("3"):
+                zarr_store = zarr.storage.ZipStore(str(path))
+
+            # try to open the zarr store
+            group = zarr.open_group(zarr_store, mode="r")
+
+            # check the group encoding-type attribute for anndata
+            if group.attrs.get("encoding-type") == "anndata":
+                return "zarr"
```
ilan-gold, line 87: I would do this same check for hdf5 instead of going through backed mode.  Same reason as before, you'll load all the data that isn't in `X`

pycytominer/cyto_utils/anndata_utils.py
ilan-gold, line 139:
```suggestion
        from anndata.io import read_elem
```

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        return df_out
+    else:
+        raise ValueError("Unrecognized AnnData type.")
+
+    # Normalize X to a dense ndarray if needed
+    if hasattr(X, "toarray"):  # works for scipy sparse and others
+        X = X.toarray()
+
+    # Convert to DataFrame using obs and var for indices and columns
+    X_df = pd.DataFrame(X, index=obs.index, columns=var.index)
```
ilan-gold, line 179: Have a look at https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sparse.from_spmatrix.html instead of dense-ifying, looks like a less expensive operation :)

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    # For any remaining object columns that aren't strings (e.g., mixed),
+    # replace pd.NA with None to avoid h5py issues.
+    for c in df_nonnumeric.columns:
+        if df_nonnumeric[c].dtype == object and not isinstance(
+            df_nonnumeric[c].dtype, pd.CategoricalDtype
+        ):
+            df_nonnumeric[c] = df_nonnumeric[c].where(~pd.isna(df_nonnumeric[c]), None)
+
+    # Build AnnData
+    adata = ad.AnnData(X=X)
```
ilan-gold: Hmm, is `X` possibly empty (looking at https://github.com/cytomining/pycytominer/pull/573/files#diff-48c567fdb1183312cdc3790d69c8b357140541538f05dd176ea9a3fed1847b7bR209-R214)? We support `None` as an entry for `X` actually!

pycytominer/cyto_utils/anndata_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    # Split numeric vs non-numeric
+    numeric_cols = df.select_dtypes(include="number").columns
+    nonnumeric_cols = df.columns.difference(numeric_cols)
+
+    # Build X explicitly (handle zero-variable case)
+    if len(numeric_cols):
+        X = df[numeric_cols].to_numpy(dtype=float)
+        var_names = numeric_cols.astype(str)
+    else:
+        X = np.empty((len(df), 0), dtype=float)
```
ilan-gold: The values in `np.empty` could in theory be anything, it's uninitialized memory.  Is that intentional?

pycytominer/aggregate.py
d33bs: I think `aggregate` is one of the only functions which include type hints at the moment (I made this change because `mypy` flagged it as an error due to mismatching types). I feel we should hold off on further changes for now but we should revisit this soon - it may matter more and more for performance or linting compatibility reasons. Another reason for holding off is that there are likely some type hint bugs we will need to navigate through for core functions. I've created an issue to follow up: https://github.com/cytomining/pycytominer/issues/578

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - great point!

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - yes I think that's true, I'll revise.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks, great point! I'll revise here.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks, I like this idea, let's move ahead with "category" as a path forward. I'll double check that this remains consistent.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks, good point, I'll revise to follow the same pattern here.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - this is a reference for older anndata versions which might be a reliance due to the version of Python someone is using. I'm going to leave this as-is for now.

d33bs: Thanks - I made this change from the assumption that we'd support `anndata` for the Python versions we support through `pycytominer`. I feel that it'd serve the audience to have this available if it's not too difficult to support (a simple try-except felt not too bad). If we find that it's more difficult to support I feel we could roll back. I'm going to leave as-is for now.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - I'll have a look!

pycytominer/cyto_utils/anndata_utils.py
d33bs: This may be removable once we drop Python 3.9 support.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks again for this comment. I did some work to check on common data we encounter through this project. I found that we typically work with dense (mostly non-zero) data and as a result `pd.DataFrame.sparse.from_spmatrix` would likely be more expensive to process than a `pd.DataFrame`. This is a reference to what I used to make a quick assessment here: https://gist.github.com/d33bs/326a0a2b77e361f577e42469b867cce9

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks for raising this! I found that I needed to make this change to accommodate how we created the anndata object below with the edge case of having no numeric columns. I found that if I created the anndata object with `X` and `obs` set it doesn't fail when using `None`. Is this a bug @ilan-gold? I might be misunderstanding too! See here for an example: https://gist.github.com/d33bs/061ee0833c97b500761a57d31d47a05d .

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - I moved to `None` as a result of these comments!

pycytominer/cyto_utils/anndata_utils.py
ilan-gold: No doubt, perhaps this was the wrong place for the comment.  When you do `X = X.toarray()` above, and then pass it here, you could instead just use the sparse data directly to generate the matrix, no?

pycytominer/cyto_utils/anndata_utils.py
ilan-gold: Hmm, no that seems somewhat expected if not an edge case.  But when you make the anndata object below, couldn't you do `adata = ad.AnnData(X=X, obs=df_nonnumeric, var=var)`?

My point was less about anndata here but rather that when you write this anndata object, you will write `np.empty` which could contain anything, not 0's or similar.

So I would think that `X=None` and `var=pd.DataFrame(index=pd.Index([], dtype=object))` would be set here in the `else` case, `obs=df_nonnumeric` after the below lines of code you have for handling, and then `adata = ad.AnnData(X=X, obs=df_nonnumeric, var=var)` which should sidestep this issue in hte gist of "case 1".  I'm not sure I see why in this case you want to go into "case 1" of the gist you attached intentionally

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks - I understand better now! I've made a change to help ensure we handle sparse data if present.

pycytominer/cyto_utils/anndata_utils.py
d33bs: Thanks @ilan-gold, understood! I think I added these lines because of needing to use `adata.obs_names` after construction (I couldn't figure out what parameter to use for these references). I've moved to use a more condensed `ad.AnnData(X=X, obs=df_nonnumeric, var=var)` as you suggested for now. Thanks again!

Issue #312, opened 2023-07-21
Reduce duplicate code with decorators


gwaybio:
@d33bs brings up a good point in #311, which we could pursue to reduce code redundancy and improve maintainability.

Seeing how the patterns with `pd.DataFrame.query` repeat in this and the other functions, consider using a utility or decorator function to decrease the amount of repeated code and documentation. A decorator might work nicely to enable both documentation and functionality for these and any other functions which may need these capabilities in the future. See the following example for how this could work:

```python
import pandas as pd


def enable_pd_df_query(func):
    """
    A decorator for enabling pd.DataFrame.query()
    """

    def wrapper(df: pd.DataFrame, samples: str = "all", *args, **kwargs):
        df = df.query(samples) if samples != "all" else df
        return func(df, samples, *args, **kwargs)

    return wrapper


@enable_pd_df_query
def some_function(df: pd.DataFrame, samples: str = "all"):
    """
    A function to demonstrate decorator functionality
    """

    return df


df = pd.DataFrame({"A": range(1, 6), "B": range(10, 0, -2), "C C": range(10, 5, -1)})
some_function(df=df, samples="A > 1")
```

_Originally posted by @d33bs in https://github.com/cytomining/pycytominer/pull/311#discussion_r1268376455_

gwaybio:
addressed partially in #600, but there are likely other opportunities. We don't need an open issue to track this

PR #600, opened 2025-11-28
refactor: add output decorator for repeated use

d33bs:
# Description

This PR adds a decorator for repeated use of output patterns found in Pycytominer. The goal is to centralize and standardize our approach to output from the core functions (and deduplicate code in the process).

References #312

# Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.15.0"
+    rev: "v0.15.1"
     hooks:
     -   id: ruff-check
         exclude: tutorials/nbconverted/


pycytominer/cyto_utils/util.py
```diff
@@ -176,6 +177,52 @@ def check_consensus_operation(operation: str) -> str:
...
+    func: Callable[..., Union[pd.DataFrame, str]],
+) -> Callable[..., Union[pd.DataFrame, str]]:
+    """Decorate a function to optionally write its output to disk.
+
+    The decorator intercepts common output-related keyword arguments
+    (``output_file``, ``output_type``, ``compression_options``, ``float_format``)
+    from the decorated function call. The wrapped function should return a
+    :class:`pandas.DataFrame`; when ``output_file`` is provided, the DataFrame is
+    written using :func:`pycytominer.cyto_utils.output` and the resulting path is
+    returned instead.
```
gwaybio:
```suggestion
    :class:`pandas.DataFrame` when ``output_file`` is provided; the DataFrame is
    written using :func:`pycytominer.cyto_utils.output` and the resulting path is
    returned instead.
```

pycytominer/cyto_utils/util.py
```diff
@@ -176,6 +177,52 @@ def check_consensus_operation(operation: str) -> str:
...
+    """
+
+    signature = inspect.signature(func)
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Union[pd.DataFrame, str]:
+        bound_arguments = signature.bind_partial(*args, **kwargs)
+        bound_arguments.apply_defaults()
```
gwaybio, line 201: Can you add a bit of context in a comment about what this is doing?

pycytominer/cyto_utils/util.py
```diff
@@ -176,6 +177,52 @@ def check_consensus_operation(operation: str) -> str:
...
+
+        result = func(**bound_arguments.arguments)
+
+        if output_file is None:
+            return result
+
+        if not isinstance(result, pd.DataFrame):
+            raise TypeError(
+                "Decorated function must return a pandas DataFrame when output_file is provided"
+            )
```
gwaybio: Maybe it's because I am not understanding the bind_partial call above, but it is unclear to me how result will ever be a dataframe

pycytominer/cyto_utils/util.py
```diff
@@ -176,6 +177,52 @@ def check_consensus_operation(operation: str) -> str:
     return operation
 
 
+def maybe_write_to_file(
```
gwaybio:
```suggestion
def write_to_file_if_user_specifies_output_details(
```

I think adding specificity here (describing the "maybe" directly) helps readability. Please feel free to modify my suggestion as well.

pycytominer/cyto_utils/util.py
d33bs: Nice, thanks!

pycytominer/cyto_utils/util.py
d33bs: Thanks, I'll add some docs to this to help clarify. This block wraps the `func` so it appears and acts like the original function as opposed to `wrapper`. We use the `bound_arguments` calls to bind `args` and `kwargs` appropriately to the functions parameters (partial here means we only bind to what we have instead of all possible args/kwargs). In the last line, we apply defaults to the bound_arguments based on the defaults of whatever we wrapped.

pycytominer/cyto_utils/util.py
d33bs: Love it, thanks!

pycytominer/cyto_utils/util.py
d33bs: Thanks for bringing this up, I think you're correct. I'll remove, this was I believe a vestige of an earlier revision.

d33bs:
Thanks @gwaybio ! I've addressed your comments and believe this is good for another review when ready.

PR #622, opened 2026-02-03
feat: add cli for core functions

### LINKED ISSUE #280: [Enhancement] Expose primary functions (annotate, feature_select, etc.) as CLI commands
For the purposes of integrating pycytominer into data pipelines like Nextflow it would be great if core functionality could be exposed via a CLI. The quickest way to implement this in my view would be via Google's [python fire](https://github.com/google/python-fire) which can quickly convert functions/modules into CLIs using just the argument names and docstrings. 

Suggested implementation:

- [ ] Create shared wrapper that can launch each of the core functions
- [ ] Implement logic for reading in population_df/profiles from file.
- [ ] Implement logic for reading in features from file

Questions: 

1. Do any input formats need to be supported besides `csv` `csv.gz` and `parquet`?

To avoid an excessively large PR, I suggest a minimal-viable-product style implementation to start. This could mean that perhaps we don't expose certain lesser-used or difficult-to-implement function arguments to start. Once something working has been integrated, then we can circle back and add them in a follow-up round.

d33bs:
# Description

This PR adds a CLI for pycytominer core functions to help make the project more convenient to use without access through additional Python code.

Closes #280

# Checklist

----
📚 Documentation preview 📚: https://pycytominer--622.org.readthedocs.build/en/622/

### DIFF
--- a/pycytominer/annotate.py
+++ b/pycytominer/annotate.py
@@ -13,10 +13,11 @@ from pycytominer.cyto_utils import (
     infer_cp_features,
     load_platemap,
     load_profiles,
-    output,
 )
+from pycytominer.cyto_utils.util import write_to_file_if_user_specifies_output_details
 
 
+@write_to_file_if_user_specifies_output_details
 def annotate(
     profiles: Union[str, pd.DataFrame],
     platemap: Union[str, pd.DataFrame],
@@ -147,14 +148,4 @@ def annotate(
     other_cols = annotated.drop(meta_cols, axis="columns").columns.tolist()
 
     annotated = annotated.loc[:, meta_cols + other_cols]
-
-    if output_file is not None:
-        return output(
-            df=annotated,
-            output_filename=output_file,
-            output_type=output_type,
-            compression_options=compression_options,
-            float_format=float_format,
-        )
-    else:
-        return annotated
+    return annotated
--- a/pycytominer/aggregate.py
+++ b/pycytominer/aggregate.py
@@ -7,13 +7,11 @@ from typing import Any, Literal, Optional, Union
 import numpy as np
 import pandas as pd
 
-from pycytominer.cyto_utils import (
-    check_aggregate_operation,
-    infer_cp_features,
-    output,
-)
+from pycytominer.cyto_utils import check_aggregate_operation, infer_cp_features
+from pycytominer.cyto_utils.util import write_to_file_if_user_specifies_output_details
 
 
+@write_to_file_if_user_specifies_output_details
 def aggregate(
     population_df: pd.DataFrame,
     strata: list[str] = ["Metadata_Plate", "Metadata_Well"],
@@ -130,13 +128,4 @@ def aggregate(
     ]:
         population_df = population_df.drop(columns=columns_to_drop, axis="columns")
 
-    if output_file is not None:
-        return output(
-            df=population_df,
-            output_filename=output_file,
-            output_type=output_type,
-            compression_options=compression_options,
-            float_format=float_format,
-        )
-    else:
-        return population_df
+    return population_df
--- a/pycytominer/consensus.py
+++ b/pycytominer/consensus.py
@@ -7,14 +7,11 @@ from typing import Any, Literal, Optional, Union, cast
 import pandas as pd
 
 from pycytominer.aggregate import aggregate
-from pycytominer.cyto_utils import (
-    check_consensus_operation,
-    load_profiles,
-    modz,
-    output,
-)
+from pycytominer.cyto_utils import check_consensus_operation, load_profiles, modz
+from pycytominer.cyto_utils.util import write_to_file_if_user_specifies_output_details
 
 
+@write_to_file_if_user_specifies_output_details
 def consensus(
     profiles: Union[str, pd.DataFrame],
     replicate_columns: list[str] = ["Metadata_Plate", "Metadata_Well"],
@@ -136,13 +133,4 @@ def consensus(
             ),
         )
 
-    if output_file is not None and isinstance(consensus_df, pd.DataFrame):
-        return output(
-            df=consensus_df,
-            output_filename=output_file,
-            output_type=output_type,
-            compression_options=compression_options,
-            float_format=float_format,
-        )
-    else:
-        return consensus_df
+    return consensus_df
--- a/pycytominer/cyto_utils/__init__.py
+++ b/pycytominer/cyto_utils/__init__.py
@@ -39,5 +39,6 @@ from .util import (
     get_default_compartments,
     get_pairwise_correlation,
     load_known_metadata_dictionary,
+    write_to_file_if_user_specifies_output_details,
 )
 from .write_gct import write_gct
--- a/pycytominer/cyto_utils/util.py
+++ b/pycytominer/cyto_utils/util.py
@@ -2,16 +2,17 @@
 Miscellaneous utility functions
 """
 
+import inspect
 import os
 import warnings
-from typing import Literal, Union, cast
+from functools import wraps
+from typing import Callable, Literal, Union, cast
 
 import numpy as np
 import pandas as pd
 
-from pycytominer.cyto_utils.features import (
-    convert_compartment_format_to_list,
-)
+from pycytominer.cyto_utils.features import convert_compartment_format_to_list
+from pycytominer.cyto_utils.output import output
 
 default_metadata_file = os.path.join(
     os.path.dirname(__file__), "..", "data", "metadata_feature_dictionary.txt"
@@ -174,6 +175,56 @@ def check_consensus_operation(operation: str) -> str:
             )
 
     return operation
+
+
+def write_to_file_if_user_specifies_output_details(
+    func: Callable[..., Union[pd.DataFrame, str]],
+) -> Callable[..., Union[pd.DataFrame, str]]:
+    """Decorate a function to optionally write its output to disk.
+
+    The decorator intercepts common output-related keyword arguments
+    (``output_file``, ``output_type``, ``compression_options``, ``float_format``)
+    from the decorated function call. The wrapped function should return a
+    :class:`pandas.DataFrame` when ``output_file`` is provided; the DataFrame is
+    written using :func:`pycytominer.cyto_utils.output` and the resulting path is
+    returned instead.
+    """
+
+    signature = inspect.signature(func)
+
+    # wraps the function to preserve docstring and function name
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Union[pd.DataFrame, str]:
+        # bind the passed arguments to the function's signature
+        bound_arguments = signature.bind_partial(*args, **kwargs)
+        # fill in default values for missing arguments
+        bound_arguments.apply_defaults()
+
+        # extract output-related arguments
+        output_file = bound_arguments.arguments.pop("output_file", None)
+        output_type = bound_arguments.arguments.pop("output_type", None)
+        compression_options = bound_arguments.arguments.pop("compression_options", None)
+        float_format = bound_arguments.arguments.pop("float_format", None)
+
+        # call the original function with the remaining arguments
+        result = func(**bound_arguments.arguments)
+
+        # if no output file is specified, return the DataFrame directly
+        if output_file is None:
+            return result
+
+        # write the DataFrame to file and return the file path
+        return output(
+            # note: we cast here because mypy cannot
+            # infer that result is a DataFrame (not str)
+            df=cast(pd.DataFrame, result),
+            output_filename=output_file,
+            output_type=output_type,
+            compression_options=compression_options,
+            float_format=float_format,
+        )
+
+    return wrapper
 
 
 def check_fields_of_view_format(
--- a/pycytominer/feature_select.py
+++ b/pycytominer/feature_select.py
@@ -11,8 +11,8 @@ from pycytominer.cyto_utils import (
     get_blocklist_features,
     infer_cp_features,
     load_profiles,
-    output,
 )
+from pycytominer.cyto_utils.util import write_to_file_if_user_specifies_output_details
 from pycytominer.operations import (
     correlation_threshold,
     get_na_columns,
@@ -21,6 +21,7 @@ from pycytominer.operations import (
 )
 
 
+@write_to_file_if_user_specifies_output_details
 def feature_select(
     profiles: Union[str, pd.DataFrame],
     features: Union[str, list[str]] = "infer",
@@ -201,13 +202,4 @@ def feature_select(
 
     selected_df = profiles.drop(excluded_features, axis="columns")
 
-    if output_file is not None:
-        return output(
-            df=selected_df,
-            output_filename=output_file,
-            output_type=output_type,
-            compression_options=compression_options,
-            float_format=float_format,
-        )
-    else:
-        return selected_df
+    return selected_df
--- a/pycytominer/normalize.py
+++ b/pycytominer/normalize.py
@@ -9,10 +9,11 @@ from sklearn.preprocessing import RobustScaler, StandardScaler
 
 from pycytominer.cyto_utils.features import infer_cp_features
 from pycytominer.cyto_utils.load import load_profiles
-from pycytominer.cyto_utils.output import output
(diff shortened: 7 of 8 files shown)


README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
 )
 ```
 
+### CLI usage (file-based workflows)
```
gwaybio:
```suggestion
### Command Line Interface (CLI) usage
```

I don't think we need to specify the workflow type

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+
+Pycytominer also provides a simple CLI for file-based workflows. These commands read
+profiles from disk and write outputs to disk.
```
gwaybio:
```suggestion
Pycytominer also provides a simple CLI for file-based workflows. 
The Pycytominer CLI calls single Pycytominer functions that read profiles from disk, perform a specific operation with provided arguments, and write outputs to disk without requiring a dedicated python file.
```

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+
+```bash
+# Aggregate profiles (parquet output)
+python -m pycytominer aggregate \
+  --profiles path/to/profiles.csv.gz \
+  --output_file path/to/profiles_aggregated.parquet \
+  --output_type parquet \
+  --strata Metadata_Plate,Metadata_Well \
+  --features Cells_AreaShape_Area,Cytoplasm_AreaShape_Area
```
gwaybio:
```suggestion
# Aggregate profiles (note the parquet output type)
pycytominer aggregate \
  --profiles path/to/profiles.csv.gz \
  --output_file path/to/profiles_aggregated.parquet \
  --output_type parquet \
  --strata Metadata_Plate,Metadata_Well \
  --features Cells_AreaShape_Area,Cytoplasm_AreaShape_Area
```

Is it possible to remove the `python -m` prompting? The suggested style above is much more conventional for bioinformatics

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+
+# Annotate profiles with a platemap (compressed CSV output)
+python -m pycytominer annotate \
+  --profiles path/to/profiles_aggregated.parquet \
+  --platemap path/to/platemap.csv \
+  --output_file path/to/profiles_augmented.csv.gz \
+  --join_on Metadata_well_position,Metadata_Well
```
gwaybio:
```suggestion
# Annotate profiles with platemap metadata
python -m pycytominer annotate \
  --profiles path/to/profiles_aggregated.parquet \
  --platemap path/to/platemap.csv \
  --output_type parquet \
  --output_file path/to/profiles_augmented.parquet \
  --join_on Metadata_well_position,Metadata_Well
```

It might not send the right message if we switch between file output types mid pipeline - users will expect equivalent functionality between the CLI and the python command

gwaybio: oh, and also, should we have to provide the output_type flag if the output file is suffixed with the output file type? (maybe beyond scope of this PR)

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+
+# Normalize profiles (parquet output)
```
gwaybio:
```suggestion
# Normalize profiles
```

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+python -m pycytominer normalize \
+  --profiles path/to/profiles_augmented.csv.gz \
+  --output_file path/to/profiles_normalized.parquet \
+  --output_type parquet \
+  --features Cells_AreaShape_Area,Cytoplasm_AreaShape_Area \
+  --meta_features Metadata_Plate,Metadata_Well \
+  --samples "Metadata_treatment == 'control'" \
+  --method standardize
+
+# Feature selection (compressed CSV output)
```
gwaybio:
```suggestion
# Feature selection
```

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+python -m pycytominer feature_select \
+  --profiles path/to/profiles_normalized.parquet \
+  --output_file path/to/profiles_feature_selected.csv.gz \
```
gwaybio:
```suggestion
  --output_file path/to/profiles_feature_selected.parquet \
  --output_type parquet
```

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
...
+  --features Cells_AreaShape_Area,Cytoplasm_AreaShape_Area \
+  --operation variance_threshold,correlation_threshold
+
+# Consensus profiling (parquet output)
```
gwaybio:
```suggestion
# Form consensus profiles
```

docs/install.rst
```diff
@@ -20,6 +20,55 @@ These images follow a tagging scheme that extends our release sematic versioning
    # pull the latest pycytominer image and run a module
    docker run --platform=linux/amd64 cytomining/pycytominer:latest python -m pycytominer.<modules go here>
 
+Command-line usage
```
gwaybio: is this block below autogenerated from the readme? If so, please re-render once you've integrated my feedback on the previous file

README.md
gwaybio: (if possible to remove the `python -m` prompt, please do in all CLI calls below)

pycytominer/annotate.py
```diff
@@ -29,8 +29,8 @@ def annotate(
     format_broad_cmap: bool = False,
     clean_cellprofiler: bool = True,
     external_metadata: Optional[Union[str, pd.DataFrame]] = None,
-    external_join_left: Optional[bool] = None,
-    external_join_right: Optional[bool] = None,
+    external_join_left: Optional[str] = None,
```
gwaybio, line 33: nice catch!

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
+"""Command-line interface for file-based Pycytominer workflows."""
```
gwaybio:
```suggestion
"""Command Line Interface (CLI) for Pycytominer operations."""
```

The CLI, presumably, can be used for more than just workflows

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
...
+    """
+    if value is None:
+        return None
+    if "," in value:
+        return _split_csv_arg(value)
+    return value
+
+
+class PycytominerCLI:
+    """Command-line interface for file-based Pycytominer workflows."""
```
gwaybio:
```suggestion
    """Command Line Interface for Pycytominer operations."""
```

pyproject.toml
```diff
@@ -74,13 +74,13 @@ zarr = [
 fsspec = { version = ">=2023.1.0", optional = true }
 s3fs = { version = ">=2023.4.0", optional = true }
 boto3 = { version = ">=1.26.79", optional = true }
-fire = { version = ">=0.5.0", optional = true }
+fire = ">=0.5.0"
```
gwaybio, line 77: i think this upgrades to an essential dependency (instead of an "extra dependency for cell locations" (i.e., please move to the right place in the toml)

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
...
+        A list of strings if the input contains commas, otherwise the original string.
+    """
+    if value is None:
+        return None
+    if "," in value:
+        return _split_csv_arg(value)
+    return value
+
+
+class PycytominerCLI:
```
gwaybio, line 59: This feels rather fragile - it is true that pycytominer functions are stable, but writing it this way will require a double update if any arguments are changed in the core functions. I'm not sure if there is a way around this!

I suppose the tests will fail if updates happen in one place and not the other, so the risk of merging broken code is low, so 🤷

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
...
+        Returns:
+            The output file path.
+        """
+        profiles_df = load_profiles(profiles)
+        strata_list = _split_csv_arg(strata)
+        if isinstance(features, str) and features == "infer":
+            features_value: str | list[str] = "infer"
+        else:
+            features_value = _split_csv_arg(features)
+        subset_df = load_profiles(subset_data_file) if subset_data_file else None
```
gwaybio, line 103:
```suggestion

        # Process CLI inputs into the data structures Pycytominer expects
        strata_list = _split_csv_arg(strata)
        if isinstance(features, str) and features == "infer":
            features_value: str | list[str] = "infer"
        else:
            features_value = _split_csv_arg(features)
        subset_df = load_profiles(subset_data_file) if subset_data_file else None
```

gwaybio: It would be helpful to add a comment of the sort to this (and other pycytominer operation calls below)

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
...
+            output_type=output_type,
+            compute_object_count=compute_object_count,
+            object_feature=object_feature,
+            subset_data_df=subset_df,
+            compression_options=compression_options,
+            float_format=float_format,
+        )
+        if isinstance(result, str):
+            return result
+        raise TypeError(
```
gwaybio: I could very easily be convinced otherwise, but this doesn't feel like a `TypeError` - is there a more specific CLI error that we could raise instead? What will this look like from a user's perspective? Could we require, within the function somehow, that would prevent returning a dataframe? If my understanding is correct, the operation will run but then error out after the compute resources are used - could we instead raise an error earlier (or avoid one alltogether)?

pycytominer/cli.py
```diff
@@ -0,0 +1,397 @@
...
+
+        result = annotate(
+            profiles=profiles,
+            platemap=platemap,
+            join_on=join_on_values,
+            output_file=output_file,
+            output_type=output_type,
+            add_metadata_id_to_platemap=add_metadata_id_to_platemap,
+            format_broad_cmap=format_broad_cmap,
+            clean_cellprofiler=clean_cellprofiler,
```
gwaybio, line 174: these are the most fragile arguments... I'm wondering your thoughts on stabilizing these prior to (or after) merging this PR. I can see both ways!

gwaybio: (they are fragile, IMO, given our direction of decoupling institution-specific code)

pycytominer/cli.py
gwaybio: I also worry about redundancy of documentation

tests/test_cli.py
```diff
@@ -0,0 +1,136 @@
...
+
+import pathlib
+
+import numpy as np
+import pandas as pd
+
+from pycytominer.cli import PycytominerCLI
+
+
+def _write_profiles(tmp_path: pathlib.Path) -> tuple[pd.DataFrame, pathlib.Path]:
```
gwaybio, line 17: nice! I love these tests

I'm wondering if we should also test other scenarios:

- Erroring out if outputting a dataframe instead of a file
- A user providing a list structure instead of a comma delimited string
- A user specifying a non-existent argument
- An error propagating through one of the core pycytominer functions (e.g., file not found or something else)
- Likely others? (I think this is about 70% coverage right now, would be good to bump up)

tests/test_cli.py
```diff
@@ -0,0 +1,136 @@
+"""Tests for the Pycytominer CLI helpers."""
```
gwaybio:
```suggestion
"""Tests for the Pycytominer CLI."""
```

README.md
```diff
@@ -143,6 +143,54 @@ normalized_df = pycytominer.normalize(
 )
 ```
 
+### CLI usage (file-based workflows)
+
+Pycytominer also provides a simple CLI for file-based workflows. These commands read
+profiles from disk and write outputs to disk.
+
+```bash
```
gwaybio, line 163: Prior to this code block, it is probably worth a few notes on the expected structure of Pycytominer usage (e.g., comma delimited strings, full coverage of pycytominer functionality, etc.)

README.md
d33bs: Thanks!

README.md
d33bs: Thanks!

README.md
d33bs: Yes, makes sense, let's!

docs/install.rst
d33bs: Thanks, will do!

pycytominer/annotate.py
d33bs: Thanks!

tests/test_cli.py
d33bs: Thanks - yes, let's try to add these in!

pyproject.toml
d33bs: Thanks, will do!

README.md
d33bs: Thanks, will add this in!

pycytominer/cli.py
d33bs: Thanks!

pycytominer/cli.py
d33bs: Thanks!

README.md
d33bs: Thanks! And good question - I think we should address this in another PR!

docs/install.rst
d33bs: Ended up removing this for single-sourcing and simplicity.

pycytominer/cli.py
d33bs: Thanks - good thoughts! Maybe there's a way to simplify things with an abstraction or decorator. I'll dig in here and see. My default stance is to make things declarative and explicit here to help make the logic observable from the code (rather than break/fix processing).

d33bs: I tried going down this avenue and found it was very complex to reduce code and docs duplication here. I'm going to leave as-is with a declarative / explicit approach.

pycytominer/cli.py
d33bs: Thanks - will add this in!

pycytominer/cli.py
d33bs: Thanks - yes, updated this with a custom exception for these!

pycytominer/cli.py
d33bs: Thanks - good thoughts, I agree, but feel we should keep this PR focused, without removing these. I'll open up a new issue to address this in future work.

tests/test_cli.py
d33bs: Added more tests :)

pycytominer/cli.py
```diff
@@ -0,0 +1,403 @@
...
+            output_file=output_file,
+            output_type=output_type,
+            compute_object_count=compute_object_count,
+            object_feature=object_feature,
+            subset_data_df=subset_df,
+            compression_options=compression_options,
+            float_format=float_format,
+        )
+        if isinstance(result, str):
+            return result
```
gwaybio, line 120: how will this look to the CLI? Will it output the filename to the terminal? Maybe better to return a string that is explicit to showing output success?

README.md
```diff
@@ -143,6 +143,68 @@ normalized_df = pycytominer.normalize(
...
+  `aggregate`, `annotate`, `normalize`, `feature_select`, and `consensus`.
+- Arguments that map to Python lists (for example `--features`, `--strata`,
+  `--replicate_columns`, and `--operation`) should be passed as comma-delimited
+  strings.
+- The CLI executes one Pycytominer function per command and writes results to
+  disk. For full pipeline orchestration, chain commands in your workflow manager.
+- Output format is controlled by `--output_type` (for example `parquet`); this is
+  not inferred from the filename extension.
+
+Pycytominer also provides a simple CLI for file-based workflows.
```
gwaybio: Maybe this is the exact opposite of what I said in the last review, but I like this brief intro prior to getting into the specific bullet points.

pycytominer/cli.py
d33bs: Ah such a good point! I'll update to something more friendly here.

README.md
d33bs: Thanks - I'll maneuver things around.

d33bs:
Thanks @gwaybio for the reviews! Merging this in.

Issue #574, opened 2025-09-02
FeatureRequest: enable mulitple dataframe support using `narwhals` (or similar)


d33bs:
[`narwhals`](https://github.com/narwhals-dev/narwhals) provides an abstraction for dataframe libraries (Pandas, Polars, etc.) to help enable a unified API regardless of what type of dataframe might be used for an implementation. We could add this to `pycytominer` where needed to help allow folks to use the project's work regardless what type of dataframe library they might be using.

d33bs:
Closing comments from #626:
> Hi @gwaybio, thanks for all the comments. After thinking about this a bit more (generally your high-level comment about narwhals at-all, I think we should close this without merge and navigate design outside of this PR (maybe in a new issue). Thank you just the same for your thorough comments in the PR.
> 
> Of the methods you mentioned I think Apache Arrow data structures would be most consistent. Handling these with Pandas is "slippery" because of the history of NumPy use in the package (there are no "strict" mode switches that I know of meaning data can "backslide" into NumPy sometimes). If all data is in Arrow then I believe converting from / to Polars DataFrames would pose minimal challenge.

axiomcura:
Hey @d33bs, do you think this issue is still relevant? I feel the underlying concern still exists, even after the migration from Pandas to Polars. That said, Polars does provide a safer and more predictable approach to handling data types because of its stricter type enforcement.

One thing I’m wondering is whether we actually need this level of support in practice. Have there been users working directly with PyArrow tables in Pycytominer?

One possible solution could be to have a dedicated utility function responsible for converting supported dataframe-like objects into a Polars DataFrame, and then rely on the Polars API internally throughout the package. I feel this would only require an additional line at the beginning of each function, something like:

```python
if not isinstance(df, pl.DataFrame):
    df = convert_to_polars(df)
```

This could also be implemented within the loader function itself. However, there are two potential downsides: 1) users may need some familiarity with how Polars behaves, which could be difficult for new users coming from Pandas workflows. 2) Converting dataframes is very expensive and may drive Pycytominer's performance to be slower.

I can definitely see why supporting multiple dataframe types is valuable. I think the main challenge is finding the simplest and most maintainable abstraction. A single conversion helper feels like a clean approach, but we would still need to decide what the outputs should be. Should functions always return Polars DataFrames, or should users be able to specify the output type (e.g., Pandas vs. Polars) through a parameter?


Looking forward to hear your thoughts c: !

d33bs:
[...]

PR #626, opened 2026-02-18
feat: add narwhals and update aggregate

d33bs:
# Description

This PR adds [narwhals](https://github.com/narwhals-dev/narwhals) to the project for dataframe-agnostic handling. I felt that it could get complex to migrate all functions to this new API so started with `aggregate` and, by proxy, `infer_cp_features` as initial updates in this area.  These changes make `aggregate` compatible with, for example, `polars` dataframes, `pyarrow` tables, and more (through `narwhals`). 

References #574 (part of what should likely be multi-PR work on this)

# Checklist

----
📚 Documentation preview 📚: https://pycytominer--626.org.readthedocs.build/en/626/

### DIFF
--- a/.github/actions/setup-env/action.yaml
+++ b/.github/actions/setup-env/action.yaml
@@ -76,4 +76,4 @@ runs:
       shell: bash
       if: ${{ inputs.install-deps == 'true' }}
       run: |
-        poetry install --with dev --all-extras -v --no-interaction
+        poetry install --with dev,docs --all-extras -v --no-interaction
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -101,12 +101,6 @@ jobs:
           cache-venv: true
           setup-poetry: true
           install-deps: true
-      - name: Install docs dependencies
-        # Keep docs dependencies in this job only. `sphinx-autobuild` pulls
-        # `watchfiles`, and current `watchfiles`/`PyO3` builds fail on Python 3.14.
-        # Avoid installing docs extras in matrix test jobs until upstream support lands.
-        run: |
-          poetry install --with docs -v --no-interaction
       - name: Run sphinx-build
         run: |
           poetry run sphinx-build docs/ docs-build/


pycytominer/aggregate.py
```diff
@@ -14,7 +14,7 @@
 @write_to_file_if_user_specifies_output_details
 def aggregate(
     population_df: pd.DataFrame,
-    strata: list[str] = ["Metadata_Plate", "Metadata_Well"],
+    strata: Union[list[str], str] = ["Metadata_Plate", "Metadata_Well"],
```
gwaybio, line 17: Shouldn't the line above (population_df) be changed as well?

pycytominer/aggregate.py
gwaybio, line 34: Shouldn't this be dataframe-like too?

pycytominer/aggregate.py
```diff
@@ -73,52 +73,54 @@ def aggregate(
 
     # Check that the operation is supported
     operation = check_aggregate_operation(operation)
+    narwhals_df = nw.from_native(population_df, eager_only=True)
```
gwaybio, line 76: for someone reading this code who isn't aware of narwals, this line will be confusing. Why does it matter internally for this to be named this way? If it is important, then I suggest something like `population_narwhals_df` and include a comment, in the docstring why this is important.

pycytominer/aggregate.py
```diff
@@ -73,52 +73,54 @@ def aggregate(
...
+    population_columns = list(narwhals_df.columns)
+
+    if isinstance(strata, str):
+        strata = [strata]
+
+    if subset_data_df is not None:
+        subset_nw = nw.from_native(subset_data_df, eager_only=True)
```
gwaybio, line 83: this variable name is different (`nw` instead of `narwhals`), we should keep consistent

pycytominer/aggregate.py
```diff
@@ -73,52 +73,54 @@ def aggregate(
...
+        subset_columns = list(subset_nw.columns)
+        subset_nw = nw.from_dict(
+            subset_nw.to_dict(as_series=False),
+            backend=nw.get_native_namespace(nw.to_native(narwhals_df)),
+        )
+        narwhals_df = subset_nw.join(
+            narwhals_df, on=subset_columns, how="inner"
+        ).select(*population_columns)
```
gwaybio, line 91: This is a bit confusing to me, maybe b/c I am unfamiliar with narwhals? The added complexity here (compared to a simple merge statement) makes me a bit itchy!

pycytominer/aggregate.py
```diff
@@ -73,52 +73,54 @@ def aggregate(
...
-    # Note: type ignore added below to address the change in variable types for
-    # label `population_df`.
-    population_df = population_df.groupby(strata, dropna=False)  # type: ignore[assignment]
+    narwhals_df = narwhals_df.select(
+        *strata,
+        *[nw.col(feature).cast(nw.Float64).alias(feature) for feature in features],
+    )
+    aggregated = narwhals_df.group_by(*strata, drop_null_keys=False).agg(*[
+        getattr(nw.col(feature), operation)().alias(feature) for feature in features
+    ])
```
gwaybio, line 110: i am cringing reading this syntax! 😬 Is this standard narwhals? It feels unpythonic

tests/test_cyto_utils/test_feature_infer.py
```diff
@@ -1,4 +1,6 @@
 import pandas as pd
+import polars as pl
+import pyarrow as pa
```
gwaybio, line 3: Is this now adding two dependencies? Oh, I suppose you can prob add them to test dependencies only

tests/test_aggregate.py
```diff
@@ -135,6 +137,106 @@ def test_aggregate_mean_subsetvar():
     assert aggregate_result.equals(expected_result)
 
 
+def test_aggregate_mean_single_feature_string():
+    """
+    Testing aggregate pycytominer function with a single string feature input
+    """
+    aggregate_result = aggregate(
+        population_df=data_df, strata=["g"], features="Cells_x", operation="mean"
```
gwaybio, line 145:
```suggestion
        population_df=data_df, strata="g", features="Cells_x", operation="mean"
```

True?

gwaybio: fwiw, I've never liked the "strata" argument. It is super important for `aggregate` but the name is confusing IMO. is there a way, in the future outside this PR, to warn about an incoming deprecation of variable naming? We could change this in 2.0

tests/test_aggregate.py
```diff
@@ -135,6 +137,106 @@ def test_aggregate_mean_subsetvar():
...
+    )
+
+    expected_result = pd.DataFrame({"g": ["a", "b"], "Cells_x": [4, 3]})
+    expected_result.Cells_x = expected_result.Cells_x.astype(float)
```
gwaybio, line 149: If my comment above on R145 is true, then we could instead compare to the aggregate call on `[g]` as well

tests/test_aggregate.py
```diff
@@ -135,6 +137,106 @@ def test_aggregate_mean_subsetvar():
...
+    expected_result = pd.concat([
+        pd.DataFrame({"g": "a", "Cells_x": [3], "Nuclei_y": [3]}),
+        pd.DataFrame({"g": "b", "Cells_x": [3], "Nuclei_y": [3]}),
+    ]).reset_index(drop=True)
+    expected_result = expected_result.astype(dtype_convert_dict)
+
+    pd.testing.assert_frame_equal(
+        aggregate_result.sort_values("g").reset_index(drop=True),
+        expected_result.sort_values("g").reset_index(drop=True),
+    )
```
gwaybio, line 174: there's a bit of duplicate code here (repeated in the functions above and below). If it is not too challenging, I recommend reducing this duplication.

pyproject.toml
```diff
@@ -101,6 +104,7 @@ pandas-stubs = [
   { version = "^2.2.2.240909", python = ">=3.10" },
 ]
 types-sqlalchemy = "^1.4.53.38"
+polars = "^1.38.1"
```
gwaybio, line 107: also pyarrow?

gwaybio (changes requested):
Loving the march to Pycytominer 2.0!

i have some serious concerns, however, with narwhals syntax. Without knowing what the speedup is, my intuition is to revert back to pandas to maintain code readability. Are there options, within pandas, that might make multi-dataframe support more readable?

A second option would be to manage multi-dataframe support in a cyto_utility file that is super documented.

A third option would be to instead be opinionated on pyarrow (or polars).

In other words, I'm not convinced that extending support to other dataframe libraries is worth the downgraded readability. Happy to be wrong!

d33bs:
Hi @gwaybio, thanks for all the comments. After thinking about this a bit more (generally your high-level comment about narwhals at-all, I think we should close this without merge and navigate design outside of this PR (maybe in a new issue). Thank you just the same for your thorough comments in the PR.

Of the methods you mentioned I think Apache Arrow data structures would be most consistent. Handling these with Pandas is "slippery" because of the history of NumPy use in the package (there are no "strict" mode switches that I know of meaning data can "backslide" into NumPy sometimes). If all data is in Arrow then I believe converting from / to Polars DataFrames would pose minimal challenge.

PR #657, opened 2026-04-13
Update annotate ingest qc parquet

jenna-tomkinson:
# Description

In this PR, I have updated the `external_metadata` parameter in `annotate` to input parquet files and add qc.parquet columns to the profiles prior to the `cp_clean` helper function to avoid issues with column naming. I also added Metadata_Site as apart of the `cp_clean` helper function as it is hardcoded for only plate and well so I previously had to make a band-aid for this for projects.

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Single shared join-key for external-metadata merges and matching CLI option.
  * Acceptance of file-backed external metadata plus automatic preprocessing to normalize/rename metadata columns and site identifiers.

* **Tests**
  * Added/expanded tests for external metadata preprocessing and merge behaviors, including parquet/file-backed cases.

* **Chores**
  * Updated pre-commit hook configuration for dependency locking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cytomining/pycytominer/pull/657?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
         entry: python -m uv lock --check
         language: python
         additional_dependencies:
-        - uv
+        -   uv
```
d33bs, line 21: This looks like a typo. Consider reverting this change.

pycytominer/annotate.py
```diff
@@ -17,6 +16,48 @@
 from pycytominer.cyto_utils.util import write_to_file_if_user_specifies_output_details
 
 
+def _prefix_external_metadata_columns(
+    external_metadata: pd.DataFrame,
+) -> pd.DataFrame:
+    """Helper function for `external_metadata` parameters in `annotate` that adds
+    `Metadata_` prefix to external metadata columns unless already CP-style.
```
d33bs, line 23: Consider documenting explicitly what "CP" means here (Pycytominer uses more than CellProfiler styles so it's important to make sure this is understandable beyond that scope).

pycytominer/annotate.py
```diff
@@ -17,6 +16,48 @@
...
+    """
+
+    protected_prefixes = (
+        "Metadata_",
+        "Image_Metadata_",
+        "Cells_",
+        "Cytoplasm_",
+        "Nuclei_",
+        "Image_",
+    )
```
d33bs, line 46: Consider making this configurable somehow. What would happen if someone has another compartment, for example?

pycytominer/annotate.py
```diff
@@ -59,11 +101,15 @@ def annotate(
     clean_cellprofiler: bool, default True
         Clean specific CellProfiler feature names.
     external_metadata : str, optional
-        File with additional metadata information
-    external_join_left : str, optional
-        Merge column in the profile metadata.
-    external_join_right: str, optional
-        Merge column in the external metadata.
+        File with additional metadata information.
+        Most common use case is a QC.parquet file with QC flags for each profile.
```
d33bs, line 105: Where does this file come from? Consider documenting.

pycytominer/annotate.py
```diff
@@ -87,6 +133,11 @@ def annotate(
 
     # Load Data
     profiles = load_profiles(profiles)
+    cleaned_profile_metadata_cols = [
+        column.replace("Image_Metadata_", "Metadata_", 1)
```
d33bs, line 137: This feels really narrow and could lead to challenges because it might be brittle and need frequent updates. Would it make sense to make this configurable to other options or abstracted to a higher level?

pycytominer/annotate.py
```diff
@@ -59,11 +101,15 @@ def annotate(
     clean_cellprofiler: bool, default True
         Clean specific CellProfiler feature names.
```
d33bs, line 102: Because we make modifications to this function, consider being more prescriptive in the documentation here to spell out exactly what this does and why.

pycytominer/annotate.py
```diff
@@ -110,42 +161,45 @@ def annotate(
...
-        # Make a copy of the external metadata to avoid modifying the original dataframe
-        external_metadata = external_metadata.copy()
-
-        external_metadata.columns = pd.Index([
-            f"Metadata_{x}" if not x.startswith("Metadata_") else x
-            for x in external_metadata.columns
-        ])
-
+        external_metadata = _prefix_external_metadata_columns(external_metadata)
+        external_merge_left = (
```
d33bs, line 170: Consider using these evaluations directly where these variables are referenced below (to avoid a confusion between "join" and "merge" and also to avoid having to create new variables which are only used once).

tests/test_annotate.py
```diff
@@ -142,6 +148,113 @@ def test_annotate_external():
     pd.testing.assert_frame_equal(result, expected_result)
 
 
+def test_annotate_external_parquet_joins_before_cp_clean(tmp_path):
```
d33bs, line 151: Consider adding a short docstring after each test describing what it does.

pycytominer/annotate.py
d33bs, line 105: "Most common" might be subjective here. Consider avoiding this to help bolster for multiple perspectives (and time tense).

d33bs (changes requested):

.pre-commit-config.yaml
jenna-tomkinson: fixed!

pycytominer/annotate.py
jenna-tomkinson: Fixed!

pycytominer/annotate.py
jenna-tomkinson: I have updated this to use infer_cp_features instead. In the annotate stage we should not care what the features are at all, only what becomes metadata. This should solve this comment.

pycytominer/annotate.py
jenna-tomkinson: Documented!

pycytominer/annotate.py
jenna-tomkinson: I have removed this irrelevant addition as it was created to assign order to the metadata columns which is unnecessary and does not enhance the output.

pycytominer/annotate.py
jenna-tomkinson: I have added more documentation.

pycytominer/annotate.py
jenna-tomkinson: Good point! I have updated that so that we don't create new variables when we don't need to.

tests/test_annotate.py
jenna-tomkinson: Updated!

pycytominer/annotate.py
jenna-tomkinson: I added it as a cue that this will be where if someone is using coSMicQC what they should be looking for. I agree it is subjective but if I hope to encourage and make a standard in the ecosystem then I choose to keep this wording.

pycytominer/annotate.py
d33bs: If you've configured this to use infer_cp_features why retain this block? Do we need it (is it duplicative)? If it's duplicative then we should remove it, relying on single responsibility principles.

pycytominer/annotate.py
```diff
@@ -110,42 +161,45 @@ def annotate(
...
     meta_cols = infer_cp_features(annotated, metadata=True)
     other_cols = annotated.drop(meta_cols, axis="columns").columns.tolist()
-
-    annotated = annotated.loc[:, meta_cols + other_cols]
+    prioritized_meta_cols = [
+        column for column in cleaned_profile_metadata_cols if column in meta_cols
+    ]
+    remaining_meta_cols = [
+        column for column in meta_cols if column not in prioritized_meta_cols
+    ]
```
d33bs: Consider using set comprehensions over set-based loop bases (we shouldn't have duplicate column names but that doesn't mean they might not be there somehow).

tests/test_annotate.py
d33bs: The added tests added here seem to "rhyme". Consider using [pytest's parametrize](https://docs.pytest.org/en/stable/example/parametrize.html) to define the test once with multiple input iterations.

d33bs (approved):

.pre-commit-config.yaml
d33bs: Consider resolving this (it still seems unfixed).

pycytominer/annotate.py
```diff
@@ -110,42 +161,45 @@ def annotate(
...
+                right_on=external_merge_right,
                 how="left",
                 suffixes=(None, "_external"),
             )
             .reset_index(drop=True)
             .drop_duplicates()
         )
 
+    if clean_cellprofiler:
+        annotated = cp_clean(annotated)
```
gwaybio, line 151: Philosophically, cleaning/helper functionality has been kept separate from the core functions. This philosophy changes in this PR with the addition of `_prefix_external_metadata_columns()` defined in `annotate.py`. 

To maintain the philosophy, we could move `_prefix_external_metadata_columns()` to cyto_utils (or even move the functionality to `cp_clean()`) or we could change philosophy given the upcoming push to 2.0

From a testing perspective (and also an understandability/readability perspective) my preference would be to separate cleaning/helpers from core while minimally (and clearly) updating column names in a single location, if possible.

pycytominer/annotate.py
gwaybio: It looks like some of this already happens in [`cp_clean()`](https://github.com/cytomining/pycytominer/blob/main/pycytominer/cyto_utils/annotate_custom.py).

I make a similar comment below, but we should 1) only sparingly rename columns and 2) minimize the locations in the code where this happens.

pycytominer/annotate.py
gwaybio: I do not understand why these will be independent sets - won't remaining_meta_cols be empty?

gwaybio (changes requested):
We should discuss reparameterization of column names prior to merge

.pre-commit-config.yaml
jenna-tomkinson: Is this another pre-commit issue? Like we saw in coSMicQC. I am not actively changing it so something must do this automatically after I manually change it.

pycytominer/annotate.py
jenna-tomkinson: This has been removed in regards to a previous comment due to being unneccessary. Thanks for pointing out though!

.pre-commit-config.yaml
d33bs:
```suggestion
          - uv
```

pycytominer/annotate.py
jenna-tomkinson: Dropped after seeing that we do not enforce metadata column order before and likely shouldn't change it.

tests/test_annotate.py
jenna-tomkinson: Oh that is a really cool feature! I have added parametrize so that we don't have duplicate tests with minor differences.

pycytominer/annotate.py
jenna-tomkinson: I decided after some thinking that what is in scope of this PR is to move this helper function to annotate_custom script where cp_clean lives but not including it in cp_clean. I also decided it does not reside in load_platemap as the naming suggests it only loads platemaps not external metadata. Unfortunately, this does mean that there is minor duplication of concepts but I think this issue would deserve its own PR.

I will open an issue after this and include these details to address later.

jenna-tomkinson: Issue created: #660

.pre-commit-config.yaml
jenna-tomkinson: `additional_dependencies: [uv]`

Pre-commit updates this during the run so if I manually change it pre-commit adds to back. This change passes the linter.

tests/test_annotate.py
```diff
@@ -105,18 +106,24 @@ def test_annotate_merge():
 
 def test_annotate_external():
     # Test that the external_metadata
+    external_metadata_df = EXTERNAL_METADATA_DF.rename(
```
gwaybio, line 109: The comment on line 108 appears to be incomplete

pycytominer/cyto_utils/__init__.py
```diff
@@ -1,4 +1,8 @@
-from .annotate_custom import annotate_cmap, cp_clean
+from .annotate_custom import (
+    annotate_cmap,
+    cp_clean,
+    prepare_external_metadata_for_annotate,
```
gwaybio, line 4: Please include tests for this function

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,64 @@
...
 
+import contextlib
+
 import numpy as np
 import pandas as pd
 
+from pycytominer.cyto_utils.features import infer_cp_features
+
+
+def prepare_external_metadata_for_annotate(
```
gwaybio, line 13: I'm wondering why this function is different now from the last version? The last version seemed more appropriate (defining protected columns, for example). As well, I'm wondering why we need to extract the existing_cp_columns in this way.

These two changes (plus not having tests) make me a bit nervous merging this in

tests/test_annotate.py
```diff
@@ -142,6 +149,96 @@ def test_annotate_external():
...
+        profiles=profiles_df,
+        platemap=platemap_df,
+        join_on=[
+            ["Metadata_Plate", "Metadata_Well"],
+            [
+                "Image_Metadata_Plate",
+                "Image_Metadata_Well",
+            ],
+        ],
+        external_metadata=str(external_metadata_path),
```
gwaybio, line 235: Should external metadata also allow a path object? I believe we should have examples in Pycytominer of handling path args

tests/test_annotate.py
```diff
@@ -142,6 +149,96 @@ def test_annotate_external():
...
+    result = annotate(
+        profiles=profiles_df,
+        platemap=platemap_df,
+        join_on=[
+            ["Metadata_Plate", "Metadata_Well"],
+            [
+                "Image_Metadata_Plate",
+                "Image_Metadata_Well",
+            ],
+        ],
```
gwaybio, line 234: Now that I'm refreshing my memory of this `join_on` parameter, I'm wondering why we've decided to include `external_join_left` and `external_join_right`. Wouldn't it be more consistent to allow a list of lists of strings in a single `external_join_on` parameter?

gwaybio (changes requested):
We're close Jenna!

tests/test_annotate.py
jenna-tomkinson: Thanks for catching that, updated!

pycytominer/cyto_utils/__init__.py
jenna-tomkinson: A new test is added to make sure that the expected output occurs, which is that if the external metadata contains columns in CP-format (e.g., Image_Metadata_Well, Image_Metadata_Plate, Image_Count_Cells, etc), then it will not rename those columns until cp_clean later (if included in the function) and only add a Metadata_ prefix to the non-CP columns (e.g., 'pathway', 'broad_sample', etc.).

In the context of coSMicQC `qc.parquet`, columns will already be named with a Metadata_ prefix, but the CP columns will include the "Image_Metadata_" prefix, which cp_clean handles later.

Let me know if there are other tests necessary here.

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: Apolgoies for not getting back to this sooner, so my recollection of the previous version is fuzzy. IIRC, the protected columns felt too hardcoded and didn't feel like it made sense to have in this PR. 

We extract existing cp_columns in this way to avoid renaming that occurs downstream, since adding external metadata will occur before the platemap merge, so renaming similar columns used for merging might lead to potential issues if it occurs multiple times.

I have added a test to confirm that this will work as expected similar to how annotation occurs for regular platemaps.

tests/test_annotate.py
jenna-tomkinson: It likely should take a path also, I will correct this.

jenna-tomkinson: I can confirm that external metadata can handle both a path and a pandas DataFrame, and I have updated the docstring accordingly to reflect that. Let me know if this does not answer your question as you expected.

tests/test_annotate.py
jenna-tomkinson: I am following the legacy code to avoid any major API changes, which was only left and right join on. If you think it is fine to drop them, then I can leave the new "external_join_on" that I had included.

tests/test_annotate.py
gwaybio: sounds good - why are we coercing a string (with `str()`) then?

jenna-tomkinson: Ah, good point! Removing it now.

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,67 @@
...
+    """
+    # Set existing CellProfiler columns to avoid adding a Metadata_ prefix
+    # which will occur during cp_clean
+    existing_cp_columns = set()
+
+    # Ty to detect CP metadata columns
+    with contextlib.suppress(ValueError):
+        existing_cp_columns.update(infer_cp_features(external_metadata, metadata=True))
+
+    with contextlib.suppress(ValueError):
```
gwaybio, line 44:
```suggestion
    # Capture non-numeric "Image_" prefixed columns
    with contextlib.suppress(ValueError):
```

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,67 @@
...
+    Returns
+    -------
+    external_metadata : pd.DataFrame
+        External metadata with columns renamed to be compatible with annotate() conventions.
+    """
+    # Set existing CellProfiler columns to avoid adding a Metadata_ prefix
+    # which will occur during cp_clean
+    existing_cp_columns = set()
+
+    # Ty to detect CP metadata columns
```
gwaybio:
```suggestion
    # Capture "Metadata_" prefixed columns
```

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,67 @@
...
+    with contextlib.suppress(ValueError):
+        existing_cp_columns.update(infer_cp_features(external_metadata, metadata=True))
+
+    with contextlib.suppress(ValueError):
+        existing_cp_columns.update(
+            infer_cp_features(external_metadata, image_features=True)
+        )
+
+    # Keep Image_Metadata_* columns unchanged so external joins can still happen
+    # before cp_clean standardizes them.
```
gwaybio:
```suggestion
    # Capture "Image_Metadata_" prefixed columns with string dtypes
```

The existing comment doesn't really fit with the line it's describing

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,67 @@
...
+    existing_cp_columns.update([
+        column
+        for column in external_metadata.columns
+        if isinstance(column, str) and column.startswith("Image_Metadata_")
+    ])
+
+    # Rename only non-CP columns with Metadata_ prefix to be in expected column naming format
```
gwaybio:
```suggestion
    # Rename unprotected columns (outside "existing_cp_columns") with Metadata_ prefix to be in expected column naming format
```

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,67 @@
...
+        External metadata to be merged in with profiles in annotate().
+
+    Returns
+    -------
+    external_metadata : pd.DataFrame
+        External metadata with columns renamed to be compatible with annotate() conventions.
+    """
+    # Set existing CellProfiler columns to avoid adding a Metadata_ prefix
+    # which will occur during cp_clean
+    existing_cp_columns = set()
```
gwaybio:
```suggestion
    # Protect certain columns with specific prefixes ("Metadata_", "Image_Metadata", and "Image_").
    # The function will protect only non-numeric columns with "Image_" prefix.
    # This prevents adding a "Metadata_" prefix, which will occur during `cp_clean`, later
    protected_columns_to_avoid_metadata_prefix = set()
```

Changing the comments and variable names to increase clarity. Please review!

I _think_ the "numeric "Image_" comment holds here, based on https://github.com/cytomining/pycytominer/blob/main/pycytominer/cyto_utils/features.py#L108

gwaybio: if you approve of the variable change, please update throughout this function

tests/test_cyto_utils/test_annotate_custom.py
```diff
@@ -55,6 +58,32 @@
 broad_platemap_df = platemap_df.assign(Metadata_broad_sample=example_broad_samples)
 
 
+def test_prepare_external_metadata_for_annotate():
```
gwaybio, line 61: 🎉

tests/test_annotate.py
gwaybio: i believe, if I'm reading correctly, the legacy code was simply `join_on` and accepted 1) a string, 2) a list, 3) a list of two strings, 4) a length-two list of lists of strings

tests/test_annotate.py
gwaybio: Worth updating the tests?

gwaybio (changes requested):
Changes requested mostly to comments and variable naming - a few other discussion items!

tests/test_annotate.py
jenna-tomkinson: As discussed at 1on1, we discovered that the legacy code for external metadata was left and right join so I will be removing them in this PR

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: Added!

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: Updated!

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: Updated!

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: I agree this is more explicit and clear, I will make this update throughout the function.

pycytominer/cyto_utils/annotate_custom.py
jenna-tomkinson: Thanks for catching, this has been corrected!

pycytominer/cli.py
```diff
@@ -165,6 +167,9 @@ def annotate(
         join_on_values = _split_csv_arg(join_on)
         if len(join_on_values) != 2:
             raise ValueError("join_on must contain exactly two values.")
+        external_join_on_values = (
+            _split_csv_arg(external_join_on) if external_join_on is not None else None
+        )
```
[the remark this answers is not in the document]
jenna-tomkinson: @coderabbitai I have removed both of those parameters to only have one parameter to join metadata, does that remove this issue?

tests/test_annotate.py
jenna-tomkinson: Agreed, will update!

pycytominer/annotate.py
```diff
@@ -110,42 +114,30 @@ def annotate(
...
+        external_metadata = prepare_external_metadata_for_annotate(external_metadata)
 
         annotated = (
             annotated
             .merge(
                 external_metadata,
-                left_on=external_join_left,
-                right_on=external_join_right,
+                on=external_join_on,
                 how="left",
```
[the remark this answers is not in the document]
jenna-tomkinson: @coderabbitai I added a new check for this in the latest commit, can you check it?

jenna-tomkinson:
@gwaybio ready when you are!

pycytominer/cyto_utils/annotate_custom.py
```diff
@@ -2,9 +2,71 @@
...
 
+import contextlib
+
 import numpy as np
 import pandas as pd
 
+from pycytominer.cyto_utils.features import infer_cp_features
+
+
+def prepare_external_metadata_for_annotate(
```
gwaybio, line 13: this function looks great!

gwaybio (approved):
congrats! I'll merge this in!

Issue #170, opened 2021-09-19
modz can result in a zero vector


shntnu:
I haven't been able to ping this down but here are some quick notes

I am using this profile file

https://github.com/shntnu/pooled-cell-painting-data-analysis/blob/a060fccda5e0395f51138bbcff1557e1783e9dcc/metrics_calculator/results/1cdf3085/params.yaml#L3-L13

```r
df <- read_parquet("metrics_calculator/results/1cdf3085/profiles.parquet")
df %>% filter(Metadata_Foci_Barcode_MatchedTo_GeneCode == "ARL14EP") %>% slice(1) %>% select(1:20) %>% glimpse()
```

```
Rows: 1
Columns: 20
$ Metadata_Foci_Barcode_MatchedTo_GeneCode <chr> "ARL14EP"
$ Metadata_reference_or_other              <chr> "reference"
$ Metadata_Foci_Barcode_MatchedTo_Barcode  <chr> "ACAGCCACTTGCGGTCACAG"
$ Metadata_Plate                           <chr> ""
$ Cells_AreaShape_Eccentricity             <dbl> 0
$ Cells_AreaShape_Extent                   <dbl> 0
$ Cells_AreaShape_FormFactor               <dbl> 0
$ Cells_AreaShape_Orientation              <dbl> 0
$ Cells_AreaShape_Zernike_0_0              <dbl> 0
$ Cells_AreaShape_Zernike_1_1              <dbl> 0
$ Cells_AreaShape_Zernike_2_2              <dbl> 0
$ Cells_AreaShape_Zernike_3_1              <dbl> 0
$ Cells_AreaShape_Zernike_3_3              <dbl> 0
$ Cells_AreaShape_Zernike_4_0              <dbl> 0
$ Cells_AreaShape_Zernike_4_2              <dbl> 0
$ Cells_AreaShape_Zernike_4_4              <dbl> 0
$ Cells_AreaShape_Zernike_5_1              <dbl> 0
$ Cells_AreaShape_Zernike_5_3              <dbl> 0
$ Cells_AreaShape_Zernike_6_2              <dbl> 0
$ Cells_AreaShape_Zernike_6_4              <dbl> 0
```

gwaybio:
related to #162

PR #662, opened 2026-04-28
fixed modz bug: dropping groups where  contains NaNs

### LINKED ISSUE #162: MODZ drops replicate groups containing missing values
Running `consensus(operation="modz")` will silently drop groups (as specified in the argument `replicate_columns`) if one of the groups contains `NA` in one of the `replicate_columns`.

The correct behavior would be to consider NA as a valid group member, or, at the very least, warn the user that MODZ is invalid for groups containing missing values.

axiomcura:
# Description
This PR addresses where consensus(operation="modz") was silently dropping groups that contained NaN values in the replicate_columns.

This was done by updating the `groupby` operation in modz.py to use `dropna=False`. This ensures that NaN values in metadata are treated as valid group identifiers rather than being excluded from the consensus output.

Tests were also generated to reflect this change. 

This closes #162 and #170 

# Checklist

- [] I have made corresponding changes to the documentation.

### DIFF
--- a/pycytominer/cyto_utils/modz.py
+++ b/pycytominer/cyto_utils/modz.py
@@ -150,13 +150,16 @@ def modz(
     subset_features = list(set(replicate_columns + features))
     population_df = population_df.loc[:, subset_features]
 
-    # warn users about using modz with NaN values
-    # NaN values will be treated as a valid group
-    if population_df.isnull().any().any():
+    # warn users about having NaN values in replicate columns
+    if population_df[replicate_columns].isnull().any().any():
         warnings.warn(
-            "NaN values detected. NaNs in 'replicate_columns' form their own "
-            "group; NaNs in 'features' are skipped in correlation but treated "
-            "as 0 in the weighted sum."
+            "NaN values detected. NaNs in 'replicate_columns' form their own group"
+        )
+
+    if population_df[features].isnull().any().any():
+        raise ValueError(
+            "NaN values detected in feature columns. NaNs miscalculate the "
+            "pairwise correlations used to weight samples."
         )
 
     modz_df = (
--- a/pycytominer/cyto_utils/modz.py
+++ b/pycytominer/cyto_utils/modz.py
@@ -2,6 +2,7 @@
 Module for performing MODZ (modified z-score) transformations
 """
 
+import warnings
 from typing import Union
 
 import numpy as np
@@ -149,12 +150,21 @@ def modz(
     subset_features = list(set(replicate_columns + features))
     population_df = population_df.loc[:, subset_features]
 
+    # warn users about using modz with NaN values
+    # NaN values will be treated as a valid group
+    if population_df.isnull().any().any():
+        warnings.warn(
+            "NaN values detected. NaNs in 'replicate_columns' form their own "
+            "group; NaNs in 'features' are skipped in correlation but treated "
+            "as 0 in the weighted sum."
+        )
+
     modz_df = (
         population_df
-        .groupby(replicate_columns)
+        .groupby(replicate_columns, dropna=False)[features]
         .apply(
             lambda x: modz_base(
-                x.loc[:, features],
+                x,
                 method=method,
                 min_weight=min_weight,
                 precision=precision,
--- a/tests/test_cyto_utils/test_modz.py
+++ b/tests/test_cyto_utils/test_modz.py
@@ -204,7 +204,7 @@ def test_modz_nan_in_replicate_columns():
     )
 
     # We expect our newly implemented warning since the input structure contains NaNs
-    with pytest.warns(UserWarning, match="Missing values \\(NaN\\) detected in input."):
+    with pytest.warns(UserWarning, match="NaN values detected."):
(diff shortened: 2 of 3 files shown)


pycytominer/cyto_utils/modz.py
```diff
@@ -149,12 +150,21 @@ def modz(
...
+            "as 0 in the weighted sum."
+        )
+
     modz_df = (
         population_df
-        .groupby(replicate_columns)
+        .groupby(replicate_columns, dropna=False)[features]
         .apply(
             lambda x: modz_base(
-                x.loc[:, features],
```
d33bs, line 157: Nice cleanup here.

pycytominer/cyto_utils/modz.py
```diff
@@ -149,12 +150,21 @@ def modz(
     subset_features = list(set(replicate_columns + features))
     population_df = population_df.loc[:, subset_features]
 
+    # warn users about using modz with NaN values
+    # NaN values will be treated as a valid group
+    if population_df.isnull().any().any():
+        warnings.warn(
+            "NaN values detected. NaNs in 'replicate_columns' form their own "
+            "group; NaNs in 'features' are skipped in correlation but treated "
+            "as 0 in the weighted sum."
```
d33bs: What are the scenarios we'd want to include NaN's? Consider documenting this. If this is a common convention I feel we should outline with a citation of some kind.

d33bs: Do nan's count as zero for the weighted sum even with a non-zero minimum weight?

pycytominer/cyto_utils/modz.py
```diff
@@ -149,12 +150,21 @@ def modz(
...
+        )
+
     modz_df = (
         population_df
-        .groupby(replicate_columns)
+        .groupby(replicate_columns, dropna=False)[features]
```
d33bs, line 167: Why not keep existing functionality and raise with an exception when encountered (since replacing Nan's with 0 isn't equivalent)?

tests/test_cyto_utils/test_modz.py
```diff
@@ -191,3 +193,33 @@ def test_modz_unbalanced_sample_numbers():
...
+        Metadata_nan=["c", "c", "c", np.nan, np.nan, np.nan]
+    )
+
+    # We expect our newly implemented warning since the input structure contains NaNs
+    with pytest.warns(UserWarning, match="NaN values detected."):
+        consensus_df = modz(
+            data_replicate_nan_df,
+            replicate_columns="Metadata_nan",
+            min_weight=0,
+            precision=precision,
```
d33bs, line 209: Consider testing with some variations of these parameters to be safe.

pycytominer/cyto_utils/modz.py
```diff
@@ -149,12 +150,21 @@ def modz(
     subset_features = list(set(replicate_columns + features))
     population_df = population_df.loc[:, subset_features]
 
+    # warn users about using modz with NaN values
+    # NaN values will be treated as a valid group
+    if population_df.isnull().any().any():
```
d33bs: Consider a performance or syntactical benefit to using something like this instead (vectorized operation over a numpy array to find nan's): (
```suggestion
    if population_df.isna().values.any():
```

pycytominer/cyto_utils/modz.py
axiomcura: Correct, that is what the current implementation does. What happens is that when a sample is poorly correlated with the others, or contains `NaN` values that prevent the correlation calculation, its `raw_weight` becomes `NaN`.

When applying:

```python
raw_weight.clip(lower=min_weight)
```

to a `NaN`, the result is still `NaN`. Therefore, if we then calculate:

```python
raw_weights.sum()
```

the total weight also becomes `NaN`.

During the weighted sum:

```python
weighted_df: pd.DataFrame = population_df.mul(weights, axis="columns")
modz_df = weighted_df.sum(axis="columns")
```

`weighted_df.sum(axis="columns")` uses `skipna=True` by default. As a result, any `NaN` values are skipped during the sum, which mathematically makes their contribution equivalent to `0`.

pycytominer/cyto_utils/modz.py
axiomcura: I prefer this approach. Let's warn users about NaNs in `replicate_columns` and raise a warning if any feature contains NaNs. 

However, is this too restrictive? What if users want to include NaNs? Should we add a parameter, "allow_nans", to the function?

Open to opinions c:

tests/test_cyto_utils/test_modz.py
axiomcura: Done!

pycytominer/cyto_utils/modz.py
axiomcura: I think the issue is that data may be unintentionally dropped when `replicate_columns` contain `NaN` values. This is primarily a data provenance concern, we want to ensure that pycytominer does not silently exclude data without the user being aware.

Therefore we should always include it and users can drop if need be.

Issue #80, opened 2020-05-15
Update `aggregate` count cells 


gwaygenomics:
in broadinstitute/lincs-cell-painting#34 @shntnu notes:

> Also, I think it will be cleaner for pycytominer to count cells by summing the Count_Cells column in the Image.csv files for that well (see sample Image.csv file) and then report that column as Count_Cells in the <plate>_cell_count.csv file (cell count bingo!). But I don't think this is a big deal and we can leave it as is.

We should update pycytominer `AggregateProfiles().count_cells()` to use this column as default. We should keep current behavior as secondary if for some reason that column name changes or is not present in future or past data.

This exercise has a relatively low barrier of entry and represents a small chunk of updates. Therefore, it is good for a first time contributor!

PR #671, opened 2026-05-05
Make `SingleCells.count_cells()` use configurable count columns

### LINKED ISSUE #261: `count_cells` function unable to work with SQLite file without TableNumber
When trying to use the `count_cells` function within the [cells.py](https://github.com/cytomining/pycytominer/blob/b2c6cc4580cf9e1c040a7370b99976916a22e756/pycytominer/cyto_utils/cells.py#L305) file, I am coming across this error:
```
OperationalError: (sqlite3.OperationalError) no such table: cells
[SQL: select TableNumber, ImageNumber, ObjectNumber from cells]
(Background on this error at: https://sqlalche.me/e/14/e3q8)
```
I noticed that within this function, the `query_cols` are hard-coded to be `query_cols = "TableNumber, ImageNumber, ObjectNumber"`. Since my SQLite file output from CellProfiler doesn't contain `TableNumber` (as I don't use groups in my pipeline), I cannot use this function.

I am not sure if it would be best to make the `query_cols` parameters. What would be the best way to make this function more flexible?

axiomcura:
In this PR, we replace hard-coded column references in `SingleCells.count_cells()` with configurable parameters. This makes cell counting more flexible for SQLite files that do not use the default CellProfiler column names.

**Changes**
- Adds configurable parameters for cell-counting columns in `SingleCells.count_cells()`.
- Allows callers to specify custom `merge_cols`, `object_col`, and `image_count_col`.
- Updates cell-count logic to use image-level count columns when available.
- Falls back to object-level counting when an image-level count column is not available.
- Adds test coverage for custom image count columns and custom merge columns.

This was done due to some  CellProfiler exports or downstream SQLite schemas may not include the default hard-coded columns such as `TableNumber`, `ObjectNumber`, or `Count_Cells`. This update lets users count cells correctly across alternate schemas without modifying their input data.

# Description

This closes #261  and #80 

# Checklist

### DIFF
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -329,9 +329,13 @@ class SingleCells:
         compartment : str, default "cells"
             Compartment to subset.
         merge_cols : list[str], default ["TableNumber", "ImageNumber"]
-            Columns used to merge image and compartment tables for counting.
+            Columns used to merge image and compartment tables when falling back
+            to object-level counting. Must include at least one column when
+            image_count_col is unavailable.
         object_col : str, default "ObjectNumber"
-            Column used as the object identifier to count.
+            Column used as the object identifier when falling back to
+            object-level counting. Must be non-empty when image_count_col is
+            unavailable.
         image_count_col : str, default "Count_Cells"
             Image-level count column to sum by strata before falling back to
             object-level counting.
@@ -401,14 +405,16 @@ class SingleCells:
                 )
                 return count_df
 
-            # if merge_cols is empty, raise an error since we need merge columns to
-            # merge image and compartment data for counting.
+            # Fall back to object-level counting when an image-level count column
+            # is unavailable in both the loaded image data and the SQL image table.
+            # This path requires merge columns to attach image metadata to object
+            # rows before grouping by strata.
             if len(merge_cols) < 1:
                 raise ValueError("merge_cols must include at least one merge column.")
 
-            # if object_col is empty, raise an error since we need an object column to
-            # count the number of cells per strata.
-            if len(object_col) == 0:
+            # This path also requires an object identifier column to count objects
+            # per strata.
+            if not object_col:
                 raise ValueError("object_col must be a non-empty column name.")
 
             # specify query to get the columns needed for counting cells per group
--- a/pycytominer/cyto_utils/cells.py
+++ b/pycytominer/cyto_utils/cells.py
@@ -315,7 +315,11 @@ class SingleCells:
         self.image_data_loaded = True
 
     def count_cells(
-        self, compartment: str = "cells", count_subset: bool = False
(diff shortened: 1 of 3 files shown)


pycytominer/cyto_utils/cells.py
```diff
@@ -347,23 +362,69 @@ def count_cells(
...
+                    .groupby(self.strata)[image_count_col]
+                    .sum()
+                    .reset_index()
+                    .rename({image_count_col: "cell_count"}, axis="columns")
+                )
+                return count_df
+            image_table_col_names = self.get_sql_table_col_names(self.image_table_name)
+
+            # if the image-level count column is present in the image table, use it to
+            # compute cell counts per strata.
```
gwaybio:
```suggestion
            
            # if the image-level count column is present in the image table, use it to
            # compute cell counts per strata.
            image_table_col_names = self.get_sql_table_col_names(self.image_table_name)
```

Rearranging comment and adding new line characters for added clarity. Please confirm my understanding

pycytominer/cyto_utils/cells.py
```diff
@@ -347,23 +362,69 @@ def count_cells(
...
+                count_df = (
+                    count_df
+                    .groupby(self.strata)[image_count_col]
+                    .sum()
+                    .reset_index()
+                    .rename({image_count_col: "cell_count"}, axis="columns")
+                )
+                return count_df
+
+            # if merge_cols is empty, raise an error since we need merge columns to
```
gwaybio: I am unclear why this block (here to line 430) exists. Can you add a comment? In what scenario does this function actually arrive at line 404? Maybe line 371 has this info?

Depending on the answer, you may consider updating the docstring to specify how to avoid value errors that will trigger in certain? all? scenarios?

gwaybio (changes requested):
I'm requesting clarity in one comment - updated tests looks good to me.

I agree with your decision to not add deprecation warnings in this PR, instead reserving for later PR

pycytominer/cyto_utils/cells.py
axiomcura: Good point! This block is only reached when `count_cells()` cannot use an image-level count column, meaning `image_count_col` is missing from both the loaded image data and the SQL image table (@jenna-tomkinson correct me if I am wrong). In that case, the function falls back to object-level counting, which requires `merge_cols` to join object rows back to image metadata and `object_col` to count objects per strata.

I updated the docstring to clarify that `merge_cols` and `object_col` are required only for this fallback path, and I added an inline comment explaining when/why the block is reached.

I hope htis is all clear!

gwaybio: thanks Erik! This makes sense

gwaybio (approved):

axiomcura:
Cool! Now merging

Issue #511, opened 2025-02-07
`feature_select` blocklist parameter might not work for all datasets


jenna-tomkinson:
After reviewing the blocklist.txt file, I noticed that the CellProfiler features were hardcoded for specific naming conventions for the channels (e.g., `DNA`, AGP`, `RNA`).

If someone were to name their channel `Hoechst` instead of `DNA`, wouldn't that mean that this file would not work if called?

Would it make more sense to change this default blocklist to avoid the features themselves (e.g., Manders, RWC, and Nuclei Granularity)?

Is there any documentation on why these features were selected as default if you use this operation? 

I am hoping more clarity on this file will provide better guidance for processing my datasets.

axiomcura:
Closing issue, implementation done in #674

Issue #176, opened 2021-10-22
Why is the blocklisted feature list only for Nuclei_* ?


dlogan:
It seems to me that the analogous Cell_* and Cytoplasm_* features would also have the same issues.  Of course, the specific compartments would depend upon which were measured in the particular pipeline.  But in general, I doubt that across the channels, for the mostly Correlation and Granularity features, that there would be any biology specific to nuclei that wouldn't also apply to the Cytoplasm and Cell compartments. @gwaygenomics

axiomcura:
Closing issue due to #674

PR #674, opened 2026-05-08
Dynamic blocklist object implementation 

### LINKED ISSUE #544: FeatureRequest: Feature selection - pass a list of blocklist features rather than a file
### Feature type

- [x] Add new functionality

- [x] Change existing functionality

Instead of passing through a file for blocklist features add the capability to pass through a list of blocklist features.

### Feature example

pycytominer.feature_select.feature_select(profiles, features='infer', image_features=False, samples='all', operation='variance_threshold', output_file=None, output_type='csv', na_cutoff=0.05, corr_threshold=0.9, corr_method='pearson', freq_cut=0.05, unique_cut=0.01, compression_options=None, float_format=None, blocklist_file=None, outlier_cutoff=500, noise_removal_perturb_groups=None, noise_removal_stdev_cutoff=None)

change the types for blocklist_file=None to accept lists too

### Alternative Solutions

_No response_

### Additional information

_No response_

axiomcura:
In this PR, we fully remove the `blocklist_file` parameter and replace it with direct `blocklist` support. The blocklist operation no longer reads blocklist features from legacy text/CSV files. Instead, users can pass blocklist features as a list, use a `Blocklist` object, or rely on the packaged default YAML blocklist.

The packaged blocklist registry now lives in `blocklists.yaml`. The `blocklist_type` parameter selects a top-level key from that YAML registry, such as `default`.

```python
from pycytominer import feature_select

selected = feature_select(
    profiles,
    operation="blocklist",
    blocklist=["Nuclei_Correlation_Manders_AGP_DNA", "Nuclei_Granularity_14_AGP"],
)
```

Use a packaged blocklist by name:

```python
from pycytominer import feature_select

selected = feature_select(
    profiles,
    operation="blocklist",
    blocklist_name="nuclei_corr_and_granularity",
)
```

Use a `Blocklist` object:

```python
from pycytominer.cyto_utils import Blocklist

custom_blocklist = Blocklist(features_to_block=[
    "Nuclei_Correlation_Manders_AGP_DNA",
    "Nuclei_Correlation_RWC_ER_RNA",
])

selected = feature_select(
    profiles,
    operation="blocklist",
    blocklist=custom_blocklist,
)
```

Extend a packaged blocklist:

```python
from pycytominer.cyto_utils import Blocklist

blocklist = Blocklist(
    blocklist_name="nuclei_corr_and_granularity",
    features_to_block=["Cells_Custom_Feature"],
)

selected = feature_select(
    profiles,
    operation="blocklist",
    blocklist=blocklist,
)
```

Note: calling `feature_select(..., operation="blocklist")` without `blocklist` or `blocklist_name` no longer loads a packaged blocklist.

Closes #544, #511,  #176

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Flexible blocklist inputs: accept string, list, or Blocklist object, plus a new blocklist_name option to load named packaged blocklists.
  * Packaged, YAML-backed named blocklist registry with a sensible default and combined blocking behavior.

* **Chores**
  * Added runtime YAML support for packaged blocklists; removed the old standalone text-based artifact.

* **Tests**
  * Expanded tests for Blocklist construction, named registries, input normalization, combination rules, and dataframe filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cytomining/pycytominer/pull/674?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
 import pathlib
+from collections.abc import Sequence
 from typing import Optional, Union
 
 import pandas as pd
 
-blocklist_file = os.path.join(
-    os.path.dirname(__file__), "..", "data", "blocklist_features.txt"
+blocklists_file = os.path.join(
+    os.path.dirname(__file__), "..", "data", "blocklists.json"
```
gwaybio: I'm wondering why we've chosen to migrate to json? I understand move away from txt, but why not a yaml?

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+        features : sequence of str, optional
+            Additional feature names to add to the blocklist.
+        blocklists_file : path-like object, default packaged blocklists.json
+            JSON file mapping blocklist names to feature lists.
+        """
+        self.type = type
+        self.features: list[str] = []
+
+        if type is not None:
+            self.features.extend(_load_named_blocklist(type, blocklists_file))
```
gwaybio: hmm, this seems confusing/problematic to me. Let me see if I understand. If type is anything except none, then this will add the existing blocklist to the features we want to drop? This type of behavior is likely to be unexpected from a user.

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+
+        if features is not None:
+            self.add(features)
+
+    def add(self, features: Union[str, Sequence[str]]) -> None:
+        """Add one or more feature names to the blocklist."""
+        features_to_add = [features] if isinstance(features, str) else list(features)
+        self.features = list(dict.fromkeys([*self.features, *features_to_add]))
```
gwaybio: hmm, this seems overly complicated and unpythonic. Why not enforce `self.features` is always a list then do?

```python
self.features = self.features + features_to_add
```

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+
+    def __iter__(self):
+        return iter(self.features)
+
+    def __len__(self):
+        return len(self.features)
+
+
+def _load_named_blocklist(
+    blocklist_type: str,
```
gwaybio: can you add docstrings? I'm wondering what `blocklist_type` is. I think I know but am not sure. Is this the json key?

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+
+    if blocklist_type not in blocklists:
+        blocklist_names = ", ".join(sorted(blocklists))
+        raise ValueError(
+            f"Unknown blocklist type '{blocklist_type}'. Choose one of: {blocklist_names}"
+        )
+
+    blocklist = blocklists[blocklist_type]
+    if not isinstance(blocklist, list) or not all(
+        isinstance(feature, str) for feature in blocklist
```
gwaybio: I'm wondering why we need to check that they are all strings? Will python convert them under the hood? This has the potential to be frustrating for a user as well - this might make them continuously edit the json file, especially if some autoconversion is happening.

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+    ):
+        raise ValueError("Blocklist registry entries must be lists of feature names.")
```
gwaybio: this value error also should explicitly describe the `str` entry requirement, otherwise users would need to look at source code (which is fine, but a more direct answer would be better)

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
 
     Parameters
     ----------
-    blocklist_file : path-like object
-        Location of the dataframe with features to exclude.
+    blocklist : str, sequence of str, or Blocklist, optional
+        Feature name(s) to exclude. If None, load the named blocklist from the
+        packaged registry.
+    blocklist_type : str, default "default"
+        Name of the packaged blocklist to load when ``blocklist`` is None.
```
gwaybio: If this is the json key, then this needs to be clearer in the dostring

pycytominer/cyto_utils/features.py
```diff
@@ -32,12 +108,15 @@ def get_blocklist_features(
...
-    if not any(x == "blocklist" for x in blocklist.columns):
-        raise ValueError("one column must be named 'blocklist'")
+    if blocklist is None:
+        blocklist_features = Blocklist(type=blocklist_type).to_list()
+    elif isinstance(blocklist, Blocklist):
+        blocklist_features = blocklist.to_list()
+    elif isinstance(blocklist, str):
+        blocklist_features = [blocklist]
+    else:
+        blocklist_features = list(blocklist)
```
gwaybio: Aren't these doing the same thing?

pycytominer/cyto_utils/features.py
```diff
@@ -2,27 +2,103 @@
...
+default_blocklist_name = "default"
+
+
+class Blocklist:
+    """Container for named and user-provided blocklist features."""
+
+    def __init__(
+        self,
+        type: Optional[str] = None,  # noqa: A002
+        features: Optional[Sequence[str]] = None,
```
gwaybio: why is it necessary to specify Sequence here? I don't think the order of the features matters.

pycytominer/cyto_utils/features.py
```diff
@@ -32,12 +108,15 @@ def get_blocklist_features(
         Features to exclude from downstream analysis.
     """
 
-    blocklist = pd.read_csv(blocklist_file)
-
-    if not any(x == "blocklist" for x in blocklist.columns):
-        raise ValueError("one column must be named 'blocklist'")
+    if blocklist is None:
+        blocklist_features = Blocklist(type=blocklist_type).to_list()
```
gwaybio: but what if a user wants to use a custom blocklist? Shouldn't this function allow for a custom `blocklist_file` parameter in this call? I believe, if I'm understanding correctly, for a user to use a custom blocklist, they need to find and edit the JSON in the package location?

tests/test_cyto_utils/test_feature_blocklist.py
```diff
@@ -24,3 +26,43 @@ def test_blocklist():
...
+    assert blocklist_from_object.to_list() == []
+
+
+def test_named_blocklist():
+    blocklist_from_object = Blocklist(type="default")
+    assert blocklist == blocklist_from_object.to_list()
+
+
+def test_named_blocklist_additional_features():
+    blocklist_from_object = Blocklist(type="default", features=["Cells_Custom"])
```
gwaybio: again, I'm not sure this should be the default behavior. Should it? Do we _always_ want to remove the `blocklist.json` features?

tests/test_cyto_utils/test_blocklist.py
```diff
@@ -24,3 +26,43 @@ def test_blocklist():
...
+
+
+def test_named_blocklist_additional_features():
+    blocklist_from_object = Blocklist(type="default", features=["Cells_Custom"])
+    assert blocklist_from_object.to_list() == [*blocklist, "Cells_Custom"]
+
+
+def test_blocklist_add_features():
+    blocklist_from_object = Blocklist(features=["Cells_Custom"])
+    blocklist_from_object.add(["Cells_Custom", "Nuclei_Custom"])
```
gwaybio, line 176: when is this method useful? I agree it could be broadly, but I can't think of a specific use case.

gwaybio (changes requested):

pycytominer/cyto_utils/features.py
axiomcura: Good point. I'm just used to JSON, but this made me think about how contributors and others should interact with the YAML file, and it's much easier. It's less verbose and requires fewer characters when typing. I have chaged it to YAML now!

pycytominer/cyto_utils/features.py
axiomcura: Good point. I updated the Blocklist API to make the additive behavior explicit: `blocklist_name` now selects  blocklist file, and `extra_features` appends user-provided features to it. This avoids the confusing type parameter and makes it clearer that custom features are additional drops, not a replacement.

pycytominer/cyto_utils/features.py
axiomcura: updated!

pycytominer/cyto_utils/features.py
axiomcura: Doc strings added!

pycytominer/cyto_utils/features.py
axiomcura: Updated, and I don't anticipate this being an issue as PyYAML automatically reads these as strings.

pycytominer/cyto_utils/features.py
axiomcura: Done!

pycytominer/cyto_utils/features.py
axiomcura: removed!

pycytominer/cyto_utils/features.py
axiomcura: Updated the blocklist parameter to only accept `list[str]` instead of a plain `str`. Now looking at this again... accepting a bare string didn't make  sense here, so that option has been removed along with the associated handling code.

tests/test_cyto_utils/test_feature_blocklist.py
axiomcura: This will no longer be the default. Users must now specify a blocklist group if they want to use one.

tests/test_cyto_utils/test_blocklist.py
axiomcura: This is to test  how flexible the object is. For example, you can instantiate a `Blocklist` object with features already set to block. However, there may be cases where you want to add more features later in the workflow. While this may be a less common use case, it can still be useful. Instead of reinstantiating the entire object, you can dynamically update the blocklist by adding additional features.

axiomcura:
@gwaybio I was able to attend all comments. ready for your review again!

pycytominer/cyto_utils/features.py
```diff
@@ -7,22 +7,134 @@
...
+            "strings."
+        )
+
+    return [str(feature) for feature in blocklist]
 
 
 def get_blocklist_features(
-    blocklist_file: Union[str, pathlib.Path] = blocklist_file,
+    blocklist: Optional[Union[list[str], Blocklist]] = None,
+    blocklist_name: Optional[str] = None,
```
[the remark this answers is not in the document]
gwaybio: let's add this @axiomcura

pycytominer/cyto_utils/features.py
```diff
@@ -32,12 +144,17 @@ def get_blocklist_features(
...
     """
 
-    blocklist = pd.read_csv(blocklist_file)
-
-    if not any(x == "blocklist" for x in blocklist.columns):
-        raise ValueError("one column must be named 'blocklist'")
+    if blocklist is None and blocklist_name is not None:
+        blocklist_features = Blocklist(blocklist_name=blocklist_name).to_list()
+    elif blocklist is None:
+        blocklist_features = []
```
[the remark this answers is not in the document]
gwaybio: let's make this fix too! Should we add a quick test of this as well?

pycytominer/data/blocklists.yaml
```diff
@@ -0,0 +1,56 @@
+nuclei_corr_and_granularity:
```
gwaybio:
```suggestion
default:
```

tests/test_cyto_utils/test_blocklist.py
gwaybio: could it, for example, be used if someone wants, in the future, to use two named blocklists in a single yaml file?

gwaybio (approved):
Nice work @axiomcura - looks good. Let's incorporate the coderabbit suggestions. I also had a few other discussion items.

tests/test_cyto_utils/test_blocklist.py
axiomcura: functionality added and test added!

axiomcura:
@gwaybio Ready for your review!

pycytominer/feature_select.py
```diff
@@ -39,7 +40,8 @@ def feature_select(
     unique_cut: float = 0.01,
     compression_options: Optional[Union[str, dict[str, Any]]] = None,
     float_format: Optional[str] = None,
-    blocklist_file: Optional[str] = None,
+    blocklist: Optional[Union[str, list[str], Blocklist]] = None,
+    blocklist_name: Optional[Union[str, list[str]]] = None,
```
gwaybio, line 44: can we make these default to the default blocklist name? If someone specifies a "blocklist" operation, they should expect default behavior.

gwaybio (approved):
one change, then I'll merge (once tests pass)!

pycytominer/feature_select.py
gwaybio: looking into now - might be simple

gwaybio: alright, this was a bit more complicated than I initially thought, see 36452c0934aa8faa2abea70693c3ea20de9b550f

Without fixes in that commit, a user of pycytominer 1.5, applying `feature_select(profiles_df, operation="blocklist")` would be shocked to learn that it no longer dropped the default blocklist.

gwaybio:
@d33bs - can you take a look at this PR? I believe it is ready to merge, but would be good to get a second set of eyes on it, particularly my changes in 36452c0934aa8faa2abea70693c3ea20de9b550f

gwaybio:
My apologies if you had a review in flight @d33bs , but I added some important changes in ad415dcb55cbeb2cf581c88f65dc90ea692996a6

Specifically, I refactored the `features.py` file to move functionality into a blocklist-specific cyto_util file. This update also changed testing files. I also increased docs in the Blocklist class, so an advanced pycytominer user can now understand how to use custom Blocklists

pycytominer/cyto_utils/blocklist.py
```diff
@@ -0,0 +1,296 @@
...
+        return self.features.copy()
+
+    def __iter__(self):
+        return iter(self.features)
+
+    def __len__(self):
+        return len(self.features)
+
+
+def _load_named_blocklist(
```
d33bs: Would it make sense to make these private functions methods of the Blocklist class? I wasn't sure if they'd get used elsewhere, so collapsing the logic into the class might make sense (as perhaps staticmethods). This might help tie the functionality more closely to the class rather than creating a more loose private function.

pycytominer/cyto_utils/blocklist.py
```diff
@@ -0,0 +1,296 @@
...
+        return []
+    if isinstance(blocklist_name, str):
+        return [blocklist_name]
+    if isinstance(blocklist_name, list):
+        return [str(name) for name in blocklist_name]
+
+    raise TypeError("blocklist_name must be a string, a list of strings, or None.")
+
+
+def get_blocklist_features(
```
d33bs, line 237: Same here, maybe this could be a staticmethod (or regular method) used within the constructor of the Blocklist class. Since we're overhauling the work anyways and plan on a v2 release (implying breaking changes) I feel we could nest this logic somehow under the new class and reduce the public class/object reference number (number of objects/classes available from the base of the project).

pycytominer/cyto_utils/blocklist.py
```diff
@@ -0,0 +1,296 @@
...
+from typing import Optional, Union
+
+import pandas as pd
+import yaml
+
+blocklists_file = os.path.join(
+    os.path.dirname(__file__), "..", "data", "blocklists.yaml"
+)
+
+default_blocklist_name = "default"
```
d33bs: Consider making variable labels here all capital labels for global use (earmarking them as such anytime they're implemented).

pycytominer/data/default_blocklists.yaml
```diff
@@ -0,0 +1,56 @@
+default:
```
d33bs, line 4: Consider adding a comment to the yaml file indicating what it does and where the data was sourced (perhaps before the txt data also if possible). Huzzah for more descriptive data storage formats like yaml and json 🤩!

pycytominer/data/default_blocklists.yaml
d33bs, line 1: This is a nit, but important for data sustainability; consider indicating "what kind" of blocklist this is within the filename (so, `pycytominer_default_blocklists.yaml` or `default_blocklists.yml`).

pycytominer/feature_select.py
```diff
@@ -39,7 +40,8 @@ def feature_select(
     unique_cut: float = 0.01,
     compression_options: Optional[Union[str, dict[str, Any]]] = None,
     float_format: Optional[str] = None,
-    blocklist_file: Optional[str] = None,
```
d33bs, line 42: What if we redirected the use of this parameter to the new arguments with a warning / deprecation notice, making a "best effort" to load things as we used to? It feels like this is a short distance to accomplish and could be a kindness we pass to users without a more stark breaking change as-is. This might imply reviving the blocklist_features.txt file and testing that we still have compatibility there.

d33bs (approved):
Great stuff @axiomcura and @gwaybio ! I left a few comments for consideration and overall felt this LGTM when you think it's ready.

tests/test_feature_select.py
d33bs, line 1: Double checking: do we also have CLI tests for this new capability? Could be important, given how CLI data are received.

pycytominer/cyto_utils/blocklist.py
gwaybio: i agree! thanks for the suggestion - added in 7322239ad53f1a45074ab12e5a3d4de828a2cc5b

pycytominer/cyto_utils/blocklist.py
gwaybio: this is an interesting suggestion, but I think keeping it distinct from the Blocklist class is actually a bit simpler for a user/reader of the function. While the "blocklist" operation in `feature_select` is the only customer of this function, users can still call it without class instantiation. Also, its name and scope are well defined within `blocklist.py`. Let's leave unchanged for now!

pycytominer/cyto_utils/blocklist.py
gwaybio: yep, good point - added in 3fbea8bc4602f68c125da950e9fee1ba94cdfdf3

pycytominer/data/default_blocklists.yaml
gwaybio: added in 6e5534198875c68d5bd52ab1491da7928d827579 - thanks!

pycytominer/data/default_blocklists.yaml
gwaybio: good point, added in 6e5534198875c68d5bd52ab1491da7928d827579

tests/test_feature_select.py
gwaybio: we do not! Probably should, thanks for the suggestion - added in e264eca89e82cfc5b6c0ecc0c0d200fc782f2a2d

pycytominer/feature_select.py
gwaybio: thanks for this thought @d33bs - i agree that we should aim for kindness on all pycytominer enhancements, and, with blocklist being the first such enhancement, we should set precedent. Luckily, claude was able to elegantly make this change! :D See 4467b2701abc3a7fcbcfba2c6ccf2e8df74d8aa5

pycytominer/feature_select.py
gwaybio: even more complicated than the first pass through! it would be best for us to add deprecation warnings in 1.5 to accommodate our users who might be using custom blocklist files. See 4467b2701abc3a7fcbcfba2c6ccf2e8df74d8aa5

gwaybio:
thanks @d33bs and @axiomcura - all tests pass, merging this in now!

PR #682, opened 2026-05-27
Add drop qc rows parameter in normalize with test

jenna-tomkinson:
# Description

In this PR, a new parameter called `drop_qc_rows` is added to the normalize function to find columns from coSMicQC with `Metadata_cqc` prefix and drop rows that failed QC. The expectation is that a user would have annotated their profiles with the coSMicQC `qc.parquet` and would not want to normalize their data using those failed cells.

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Add option to remove QC‑flagged rows before normalization so flagged samples are excluded from scaling; if QC flags are requested but none are present, an error is raised.

* **Tests**
  * Added tests verifying QC‑flagged rows are filtered (including multiple QC columns), remaining samples are standardized correctly, and appropriate errors occur when QC columns are missing or all rows would be dropped.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cytomining/pycytominer/pull/682?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pycytominer/normalize.py
+++ b/pycytominer/normalize.py
@@ -22,7 +22,7 @@ def normalize(
     meta_features: Union[str, list[str]] = "infer",
     samples: str = "all",
     method: str = "standardize",
-    drop_qc_rows: bool = False,
+    drop_cosmicqc_rows: bool = False,
     output_file: Optional[str] = None,
     output_type: Optional[
         Literal["csv", "parquet", "anndata_h5ad", "anndata_zarr"]
@@ -71,9 +71,10 @@ def normalize(
     method : str
         How to normalize the dataframe. Defaults to "standardize". Check avail_methods
         for available normalization methods.
-    drop_qc_rows : bool
+    drop_cosmicqc_rows : bool
         Whether to drop rows that are flagged as QC failures. The function looks for columns
         from coSMicQC with "Metadata_cqc_" prefix and drop rows with True. Defaults to False.
+        Suggested use after a prior call to `pycytominer.annotate(external_metadata=qc.parquet)`.
     output_file : str, optional
         If provided, will write normalized profiles to file. If not specified, will
         return the normalized profiles as output. We recommend that this output file be
@@ -161,8 +162,8 @@ def normalize(
     # Load Data
     profiles = load_profiles(profiles)
 
-    # If drop_qc_rows is True, drop rows that are flagged (True) as QC failures from coSMicQC
-    if drop_qc_rows:
+    # If drop_cosmicqc_rows is True, drop rows that are flagged (True) as QC failures from coSMicQC
+    if drop_cosmicqc_rows:
         qc_columns = [
             col for col in profiles.columns if col.startswith("Metadata_cqc_")
         ]
--- a/pycytominer/normalize.py
+++ b/pycytominer/normalize.py
@@ -169,6 +169,10 @@ def normalize(
         if qc_columns:
             passed_qc = ~profiles[qc_columns].any(axis="columns")
             profiles = profiles[passed_qc]
+            if profiles.empty:
+                raise ValueError(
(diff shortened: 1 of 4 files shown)


pycytominer/normalize.py
```diff
@@ -157,6 +161,17 @@ def normalize(
...
+    # If drop_qc_rows is True, drop rows that are flagged (True) as QC failures from coSMicQC
+    if drop_qc_rows:
+        qc_columns = [col for col in profiles.columns if col.startswith("Metadata_cqc_")]
+        if qc_columns:
+            qc_filter = np.logical_not(profiles[qc_columns].any(axis="columns"))
+            profiles = profiles[qc_filter]
+        else:
+            raise ValueError(
+                "No QC columns found with prefix 'Metadata_cqc_'. Cannot drop QC rows."
+            )
```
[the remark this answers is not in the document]
jenna-tomkinson: Agreed this is a good call to add a clear error here instead of a potential vague one. This technically should not happen but better to have it in case. I have added it in commit [1fc6e15](https://github.com/cytomining/pycytominer/pull/682/commits/1fc6e1534d46749520350bbde256339036db5a93)

pycytominer/normalize.py
```diff
@@ -70,6 +71,9 @@ def normalize(
     method : str
         How to normalize the dataframe. Defaults to "standardize". Check avail_methods
         for available normalization methods.
+    drop_qc_rows : bool
```
gwaybio:
```suggestion
    drop_cosmicqc_rows : bool
```

i think it would be helpful to specify in the variable name that this function is specific for cosmicQC

pycytominer/normalize.py
```diff
@@ -70,6 +71,9 @@ def normalize(
...
+        Whether to drop rows that are flagged as QC failures. The function looks for columns
+        from coSMicQC with "Metadata_cqc_" prefix and drop rows with True. Defaults to False.
```
gwaybio, line 76:
```suggestion
        from coSMicQC with "Metadata_cqc_" prefix and drop rows with True. Defaults to False.
        Suggested use after a prior call to `pycytominer.annotate(external_metadata=qc.parquet)`.
```

tests/test_normalize.py
```diff
@@ -430,6 +430,38 @@ def test_normalize_standardize_ctrlsamples():
     pd.testing.assert_frame_equal(normalize_result, expected_result)
 
 
+def test_normalize_drop_qc_rows_before_normalization():
```
gwaybio: I'd like to see three more tests:

1. When there are two columns with `Metadata_cqc` prefix and they don't agree (testing that it drops any instance of True in mutiple columns)
2. When someone sets drop_qc_rows=True and there are no Metadata_cqc columns.
3. When all rows have a metadata_cqc column that is true (dropping all rows)

gwaybio (changes requested):
Thanks Jenna! Looking good. A few comments/suggestions.

Can you also confirm the CLI captures this functionality?

jenna-tomkinson:
@gwaybio Good catch on the CLI, I will make sure to check that! Thanks!

pycytominer/normalize.py
jenna-tomkinson: Good call, I will update the name to be more explicit.

pycytominer/normalize.py
jenna-tomkinson: Added!

tests/test_normalize.py
jenna-tomkinson: Added all three new tests and they are passing! Thanks for these suggestions!

jenna-tomkinson:
@gwaybio Back to you! 🏈

gwaybio (approved):
Looks great Jenna, merging in! 🎉

PR #687, opened 2026-06-03
fix modz bug

### LINKED ISSUE #162: MODZ drops replicate groups containing missing values

axiomcura:
# Description
in this PR, `modz` is updated so replicate groups with missing values in `replicate_columns` are no longer silently dropped, and MODZ handles edge cases where correlation-based weights become undefined.

The changes are intentionally small and targeted:
- `modz()` now uses `groupby(..., dropna=False)` so groups containing `NA` metadata are preserved during consensus generation.

This PR also adds tests in `tests/test_cyto_utils/test_modz.py` for:
- replicate groups containing `NA` values
- profiles with zero-valued features that previously caused invalid weighting behavior

No existing tests were changed, and the overall structure of the MODZ implementation remains the same to keep the fix small and ready for review.

This PR closes #162 and #170

# Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved weight normalization with safeguards against division-by-zero errors
  * Enhanced handling of missing data in replicate groups—now retained instead of excluded

* **Tests**
  * Added test coverage for NaN values in replicate group preservation
  * Added test coverage for zero-value profiles in consensus computation

### DIFF
--- a/pycytominer/cyto_utils/modz.py
+++ b/pycytominer/cyto_utils/modz.py
@@ -73,13 +73,17 @@ def modz_base(
     cor_df = cor_df.clip(lower=0)
 
     # Get average correlation for each profile (will ignore NaN)
-    raw_weights = cor_df.mean(axis=1)
+    raw_weights = cor_df.mean(axis=1).fillna(min_weight)
 
     # Threshold weights (any value < min_weight will become min_weight)
     raw_weights = raw_weights.clip(lower=min_weight)
 
     # normalize raw_weights so that they add to 1
-    weights = raw_weights / sum(raw_weights)
+    weight_sum = raw_weights.sum()
+    if weight_sum == 0:
+        weights = pd.Series(1 / len(raw_weights), index=raw_weights.index)
+    else:
+        weights = raw_weights / weight_sum
     weights = weights.round(precision)
 
     # Step 3: Normalize
@@ -151,7 +155,7 @@ def modz(
 
     modz_df = (
         population_df
-        .groupby(replicate_columns)
+        .groupby(replicate_columns, dropna=False)
         .apply(
             lambda x: modz_base(
                 x.loc[:, features],
--- a/pycytominer/cyto_utils/modz.py
+++ b/pycytominer/cyto_utils/modz.py
@@ -73,14 +73,16 @@ def modz_base(
     cor_df = cor_df.clip(lower=0)
 
     # Get average correlation for each profile (will ignore NaN)
-    raw_weights = cor_df.mean(axis=1).fillna(min_weight)
+    raw_weights = cor_df.mean(axis=1)
 
     # Threshold weights (any value < min_weight will become min_weight)
     raw_weights = raw_weights.clip(lower=min_weight)
 
-    # normalize raw_weights so that they add to 1
+    # Normalize raw weights so the weighted sum remains on the original feature scale.
     weight_sum = raw_weights.sum()
     if weight_sum == 0:
+        # If all profiles have zero weight, use equal weights to avoid division by zero
+        # and still produce a valid consensus profile.
         weights = pd.Series(1 / len(raw_weights), index=raw_weights.index)
     else:
         weights = raw_weights / weight_sum
--- a/tests/test_cyto_utils/test_modz.py
+++ b/tests/test_cyto_utils/test_modz.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pandas as pd
 
 from pycytominer.cyto_utils import modz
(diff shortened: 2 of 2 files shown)


pycytominer/cyto_utils/modz.py
```diff
@@ -73,13 +73,17 @@ def modz_base(
     cor_df = cor_df.clip(lower=0)
 
     # Get average correlation for each profile (will ignore NaN)
-    raw_weights = cor_df.mean(axis=1)
+    raw_weights = cor_df.mean(axis=1).fillna(min_weight)
```
d33bs: This feels like a larger implied change. Consider providing documentation to help ground this decision somehow.

pycytominer/cyto_utils/modz.py
```diff
@@ -73,13 +73,17 @@ def modz_base(
...
 
     # Threshold weights (any value < min_weight will become min_weight)
     raw_weights = raw_weights.clip(lower=min_weight)
 
     # normalize raw_weights so that they add to 1
-    weights = raw_weights / sum(raw_weights)
+    weight_sum = raw_weights.sum()
+    if weight_sum == 0:
```
d33bs, line 83: Do we test for negative cases too somehow?

tests/test_cyto_utils/test_modz.py
```diff
@@ -191,3 +192,51 @@ def test_modz_unbalanced_sample_numbers():
         },
     )
     pd.testing.assert_frame_equal(expected_result, consensus_df)
+
+
+def test_modz_preserves_na_replicate_groups():
```
d33bs, line 197: Have we tested for multi-column replicate groups?

d33bs, line 197: Consider providing docstrings for both this and the other test to ensure we describe what the goal of the test is.

d33bs:
Thanks @axiomcura for the changes here! I left some comments but wasn't as familiar with the functionality and history here. @gwaybio, could I ask you to take a look as well?

pycytominer/cyto_utils/modz.py
gwaybio: line 75 also needs to change - what is the justification to fill NAs with min weight?

pycytominer/cyto_utils/modz.py
gwaybio: please also add a comment here - what is this edge case? why is it important to prevent?

tests/test_cyto_utils/test_modz.py
gwaybio: agree - all tests need docs like this

gwaybio (changes requested):
A few comments, which may trigger additional comments once resolved. I also don't see a test where `min_weight` is triggered

tests/test_cyto_utils/test_modz.py
axiomcura: tests added!

axiomcura: I think there's a test that checks for this: https://github.com/cytomining/pycytominer/blob/341cbe603abb9125149802a7eae793df36b159cd/tests/test_cyto_utils/test_modz.py#L82-L110

pycytominer/cyto_utils/modz.py
axiomcura: I noticed that there wasn’t a good way to handle the case where all raw weights are `0`, which can generate a series of `NaN` weights during normalization. This can happen when all replicate correlations are non-positive and `min_weight=0`; falling back to equal weights prevents invalid consensus values while still producing a reasonable profile.

Another option is to just raise an error for this edge case were all weight are 0. 

A 3rd option is just to remove this check c: (if this edge cases is likely to never happen) 

LMK what you think!

axiomcura: also updated with comments !

pycytominer/cyto_utils/modz.py
axiomcura: This was a mistake on my end! now removed!  (why I didn't change the document) code that I forgot to remove

gwaybio (approved):

axiomcura:
@d33bs LMK if I have answered all your comments and suggestions before I merge!

axiomcura:
@d33bs Greg approved the changes, just wanted to know if you had any comments, if not, I'd be happy to merge c:

pycytominer/cyto_utils/modz.py
d33bs: What would happen for a -1 min weight / weight sum? I'm not sure of the circumstances which would make this possible, but bringing it up because it seems technically an option.

pycytominer/cyto_utils/modz.py
```diff
@@ -151,7 +157,7 @@ def modz(
 
     modz_df = (
         population_df
-        .groupby(replicate_columns)
+        .groupby(replicate_columns, dropna=False)
```
d33bs, line 160: Would it be worth including optionality here in case dropna is undesired?

d33bs (approved):
Generally looking good to me, left a couple more comments.

pycytominer/cyto_utils/modz.py
gwaybio: for this, i think better for us to be opinionated

gwaybio:
thanks @axiomcura - i'm merging this in!

PR #694, opened 2026-06-03
Add minimal install/import tests on Linux and macOS

### LINKED ISSUE #354: FeatureRequest: Minimal install and import CI/CD test
### Feature type

- [X] Add new functionality

- [ ] Change existing functionality

As suggested below we should add a minimal environment install and import test to catch potential issues before we run into them at the conda stage. 

>Thanks @kenibrewer , I think the changes you've added look great for now. Besides the changes you added, generally I wonder if a short GitHub Actions test for the installation and import of Pycytominer might avoid potential issues with the conda package testing (in advance of recipe updates). This might be overdoing it but could help avoid issues in the future.

_Originally posted by @d33bs in https://github.com/cytomining/pycytominer/pull/351#discussion_r1383860016_
            

### Feature example

_No response_

### Alternative Solutions

_No response_

### Additional information

_No response_

axiomcura:
# Description

In this PR, a lightweight CI test is added to catch package installation and import issues earlier in the development cycle.

A new GitHub Actions job was added to integration-test.ym that:
- runs on `ubuntu-24.04` and `macos-14`
- installs Pycytominer with core dependencies only using `python -m pip install .`
- verifies that `pycytominer` and `pycytominer.cyto_utils` can be imported successfully

This adds an early packaging check without expanding the existing pytest matrix or requiring broader code changes.

Closes #354 

# Checklist

## Summary by CodeRabbit

* **Tests**
  * Added tests to validate core functionality in minimal installation environments, including module imports, CLI operations, and feature selection with blocklists.

* **Chores**
  * Updated development tool dependencies to latest stable versions.


.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
...
+    if: github.event_name != 'schedule'
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v6
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.TARGET_PYTHON_VERSION }}
+      - name: Install pycytominer with core dependencies only
+        run: python -m pip install .
```
d33bs: Instead of installing from source here, consider building the package and then installing from the build in isolation to help simulate what will happen when we deploy the package in the wild. This will mean that we get closer to a "real life" vision of what will happen for users. We may want to isolate the build using a single version of Python to keep things consistent with our build process (see the pypi deployment process for a version check here) instead of trying to validate the build across all envs.

.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
...
+      - name: Import pycytominer in a minimal environment
+        run: python -c "import pycytominer; import pycytominer.cyto_utils"
```
d33bs: In addition to this import, is there a simple and fast test script we could run to safely say: "this will work for most people."? Perhaps a notebook or one of the tests can provide an example. This need not be exhaustive but I feel we can do just a tad more to make failures more observable than an import.

.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
   UV_CACHE_DIR: ${{ github.workspace }}/.cache/uv
 
 jobs:
+  minimal-install-test:
```
d33bs: Consider making pre-commit checks a requirement before running this.

.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
...
+    name: Minimal install test
+    strategy:
+      fail-fast: false
```
d33bs: I feel that fail fast could be helpful here.

.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
   UV_CACHE_DIR: ${{ github.workspace }}/.cache/uv
 
 jobs:
+  minimal-install-test:
+    name: Minimal install test
```
d33bs: What would you think of appending these checks to the end of the existing pytest test job so that we don't increase our ci job listing too much? It would likely mean minimal increase of our ci job run times but from a human perspective we'd have `os * py-version` less jobs to look through within GH.

d33bs (changes requested):

.github/workflows/integration-test.yml
```diff
@@ -27,6 +27,27 @@ env:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04
+          - macos-14
```
d33bs: It could be important to add windows here too, given #698

.github/workflows/integration-test.yml
axiomcura: Done !

.github/workflows/integration-test.yml
axiomcura: Done!

.github/workflows/integration-test.yml
axiomcura: Done! Switched it to`true`

.github/workflows/integration-test.yml
axiomcura: Ah, good point! Added!

.github/workflows/integration-test.yml
axiomcura: That sounds like a good approach. I will add these checks into the existing pytest suite rather than running them as a separate CI job to help keep the GitHub Actions interface simple and cleaner

.github/workflows/integration-test.yml
axiomcura: Now that this implementation is covered by pytest, I believe that once the PR is merged, it will automatically be tested on Windows as well.

.github/workflows/integration-test.yml
axiomcura: I like this idea, changes are found here [e17128c](https://github.com/cytomining/pycytominer/pull/694/commits/e17128cea877c2e85c9c8c05147d3b16d99378ae) and here [dd0dc44](https://github.com/cytomining/pycytominer/pull/694/commits/dd0dc44d3000b1b07d12c9b1c9629793b3c6ac81)

Lmk what you think

axiomcura:
@d33bs , I have attended all your comments. Ready for your eyes again!

tests/minimal_wheel_build_tests/test_minimal_wheel_build.py
d33bs, line 1: Cool approach!

.github/workflows/integration-test.yml
d33bs: I think we're still installing from source and then running the new pytest tests. Consider using the built package, that is the output from `uv build` so that we can ensure we're simulating what the user will experience when they download and install and run the work on their systems. Given the pytest approach, you could skip those tests at first, and then isolate to that layer later with a package install.

d33bs (changes requested):
Thanks @axiomcura ! I think we're still using the source-based code to run the tests (see comment). Nice approach generally!

.github/workflows/integration-test.yml
axiomcura: Hey @d33bs, I updated this so the minimal install tests no longer run from the source-installed pytest environment.

The regular integration pytest job now skips `tests/minimal_tests`, and I added a separate `minimal-install-test` job that waits for `uv build`, downloads the built wheel artifact, installs that wheel into a clean venv, and runs only the minimal tests from a temp directory. 

I think this better simulates the user path of downloading/installing the package and then using it.

Lmk what your thoughts are. 

commit: https://github.com/cytomining/pycytominer/pull/694/commits/04bb81d2a6622820ea80ca019c24453d49051f7d

axiomcura:
All tests are passing @d33bs, ready for your review!

.github/workflows/integration-test.yml
```diff
@@ -209,6 +209,75 @@ jobs:
...
+    # Runs tests that validate the built wheel in an isolated install environment.
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-14
+          - ubuntu-24.04
+          - windows-2022
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
d33bs: Consider picking one version here or move the rest of this job to a deeper version of the pytest job. As-is, this would mean we'd have 5*3 more jobs to look through and run per change, which is a lot.

d33bs (changes requested):
Thanks for addressing things Erik! I think we need to take care for the devex here (developer experience) and either reduce the number of jobs or make this a deeper addition to the pytest job. Open to your thoughts here too.

.github/workflows/integration-test.yml
axiomcura: Good point. I agreed with the devex concern, so I changed this to avoid adding a separate 3 x 5 job matrix.

I folded the minimal-install check into the existing pytest job and gated it to run only on the existing ubuntu-24.04 + target Python matrix entry. This removes Windows-specific minimal-install coverage, but the regular Windows `pytest` jobs are still covered by the main test matrix, so we avoid the extra 15 jobs per change while keeping Windows in CI.

LMK what you think! 

Here's the commit: 
[b0817c0](https://github.com/cytomining/pycytominer/pull/694/commits/b0817c0feeccb4659ae35d0696675f9e295d8b82)

axiomcura:
Hey @d33bs I have updated the changes, LMK what you think!

.github/workflows/integration-test.yml
```diff
@@ -146,7 +148,7 @@ jobs:
       - name: Run pytest and generate coverage report
         # For the target version and ubuntu, run pytest and generate coverage report
         if: (matrix.os == 'ubuntu-24.04')  && (matrix.python-version == env.TARGET_PYTHON_VERSION)
-        run: uv run pytest --cov=./ --cov-report=xml ${{ github.event.inputs.pytest_addopts }}
+        run: uv run pytest --ignore=tests/minimal_tests --cov=./ --cov-report=xml ${{ github.event.inputs.pytest_addopts }}
```
gwaybio: why wouldn't we simply include the minimal tests within the existing pytest infrastructure?

.github/workflows/integration-test.yml
```diff
@@ -167,7 +169,37 @@ jobs:
       - name: Run pytest
         # For every other version and/or OS, run pytest without coverage
         if: (matrix.os != 'ubuntu-24.04') || (matrix.python-version != env.TARGET_PYTHON_VERSION )
-        run: uv run pytest ${{ github.event.inputs.pytest_addopts }}
+        run: uv run pytest --ignore=tests/minimal_tests ${{ github.event.inputs.pytest_addopts }}
```
gwaybio: from a devex perspective, this looks quite challenging! See my comment above - why can't we simply run the tests within the existing pytest convention and drop this?

tests/minimal_tests/conftest.py
```diff
@@ -0,0 +1,36 @@
+"""
```
gwaybio: there's no assertions in this file, so no actual tests occurring?

tests/minimal_wheel_build_tests/test_minimal_wheel_build.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert pycytominer is not None
+    assert pycytominer.cyto_utils is not None
+
+
+def test_minimal_install_cli_aggregate(
+    minimal_install_profiles_file: Path,
+    tmp_path: Path,
+) -> None:
+    """Ensure the installed CLI can aggregate a tiny checked-in profiles file."""
```
gwaybio, line 32: i don't see why we need this - don't we have nearly full coverage of the CLI and aggregate? Wouldn't this "test_minimal_install" simply check that the CLI is imported?

tests/minimal_tests/test_minimal_install.py
```diff
@@ -0,0 +1,73 @@
...
+    assert result.shape[0] == 2
+    assert np.isclose(
+        result.loc[result["Metadata_Well"] == "A01", "Feature_1"].item(), 1.5
+    )
+    assert np.isclose(
+        result.loc[result["Metadata_Well"] == "A02", "Feature_2"].item(), 7.5
+    )
+
+
+def test_minimal_install_feature_select_uses_packaged_blocklist(
```
gwaybio: It's unclear to me why this test needs to exist. Don't we fully test blocklist elsewhere? Don't we fully test feature select elsewhere as well?

gwaybio (changes requested):
Much of this PR seems, from my perspective at least (and happy to be wrong about this), unnecessary. 

- We're testing feature select, blocklist, and aggregate, but these functions are covered much more extensively elsewhere
- the integration-test.yml, lines 172 to 202, seem very challenging to maintain (lots of manual directory management). It seems to me that both 172-202 and the ignore flag on line 151 could be dropped and the minimal build test would be satisfied
- Adding the new data files, because we have these covered elsewhere, just add bloat (a small amount, but some nonetheless)

Please tell me what I'm missing!

tests/minimal_tests/conftest.py
axiomcura: Good point. This file is not intended to contain tests directly; it defines pytest fixtures used by the actual tests in tests/minimal_tests/test_minimal_install.py. The assertions are in that test file, while conftest.py only provides reusable test data paths.

.github/workflows/integration-test.yml
axiomcura: We ignore `tests/minimal_tests` during the normal pytest run because those tests are meant to validate the built package, not the source checkout. They run separately after the wheel is built, in a fresh venv with only `pytest` and the wheel installed, so dev dependencies or extras do not mask missing runtime dependencies.

tests/minimal_wheel_build_tests/test_minimal_wheel_build.py
axiomcura: This is not for extra `aggregate` coverage. It checks that the built wheel works in a fresh minimal venv, where only `pytest` and the wheel are installed. Running a small CLI command helps catch missing runtime dependencies or packaging issues that source-based tests with dev dependencies could hide.

LMK your thoughts

tests/minimal_tests/test_minimal_install.py
axiomcura: This is not for extra `feature_select` or blocklist coverage. It verifies that the built wheel includes the packaged blocklist data and that `feature_select(operation="blocklist")` works in a fresh minimal install.

.github/workflows/integration-test.yml
axiomcura: Good point! I simplified this by moving the minimal install check into its own job.  I think this make it easier to read and maintain 

 However, we cannot simply run these tests in the normal pytest convention because that runs from the source checkout with dev dependencies and extras installed, which can hide packaging or missing runtime dependency issues. 

These tests need to run after the wheel is built, in a fresh venv, so imports and packaged data are validated from the installed artifact.

axiomcura:
I attended your comment @gwaybio lmk what you think!

.github/workflows/integration-test.yml
gwaybio: gotcha! Perhaps changing the name to indicate this more clearly is worth it? Something like `tests/wheel_build_tests`?

.github/workflows/integration-test.yml
gwaybio: thanks for the explanation!

tests/minimal_wheel_build_tests/test_minimal_wheel_build.py
gwaybio: thanks for the explanation! The change needed is probably with variable naming more than anything. Please see my comment above about "minimal_tests" - also, the test name should clearly indicate what _we're actually testing_

tests/minimal_tests/test_minimal_install.py
gwaybio: similar comment about "minimal_install"

Maybe `test_wheel_build_includes_the_packaged_blocklist_yml_file`?

.github/workflows/integration-test.yml
gwaybio: oh, is the ignore formulated correctly? It looks formulated for the "Run pytest and generate coverage report"?

.github/workflows/integration-test.yml
```diff
@@ -209,6 +213,43 @@ jobs:
...
+          cache-pre-commit: false
+          cache-venv: false
+          setup-uv: true
+          install-deps: false
+      - name: Download wheel artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: wheel
+          path: dist
+      - name: Run minimal install tests
```
gwaybio: also be sure to change this name to match the name of the folder you're placing these tests

tests/minimal_tests/conftest.py
```diff
@@ -0,0 +1,34 @@
+"""Shared pytest fixtures for minimal install tests."""
```
gwaybio: change name here too

gwaybio (changes requested):
i made several in-line comments - please clarify variable names then I think good to go! @d33bs should have a chance to request changes as well.

tests/minimal_tests/test_minimal_install.py
axiomcura: Done!

.github/workflows/integration-test.yml
axiomcura: Yup! this is intentional. The ignore is not specific to the coverage step. It excludes tests/minimal_tests from all regular pytest matrix runs. Those tests are then run separately in the minimal-install job against the built wheel, which is the environment they are meant to validate.

.github/workflows/integration-test.yml
axiomcura: I have updated the name!

tests/minimal_tests/conftest.py
axiomcura: Done!

.github/workflows/integration-test.yml
axiomcura: Done!

axiomcura:
I have updated the variable names in the integration-test.yml and also in the testing folder! (I hope)

.github/workflows/integration-test.yml
```diff
@@ -145,8 +147,9 @@ jobs:
         run: uv sync --frozen --all-extras --group dev
       - name: Run pytest and generate coverage report
         # For the target version and ubuntu, run pytest and generate coverage report
+        # Minimal wheel build tests run separately against the built wheel.
```
gwaybio:
```suggestion
        # Minimal wheel build tests run separately to confirm successful install.
```

.github/workflows/integration-test.yml
```diff
@@ -166,8 +169,9 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: Run pytest
         # For every other version and/or OS, run pytest without coverage
+        # Minimal wheel build tests run separately against the built wheel.
```
gwaybio:
```suggestion
        # Minimal wheel build tests run separately to confirm successful install.
```

tests/minimal_wheel_build_tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+import pytest
+
+
+@pytest.fixture(name="minimal_wheel_build_test_data_dir")
+def fixture_minimal_wheel_build_test_data_dir() -> Path:
+    """
+    Provide a data directory for minimal wheel build test data
+    """
+
+    return Path(__file__).resolve().parent.parent / "test_data" / "minimal_install"
```
gwaybio: i think this path is no longer correct - please make sure the variable naming updates are consistent

tests/minimal_wheel_build_tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.fixture(name="minimal_install_profiles_file")
```
gwaybio: here too

tests/minimal_wheel_build_tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+    minimal_wheel_build_test_data_dir: Path,
+) -> Path:
+    """
+    Provide a profiles input file for minimal wheel build tests
+    """
+
+    return minimal_wheel_build_test_data_dir / "profiles.parquet"
+
+
+@pytest.fixture(name="minimal_install_blocklist_profiles_file")
```
gwaybio: here too

gwaybio (changes requested):
Please make sure consistent variable naming language throughout (some "minimal_install" leftovers).

Also, do we need both the `.csv` and `.parquet` files for blockist and profiles? Maybe just the parquet files?

Please also make sure the integration tests work locally (i think this might fail due to variable naming errors)

tests/minimal_wheel_build_tests/conftest.py
gwaybio: please note that i'm talking about the _whole_ function (not just the line) Please confirm updated language _everywhere_

gwaybio (approved):
thanks @axiomcura LGTM

axiomcura:
@d33bs bump. PR is ready for your re-review c: ! thanks!

d33bs (approved):
LGTM!

gwaybio:
thanks @axiomcura - going to resolve comments and merge this in!

PR #697, opened 2026-06-03
Adding CLI walkthrough

gwaybio:
# Description

Adding new walkthrough to demonstrate pycytominer CLI usage.

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [x] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [ ] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [ ] My code follows the style guidelines of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have deleted all non-relevant text in this pull request template.

## Summary by CodeRabbit

* **Documentation**
  * Added an interactive Jupyter tutorial demonstrating the pycytominer CLI across all five pipeline stages with captured help/usage examples; includes a runnable end-to-end workflow on synthetic data and tips for scripting and pandas-style query strings.
* **Chores**
  * Bumped ruff pre-commit hook to a newer patch version.

PR #695, opened 2026-06-03
Added deprecation warnings for DeepProfiler_processing.py and cell_locations.py 

axiomcura:
# Description
This PR introduces deprecation warnings to the `DeepProfiler_processing.py` and `cell_locations.py` modules. These warnings are triggered when importing the modules to notify users that they will be removed in a future release of pycytominer.

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [x] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have deleted all non-relevant text in this pull request template.

## Summary by CodeRabbit

* **Chores**
  * Added deprecation warnings for DeepProfiler-related APIs to notify users they will be removed in a future release.
  * Added deprecation warnings for the legacy cell-location utility and the legacy single-cell API (advises migration to CytoTable).
  * Added deprecation warnings for legacy annotation behavior and the Broad CMAP formatting parameter.
  * Updated a pre-commit tooling hook to a newer patch version.

PR #698, opened 2026-06-03
Adding windows support

gwaybio:
# Description

Adding windows support.

# Checklist

## Summary by CodeRabbit

* **Tests**
  * Expanded integration testing to include Windows runners; added Windows-specific test skips, broadened error-message assertions, and added tests for explicit platemap separator handling.
* **Bug Fixes**
  * Resolved Windows file-handling to avoid permission errors during temporary downloads.
  * Ensured database connections/resources are always released after aggregation to prevent file locks.
  * Fixed filename parsing so plate/well/site metadata is extracted reliably across path styles.
* **New Features**
  * Added an option to specify platemap delimiter when loading/annotating platemaps.

### DIFF
--- a/pycytominer/cyto_utils/cell_locations.py
+++ b/pycytominer/cyto_utils/cell_locations.py
@@ -3,6 +3,7 @@ Utility function to augment a metadata file with X,Y locations of cells in each 
 """
 
 import collections
+import os
 import pathlib
 import tempfile
 from typing import Optional, Union
@@ -171,19 +172,23 @@ class CellLocation:
 
         bucket, key = self._parse_s3_path(uri)
 
-        with tempfile.NamedTemporaryFile(
-            delete=False, suffix=pathlib.Path(key).name
-        ) as tmp_file:
-            self.s3.download_file(bucket, key, tmp_file.name)
+        # Use mkstemp so the file descriptor is closed before boto3 touches it.
+        # NamedTemporaryFile holds an exclusive OS lock while open on Windows;
+        # s3transfer does os.remove() + rename onto the same path internally,
+        # which raises PermissionError [WinError 32] if the fd is still open.
+        fd, tmp_path = tempfile.mkstemp(suffix=pathlib.Path(key).name)
+        os.close(fd)
 
-            # Check if the downloaded file exists and has a size greater than 0
-            tmp_file_path = pathlib.Path(tmp_file.name)
-            if tmp_file_path.exists() and tmp_file_path.stat().st_size > 0:
-                return tmp_file.name
-            else:
-                raise ValueError(
-                    f"Downloaded file '{tmp_file.name}' is empty or does not exist."
-                )
+        self.s3.download_file(bucket, key, tmp_path)
+
+        # Check if the downloaded file exists and has a size greater than 0
+        tmp_file_path = pathlib.Path(tmp_path)
+        if tmp_file_path.exists() and tmp_file_path.stat().st_size > 0:
+            return tmp_path
+        else:
+            raise ValueError(
+                f"Downloaded file '{tmp_path}' is empty or does not exist."
+            )
 
     def _load_metadata(self):
         """Load the metadata into a Pandas DataFrame
--- a/pycytominer/cyto_utils/cell_locations.py
+++ b/pycytominer/cyto_utils/cell_locations.py
@@ -179,16 +179,21 @@ class CellLocation:
         fd, tmp_path = tempfile.mkstemp(suffix=pathlib.Path(key).name)
         os.close(fd)
 
-        self.s3.download_file(bucket, key, tmp_path)
+        try:
+            self.s3.download_file(bucket, key, tmp_path)
 
-        # Check if the downloaded file exists and has a size greater than 0
-        tmp_file_path = pathlib.Path(tmp_path)
-        if tmp_file_path.exists() and tmp_file_path.stat().st_size > 0:
-            return tmp_path
-        else:
+            tmp_file_path = pathlib.Path(tmp_path)
+            if tmp_file_path.exists() and tmp_file_path.stat().st_size > 0:
+                return tmp_path
+
             raise ValueError(
                 f"Downloaded file '{tmp_path}' is empty or does not exist."
             )
+        except Exception:
+            tmp_file_path = pathlib.Path(tmp_path)
+            if tmp_file_path.exists():
+                tmp_file_path.unlink()
+            raise
 
     def _load_metadata(self):
         """Load the metadata into a Pandas DataFrame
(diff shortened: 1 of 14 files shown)


pycytominer/cyto_utils/cell_locations.py
```diff
@@ -385,6 +390,11 @@ def _get_joined_image_nuclei_tables(self):
...
         joined_df = pd.read_sql_query(join_query, engine, dtype=column_types)
 
+        # Dispose the engine before removing the temp file.
+        # On Windows, SQLAlchemy's connection pool keeps the file open;
+        # unlink() would raise PermissionError unless the engine is disposed first.
+        engine.dispose()
+
         # if the single_cell file was downloaded from S3, delete the temporary file
         if temp_single_cell_input is not None:
             pathlib.Path(temp_single_cell_input).unlink()
```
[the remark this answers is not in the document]
gwaybio: thanks!

gwaybio:
ok @d33bs - tests now pass 🎉 please review when you're able. It would be great to get this (and a few other open PRs, such as #697 and #686  ) merged in and then release 1.5.2

pycytominer/cyto_utils/cell_locations.py
```diff
@@ -358,38 +368,40 @@ def _get_joined_image_nuclei_tables(self):
...
+            # check that the single_cell file has the required tables and columns
+            self._check_single_cell_correctness(engine)
+
+            image_index_str = ", ".join(self.image_key)
+
+            join_query = f"""
+            SELECT Nuclei.{self.table_column},Nuclei.{self.image_column},Nuclei.{self.object_column},Nuclei.{self.cell_x_loc},Nuclei.{self.cell_y_loc},Image.{image_index_str}
+            FROM Nuclei
+            INNER JOIN Image
+            ON Nuclei.{self.image_column} = Image.{self.image_column} and Nuclei.{self.table_column} = Image.{self.table_column};
```
d33bs: Consider adding a bit of formatting to this SQL (e.g., select cols on newlines, adding indents, etc) to help make it more readable. [This guide](https://handbook.gitlab.com/handbook/enterprise-data/platform/sql-style-guide/) is over-exhaustive for right now but might provide an example of what we can do (please don't mistake my meaning that we need to abide this as a requirement so much as a guide).

pycytominer/cyto_utils/cell_locations.py
```diff
@@ -358,38 +368,40 @@ def _get_joined_image_nuclei_tables(self):
...
+            """
+
+            column_types = {
+                self.image_column: "int64",
+                self.table_column: "int64",
+                self.object_column: "int64",
+                self.cell_x_loc: "float",
+                self.cell_y_loc: "float",
+            }
```
d33bs: Consider adding casting to the above SQL to help speed up or avoid this col-casting process from pandas (which is likely to be slower than just telling the database to send back data in the way we need). That said, I'm unsure exactly why this was added originally - it could be due to SQLite's flexibility with value types (SQLite col types are not strictly enforced on col values).

pycytominer/cyto_utils/cell_locations.py
```diff
@@ -358,38 +368,40 @@ def _get_joined_image_nuclei_tables(self):
...
+            return pd.read_sql_query(join_query, engine, dtype=column_types)
+        finally:
+            # Always dispose the engine and remove the temp file.
+            # On Windows, SQLAlchemy's connection pool keeps the SQLite file open;
+            # unlink() raises PermissionError [WinError 32] unless disposed first.
+            engine.dispose()
+            if temp_single_cell_input is not None:
+                temp_path = pathlib.Path(temp_single_cell_input)
+                if temp_path.exists():
+                    temp_path.unlink()
```
d33bs, line 403: We could probably benefit from using DuckDB instead of SQLite for any SQLite-based processing to avoid strange artifacts of SQLite's heritage (for ex. we might be able to avoid odd code like this and also enforce stricter types at the data access layer). This is probably a call-out for future work rather than something we need to address immediately within this PR.

tests/test_cyto_utils/test_load.py
```diff
@@ -146,6 +147,13 @@
 np.savez_compressed(output_npz_without_metadata_file, features=npz_feats)
 
 
+# csv.Sniffer unreliably detects tab delimiters on Windows due to known CPython bugs:
+# https://github.com/python/cpython/issues/119123
+# https://github.com/python/cpython/issues/97611
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="csv.Sniffer tab detection unreliable on Windows (cpython#119123, cpython#97611)",
```
d33bs, line 161: Yikes! Will this and the other skip have production impacts for people using Windows? Is it the csv.Sniffer in general or the error message that gave this part trouble? Perhaps we need an added dependency to help bridge into Windows land.

d33bs (changes requested):
Generally feeling good about this PR! The skipped tests made me wonder if we need further work with the csv sniffer from a Windows perspective - it felt like a sign of creakiness that might be addressable sooner than later. Let me know what you think!

pycytominer/cyto_utils/cell_locations.py
gwaybio: thanks @d33bs - i will make these formatting changes!

pycytominer/cyto_utils/cell_locations.py
gwaybio: sounds good, will add! However, to note, this to me feels right on the boundary of enhancements to cell_locations, especially given #695 - there is an argument to be made about optimizing cell locations (so the definitive usage is the version prior to deprecation), but I believe this is outside scope of this PR (adding windows support)

pycytominer/cyto_utils/cell_locations.py
gwaybio: i'll take a look at this, to see how simple with Claude - if it's easy, then we can switch here. If i run into challenges, we can reserve for future work (see my prior comment https://github.com/cytomining/pycytominer/pull/698#discussion_r3357084790)

gwaybio: Let's reserve for future work (if at all):

> Mechanically, it's quite simple. DuckDB has a built-in SQLite scanner that can attach and query a .sqlite file directly:

```python
import duckdb

con = duckdb.connect()
con.execute(f"ATTACH '{sqlite_path}' AS sc (TYPE sqlite, READ_ONLY)")
df = con.execute("""
    SELECT
        CAST(Nuclei.TableNumber  AS BIGINT)  AS TableNumber,
        ...
    FROM sc.Nuclei
    INNER JOIN sc.Image ON ...
""").df()
con.close()
```

> That replaces _get_single_cell_engine, _check_single_cell_correctness, and _get_joined_image_nuclei_tables in one shot. The CAST awkwardness goes away because DuckDB enforces strict types natively. The Windows file-locking problem also goes away because DuckDB closes the file the moment con.close() is called — no connection pool.
>
> The complications:
>
> - Public API accepts a sqlalchemy.engine.Engine as single_cell_input — that's part of the documented interface. Dropping SQLAlchemy entirely would be a breaking change for any caller passing an engine directly.
> - duckdb is not a current dependency — it would need to be added (either as a core dep or a new optional extra like cell-locations).
> - Schema introspection — _check_single_cell_correctness uses sqlalchemy.inspect() to list tables and columns. DuckDB equivalent is PRAGMA database_list / SHOW TABLES / DESCRIBE, which is straightforward but needs rewriting.
> - Test fixtures — conftest.py creates a sqlalchemy.engine.Engine fixture (single_cell_input_engine) that's passed directly to CellLocation. Those tests would need updating.
> - Bottom line: the core query path is a ~20-line swap. The surface area grows to maybe 60–80 lines once you account for the public API compatibility shim and the schema-check rewrite. Reasonable future-work scope — probably a focused PR of its own.

gwaybio: Given the dependency swap, other potential environment challenges, and our deprecation plan, my instinct is for this to not consume our time for future work.

tests/test_cyto_utils/test_load.py
gwaybio: it is the csv.Sniffer itself failing on windows - i cringe at adding a dependency just for this single function for a single OS. It feels to me like a big shim for a small issue. 

That being said, I can dig a bit deeper to find an alternative solution that doesn't require sniffing or adding a dependency. This fix will likely keep the windows test skip here for the sniffer and an added test for windows-specificity.

pycytominer/cyto_utils/cell_locations.py
gwaybio: fixed in a5472a8c1dfad12d4159ae3770b92d2158513bc8!

pycytominer/cyto_utils/cell_locations.py
gwaybio: formatted in a5472a8c1dfad12d4159ae3770b92d2158513bc8

tests/test_cyto_utils/test_load.py
gwaybio: ok! I think i have an elegant solution. It is two part:

# Solution 1: Loading platemaps

Fixed in e43025ba3c2492986d5ddcd3dd36eb9447126564

- I added `sep` as an optional parameter to `load_platemap()`
- Added `platemap_sep` to `annotate()`, which is the only function that uses `load_platemap()`
- Added to docstring why this is important for windows users

The rationale is that a windows user now has the ability to explicitly state how pandas should read the file, without relying on unstable csv sniffing

# Solution 2: Loading profiles

Fixed in 1bc59d1f1d38031dfaf7b325221670a83b551f90

This is the more challenging solution, and I'm glad you poked on this one! If we didn't perform this, a subset of windows users would likely become frustrated by lack of support.

Rather than pursuing the same solution for platemaps, I elected to **drop** support for windows users and flat text files. It would be too much edits to core functions for an edge case. We recommend using parquet and I can see us phasing out csv entirely in a future version. 

- Raising an OSError if load_profiles gets to the infer_delim line. Hopefully this happens in the first call to Pycytominer for a windows users and this functionality, so that they can go and follow the support we describe in the error message
- Keeping the test skip for csv sniffing in infer_delim
- Adding tests to run only for windows matching the expected error message, but also that all other formats behave as expected.

pycytominer/cyto_utils/cell_locations.py
d33bs: I understand. Just adding in here from a code smell perspective: I feel that the sooner we get out of the binding between SQLite, SQLAlchemy, and Pandas the better. These are technologies which are ergonomically difficult, include many historical gotcha's, and data type vague.

tests/test_cyto_utils/test_load.py
gwaybio: These solutions required additional fixes! The two main ones are:

- Switching to parquet files in tests (we probably should be doing this anyway)
- Adding additional warnings for windows users using csv files in `external_metadata` for `annotate.py`

gwaybio:
thanks for the review @d33bs - i'm very happy you prodded on the test skip! my additional research uncovered a lot more buried in the depths. 

I kept comments open for additional discussion if necessary.

When you are able, please provide another review.

Thanks!

d33bs (approved):
Thanks for addressing things and for the follow-ups, I think this LGTM!

gwaybio:
Thanks @d33bs !!

Issue #704, opened 2026-06-04
FeatureRequest: Replace csv.sniffer


gwaybio:
### Feature type

- [ ] Add new functionality

- [x] Change existing functionality

### General description of the proposed functionality

Currently, `infer_delim()` uses [csv.Sniffer](https://docs.python.org/3/library/csv.html#csv.Sniffer), and `infer_delim()` is used by `load_platemap()` and `load_profiles()`.

We ran into windows-specific problems for csv.Sniffer in #698. 

We should consider swapping out the csv module for something else, or consider removing support for csv files altogether

### Feature example

N/A

### Alternative Solutions

change sniffer approach, drop csv support, enforce certain delimiters, expose all potential delimiter options to users

### Contribution interest

- [ ] I would be interested in submitting a pull request to help implement this feature.

axiomcura:
I came across this package and they claim to be a drop in replacement for python's `csv`  package. 

GitHub: https://github.com/alan-turing-institute/CleverCSV
Docs: https://clevercsv.readthedocs.io/en/latest/

This might be a good option.

Issue #624, opened 2026-02-13
Address automatic inference of output file type using filenames (CLI and core functions)


d33bs:
> oh, and also, should we have to provide the output_type flag if the output file is suffixed with the output file type? (maybe beyond scope of this PR) 

 _Originally posted by @gwaybio in [#622](https://github.com/cytomining/pycytominer/pull/622/changes#r2768467570)_

PR #708, opened 2026-06-11
Adding roadmap

gwaybio:
Explicitly signaling our pycytominer roadmap

## Summary by CodeRabbit

* **Documentation**
  * Updated the project roadmap with a new milestone plan covering v1.7 through v2.0 and beyond.
  * Added a Mermaid-based timeline outlining planned API improvements, a fluent `CytoDataFrame`-based pipeline direction, and upcoming deprecations/breaking changes.
  * Clarified the current v1.6 state and expanded contribution guidance to point to the appropriate contributor instructions.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -4,7 +4,7 @@ Pycytominer's vision is to perform the image-based profiling pipeline **reproduc
 This roadmap outlines the milestones toward a v2 release that fulfills that vision through a modernized API, a structured data model, and a high-performance DataFrame backend.
 
 ```mermaid
-%%{init: {'theme': 'dark', 'themeVariables': {'cScale2': '#6d28d9', 'cScale4': '#111111'}}}%%
+%%{init: {'theme': 'dark', 'themeVariables': {'cScale0': '#1e40af', 'cScale1': '#0f766e', 'cScale2': '#6d28d9', 'cScale3': '#b91c1c', 'cScale4': '#111111'}}}%%
 timeline
     title Pycytominer Roadmap
     section New API Foundations
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,135 @@
+# Pycytominer Roadmap
+
+Pycytominer's vision is to perform the image-based profiling pipeline **reproducibly and extremely fast** for humans and AI agents.
+This roadmap outlines the milestones toward a v2 release that fulfills that vision through a modernized API, a structured data model, and a high-performance DataFrame backend.
+
+```mermaid
+%%{init: {'theme': 'dark', 'themeVariables': {'cScale2': '#6d28d9', 'cScale4': '#111111'}}}%%
+timeline
+    title Pycytominer Roadmap
+    section New API Foundations
+        v1.7 : ProfileData and OutputConfig dataclasses
+             : Feature selection improvements
+             : API consistency fixes
+             : Deprecation warnings on legacy functions
+    section Fluent Pipeline
+        v1.8 : ProfileData pipeline methods
+             : Core functions accept ProfileData as input
+             : Provenance and logging via chaining
+    section Performance
+        v1.9 : Parallel execution across plates and batches
+             : Pandas to Polars migration inside ProfileData
+             : Benchmark suite
+    section Clean API and Deprecation
+        v2.0 : Remove legacy functions and deprecated shims
+             : Finalize clean public API
+             : Schema improvements
+             : Decouple institution-specific utilities
+    section Agentic Infrastructure
+        v2+ : Expose pipeline steps as agent-callable skills
+            : Automated reproducible workflows
+            : Native support for humans and AI agents
+```
+
+---
+
+## Current State — v1.6
+
+Pycytominer provides a suite of standalone functions (`aggregate`, `normalize`, `annotate`, `feature_select`, `consensus`) that cover the full image-based profiling pipeline.
+The library supports multiple file formats (CSV, Parquet, AnnData, CytoTable Warehouse), runs on Linux, macOS, and Windows, and is tested across Python 3.10–3.14.
+
+---
+
+## Milestone 1 — API Foundations (v1.7)
+
+**Goal:** Introduce a structured data model and eliminate repeated boilerplate across the API.
+
+### ProfileData
+
+- [ ] Add `ProfileData` dataclass ([#327](https://github.com/cytomining/pycytominer/issues/327)) to bundle a DataFrame with its resolved feature and metadata column lists so they are computed once, not inferred repeatedly across function calls
+- [ ] Add `OutputConfig` dataclass to consolidate the four repeated output parameters (`output_file`, `output_type`, `compression_options`, `float_format`) shared across all core functions into a single per-call object
+
+### Feature Selection
+
+- [ ] Faster correlation computation ([#633](https://github.com/cytomining/pycytominer/issues/633))
+- [ ] Numerical variance filtering ([#656](https://github.com/cytomining/pycytominer/issues/656))
+- [ ] Rename variance thresholding for clarity ([#634](https://github.com/cytomining/pycytominer/issues/634))
+
+### API Consistency
+
+- [ ] Align `aggregate` parameter naming with other core functions ([#635](https://github.com/cytomining/pycytominer/issues/635))
+- [ ] Improve `annotate` to avoid unintended column renaming ([#660](https://github.com/cytomining/pycytominer/issues/660))
+- [ ] Minor schema decisions (e.g., handling of location columns ([#224](https://github.com/cytomining/pycytominer/issues/224)))
+
+### Documentation
+
+- [ ] Fix `Returns` docstrings across all core functions ([#636](https://github.com/cytomining/pycytominer/issues/636))
+
+---
+
+## Milestone 2 — Fluent Pipeline (v1.8)
+
+**Goal:** Make `ProfileData` the primary entry point for running the profiling pipeline (still supporting original API), enabling method chaining and natural provenance tracking.
+
+- [ ] Add pipeline methods to `ProfileData` (`.normalize()`, `.feature_select()`, `.aggregate()`, `.annotate()`, `.consensus()`) — each delegates to the existing standalone function and returns a new `ProfileData`
+- [ ] Core functions accept `ProfileData` as input in addition to `str` / `pd.DataFrame` — no breaking changes
+- [ ] Provenance and logging emerge naturally from chaining: for example, callers can compare `.features` before and after `feature_select` to see exactly which features were dropped by which operation
+
+```python
+# Example chaining
+result = (
+    ProfileData.from_file("profiles.parquet")
+    .aggregate(strata=["Metadata_Well"])
+    .normalize(method="standardize", samples="Metadata_treatment == 'DMSO'")
+    .feature_select(operations=["variance_threshold", "correlation_threshold"])
+)
+```
+
+---
+
+## Milestone 3 — Parallelization and High Performance Backend (v1.9)
+
+**Goal:** Make pycytominer extremely fast by enabling parallel execution across the pipeline.
+
+- [ ] Parallel execution of independent pipeline steps across plates, batches, or wells
+- [ ] Leverage `ProfileData` as the natural unit of parallelism — each object is self-contained and stateless
+- [ ] Benchmark suite to track performance across releases
+
+### Polars Migration
+
+- [ ] Swap the DataFrame backend inside `ProfileData` from pandas to Polars — the change is contained within `ProfileData`, isolating all call sites from the migration
+- [ ] Validate performance improvements across the full pipeline
+- [ ] Maintain backward compatibility where possible; document breaking changes clearly
+
+---
+
+## Milestone 4 — Clean API (v2.0)
+
+**Goal:** Finalize a clean, stable public API, deprecate old institution-specific functions and introduce other minor, but breaking changes (e.g., some stale parameter names).
+
+### API Cleanup
+
+- [ ] Deprecate and remove legacy API ([#705](https://github.com/cytomining/pycytominer/issues/705))
+- [ ] Remove `normalize()` string-encoded missing value shim ([#646](https://github.com/cytomining/pycytominer/issues/646))
+- [ ] Decouple `format_broad_cmap` / `clean_cellprofiler` ([#625](https://github.com/cytomining/pycytominer/issues/625))
+- [ ] Refactor `SingleCells` ([#269](https://github.com/cytomining/pycytominer/issues/269))
+- [ ] Replace `csv.sniffer` ([#704](https://github.com/cytomining/pycytominer/issues/704))
+- [ ] Retire `collate.py` upload/download flags ([#231](https://github.com/cytomining/pycytominer/issues/231))
+- [ ] Fix `collate.py` compartment handling ([#272](https://github.com/cytomining/pycytominer/issues/272))
+
+---
+
+## Beyond v2 — Agentic Infrastructure
+
+**Goal:** Make pycytominer natively usable by AI agents as a set of composable, callable skills.
+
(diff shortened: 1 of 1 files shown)


gwaybio:
cc @axiomcura

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+             : Feature selection improvements
+             : API consistency fixes
+             : Deprecation warnings on legacy functions
+    section Fluent Pipeline
+        v1.8 : ProfileData pipeline methods
+             : Core functions accept ProfileData as input
+             : Provenance and logging via chaining
+    section Performance
+        v1.9 : Parallel execution across plates and batches
+             : Pandas to Polars migration inside ProfileData
```
d33bs: Polars will likely create possible breaking changes with internals. I suggest moving this to 2.0 .

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+- [ ] Core functions accept `ProfileData` as input in addition to `str` / `pd.DataFrame` — no breaking changes
+- [ ] Provenance and logging emerge naturally from chaining: for example, callers can compare `.features` before and after `feature_select` to see exactly which features were dropped by which operation
+
+```python
+# Example chaining
+result = (
+    ProfileData.from_file("profiles.parquet")
+    .aggregate(strata=["Metadata_Well"])
+    .normalize(method="standardize", samples="Metadata_treatment == 'DMSO'")
+    .feature_select(operations=["variance_threshold", "correlation_threshold"])
```
d33bs: Consider keeping the same existing api accepting similar input and sending to ProfileData on the backend so the workflow remains recognizable and requires no additional change. Here, there could be a comment made to imply that the data are wrapped / converted using the dataclass.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+---
+
+## Milestone 2 — Fluent Pipeline (v1.8)
+
+**Goal:** Make `ProfileData` the primary entry point for running the profiling pipeline (still supporting original API), enabling method chaining and natural provenance tracking.
+
+- [ ] Add pipeline methods to `ProfileData` (`.normalize()`, `.feature_select()`, `.aggregate()`, `.annotate()`, `.consensus()`) — each delegates to the existing standalone function and returns a new `ProfileData`
+- [ ] Core functions accept `ProfileData` as input in addition to `str` / `pd.DataFrame` — no breaking changes
+- [ ] Provenance and logging emerge naturally from chaining: for example, callers can compare `.features` before and after `feature_select` to see exactly which features were dropped by which operation
```
d33bs: Consider better clarifying what is meant by "emerge naturally". I'd suggest keeping things relatively open when it comes to the implementation, which might have been the intent here but wasn't sure.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+            : Automated reproducible workflows
+            : Native support for humans and AI agents
+```
+
+---
+
+## Current State — v1.6
+
+Pycytominer provides a suite of standalone functions (`aggregate`, `normalize`, `annotate`, `feature_select`, `consensus`) that cover the full image-based profiling pipeline.
+The library supports multiple file formats (CSV, Parquet, AnnData, CytoTable Warehouse), runs on Linux, macOS, and Windows, and is tested across Python 3.10–3.14.
```
d33bs: The versioning here for Python feels overspecific, consider dropping.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+---
+
+## Milestone 1 — API Foundations (v1.7)
+
+**Goal:** Introduce a structured data model and eliminate repeated boilerplate across the API.
+
+### ProfileData
+
+- [ ] Add `ProfileData` dataclass ([#327](https://github.com/cytomining/pycytominer/issues/327)) to bundle a DataFrame with its resolved feature and metadata column lists so they are computed once, not inferred repeatedly across function calls
```
d33bs: How will ProfileData integrate with CytoDataFrame, if at all? If we plan to have both it probably means wrapper-inception (wrappers around wrappers). Imagine: we present Pycytominer.ProfileData and CytoDataFrame at the same time and the question comes up: "why have two data objects for similar goals"?

I vote we integrate them and make a push towards a common image-based profiling data type which can store the column information here and much more. While this might make the immediate path complex at first, it shows a more unified front for Cytomining with existing lessons learned from the project there.

d33bs: Another idea: instead of an upfront cost to the user API and abstractions, could ProfileData live as a decorator pattern to core functions where needed (column filtering cost is likely not high and may change through each core function)? Or is there implied value to users by supplying the dataclass in return? If the latter, maybe we should spell this out briefly here.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+- [ ] Add `OutputConfig` dataclass to consolidate the four repeated output parameters (`output_file`, `output_type`, `compression_options`, `float_format`) shared across all core functions into a single per-call object
```
d33bs: What if we expanded on the decorator pattern through `@write_to_file_if_user_specifies_output_details` to help with this instead of a public/user-facing dataclass?

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Goal:** Make pycytominer extremely fast by enabling parallel execution across the pipeline.
+
+- [ ] Parallel execution of independent pipeline steps across plates, batches, or wells
+- [ ] Leverage `ProfileData` as the natural unit of parallelism — each object is self-contained and stateless
+- [ ] Benchmark suite to track performance across releases
+
+### Polars Migration
+
+- [ ] Swap the DataFrame backend inside `ProfileData` from pandas to Polars — the change is contained within `ProfileData`, isolating all call sites from the migration
```
d33bs: If we move from Pandas to Polars it's very likely not possible to isolate the changes to only the dataclass. For ex. any dataframe operations would need to undergo an overhaul (the API's and processing decisions differ quite a bit).

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+- [ ] Validate performance improvements across the full pipeline
```
d33bs, line 105: Consider addressing "how" we validate, getting a little more specific; for ex. is this through continuous benchmarking?

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+- [ ] Maintain backward compatibility where possible; document breaking changes clearly
```
d33bs: Breaking changes imply we move this work to a major release, if following https://semver.org/ conventions.

d33bs (commented):
Exciting! Adding some initial thoughts.

ROADMAP.md
gwaybio: maybe this is flawed thinking, but we would continue to support v1 API until dropping it in v2. In other words, the functions would continue working with polars or pandas, through ProfileData. It's possible i don't know enough about polars to be 100% sure this is possible.

ROADMAP.md
gwaybio: sounds good - yeah, this text is an odd mix of specificity and general thinking. Will revise

ROADMAP.md
gwaybio: i don't think i understand this comment. What do you mean "same existing API"? and "sending to ProfileData on the backend"?

Is your suggestion to drop the example chain? I'll add something in the next commit, which may be a middle ground. LMK!

ROADMAP.md
gwaybio: i'm ok keeping this vague for now

ROADMAP.md
gwaybio: got it! related to a previous comment - i'll adjust

ROADMAP.md
gwaybio: sounds good! so you're intuition is to make all the ProfileData changes we discuss here inside CytoDataFrame? I agree it will be more complex, but probably more stable in the long run!

This begs a question then, why are we creating CytoDataFrame and not simply using AnnData? I think we need to have a compelling answer to this question.

gwaybio: my thought is to reduce redundant code, have a unified place for storing methods, enable functionality for all API functions consistently, and to enable chaining. I like the idea of CytoDataFrame handling this - i'll make a few edits in response

ROADMAP.md
gwaybio: how does the decision to use cytodataframe influence your thinking on this expansion?

ROADMAP.md
gwaybio: yes, this is under v2

ROADMAP.md
axiomcura: As a person who has not used CytoDataFrame, one immediate concern with integrating `ProfileData` directly with `CytoDataFrame` is dependency weight and compatibility. If `CytoDataFrame` becomes a hard dependency of Pycytominer, it could narrow Pycytominer’s currently supported Python range and introduce a much larger dependency stack, including notebook, image, and visualization-related packages. 

My questions is... do we want that ? 🤔

gwaybio:
i responded to your comments @d33bs - thanks! I think this will help our thinking quite a bit for the months ahead. Can you take another look? There are some additional discussion items. please also make any changes directly

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+            : Automated reproducible workflows
+            : Native support for humans and AI agents
+```
+
+---
+
+## Current State — v1.6
+
+Pycytominer provides a suite of standalone functions (`aggregate`, `normalize`, `annotate`, `feature_select`, `consensus`) that cover the full image-based profiling pipeline.
+The library supports multiple file formats (CSV, Parquet, AnnData, CytoTable Warehouse), runs on Linux, macOS, and Windows.
```
axiomcura, line 35: This make it seem that all file formats work with all OSes, which I'd argue is not true: 
https://github.com/gwaybio/pycytominer/blob/edbe8d6203732f13813c90abe086d5d74471b3d3/pycytominer/cyto_utils/load.py#L394

Given the csv.sniffer problem we have.  #704

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+---
+
+## Milestone 1 — API Foundations (v1.7)
+
+**Goal:** Introduce a structured data model and eliminate repeated boilerplate across the API.
+
+### ProfileData
+
+- [ ] Add `ProfileData` dataclass ([#327](https://github.com/cytomining/pycytominer/issues/327)) to bundle a DataFrame with its resolved feature and metadata column lists so they are computed once, not inferred repeatedly across function calls
+- [ ] Add `OutputConfig` dataclass to consolidate the four repeated output parameters (`output_file`, `output_type`, `compression_options`, `float_format`) shared across all core functions into a single per-call object
```
axiomcura: I like this idea! but do you think that this should probably include or explicitly defer filename-based output type inference. #624

ROADMAP.md
```diff
@@ -0,0 +1,126 @@
...
+)
+```
+
+---
+
+## Milestone 3 — Performance enhancements (v1.9)
+
+**Goal:** Make pycytominer fast by enabling parallel execution across the pipeline.
+
+- [ ] Parallel execution of independent pipeline steps across plates, batches, or wells
```
axiomcura: Maybe this may not be a problem in the field, but I feel like we need some sort of scope rule, not just “plates, batches, or wells.”

One concern with the parallelization  is that “across plates, batches, or wells” may be too broad without defining the valid scope for each operation. Some steps are safe to parallelize by plate or batch, but others depend on the full dataset or a specific reference population. For example, normalization, variance filtering, and correlation filtering can produce different results depending on whether they are computed globally or within each partition.

Maybe it is not worth mentioning here, but it could be an issue to keep in mind when developing parallelization support.

axiomcura (commented):
Looks good, left some comments !

ROADMAP.md
gwaybio: yep, good point. i'll make a note

ROADMAP.md
gwaybio: it's a great point Erik - I will adjust!

ROADMAP.md
gwaybio: there's multiple potential visions here - Dave is postulating that energy placed toward ProfileData would actually be better spent placed toward CytoDataFrame so that ProfileData would not integrate but be replaced. All of these are good thoughts regarding tradeoffs here - I don't think we have enough visibility here yet to make a determination

gwaybio:
thanks for your comments @axiomcura and @d33bs - may I ask for another review? Would be nice to get this merged in

ROADMAP.md
d33bs: What if we retained parallel 1.x releases which are pandas compatible and 2.x releases which were polars compatible until we reach a certain point with 1.x expectations (so here, 1.x releases would still be deployed past 2.0.0)? This would mean people need not be forced into new processing until they're ready.

ROADMAP.md
d33bs: Mostly here I felt we should avoid constraining users to a required pre-processing step where they're used to just sending the data along, so something like the following. I think we should be explicit in using a consistent alias where possible (I think we might do this in some places but not others).

```python
import pycytominer as pm

profiles = ... # load the data somehow
profiles = pm.aggregate(profiles)
profiles = pm.normalize(profiles)
profiles = pm.feature_select(profiles)
```

ROADMAP.md
d33bs: I think using an external library would alleviate pressures on Pycytominer to provide output beyond in-memory (via CytoDataFrame or otherwise). I feel this would be a healthy practice of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) at a project level.

ROADMAP.md
d33bs: Sounds good!

d33bs: Thanks @gwaybio and @axiomcura for the thoughts here!

One possibility is that `ProfileData` evolves into a lightweight `ProfileSchema` rather than another dataframe-like object. Pycytominer could use it internally for inference/validation today, while pandas, Polars, CytoDataFrame, and even AnnData adapters could all support the same schema (to your point about "why not AnnData" - all are welcome in Cytomining 😃). 

This also reduces the risk + pressure of any one project needing to own too much complexity if an audience isn't interested in owning up to technology which isn't a fit for them. Maybe this way it feels more comfortable, a BYOD (bring-your-own-data) occasion.

That would let us experiment with richer implementations like CytoDataFrame without immediately making them hard dependencies, while still moving toward a common profiling data model within Pycytominer (and elsewhere if desired).

Here's some pseudocode which represents what I'd see happening:

```python
from dataclasses import dataclass


# a schema for profile data
@dataclass(frozen=True)
class ProfileSchema:
    metadata_cols: tuple[str, ...]
    feature_cols: tuple[str, ...]

    @classmethod
    def infer(cls, df):
        return cls(
            metadata_cols=tuple(infer_cp_metadata(df)),
            feature_cols=tuple(infer_cp_features(df)),
        )


# a decorator for inferring from the schema based on flexible input data
def with_profile_schema(func):
    def wrapper(data, schema: ProfileSchema | None = None, *args, **kwargs):
        schema = schema or getattr(data, "schema", None)

        if hasattr(data, "to_profile_dataframe"):
            data = data.to_profile_dataframe()

        schema = schema or ProfileSchema.infer(data)

        return func(
            data,
            schema=schema,
            *args,
            **kwargs,
        )

    return wrapper

# normalize (or other core fxns) gets decorated to reduce code
@with_profile_schema
def normalize(df, schema: ProfileSchema | None = None):
    features = df.loc[:, schema.feature_cols]
    metadata = df.loc[:, schema.metadata_cols]

    # normalization logic...
    return concat([metadata, features])

# cytodataframe can accept the same schema
class CytoDataFrame(pd.DataFrame):  # or pl.DataFrame, in the near future
    schema: ProfileSchema

# anndata can be adapted to use the same schema
class AnnDataAdapter:
    def __init__(self, adata):
        self.adata = adata
        self.schema = ProfileSchema.infer(
            self.to_profile_dataframe()
        )

    def to_profile_dataframe(self):
        return adata_to_dataframe(self.adata)


# All of these would work:
normalize(df)
normalize(CytoDataFrame(df))
normalize(AnnDataAdapter(adata))
````

d33bs: Hi @axiomcura - would you have any thoughts on this? Want to make sure we incorporate your vision of things!

d33bs (approved):

gwaybio:
thanks all! merging this in now

Issue #634, opened 2026-03-04
FeatureRequest: Variance Thresholding Feature-Selection Name Change


MattsonCam:
### Feature type

- [x] Add new functionality

- [x] Change existing functionality

### General description of the proposed functionality

From what I understand, the variance thresholding in `pycytominer` is performing near-zero variance thresholding. The current filters still account for the variance to a degree, but I don't think this type of thresholding will filter continuous-valued features very well. The current filtering here would perform better with discrete or psuedo-discrete features. 

Variance thresholding would be more akin to this implementation in `scikit-learn`:
https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.VarianceThreshold.html#variancethreshold

For these reasons, consider renaming the option and function names to reflect this other type of thresholding. Also, consider introducing what `scikit-learn` would term variance thresholding. 

This is the function that contains the current implementation of the variance thresholding:
https://github.com/cytomining/pycytominer/blob/6493ec33319acc7bc516fbc86fbdaa8e03388e44/pycytominer/operations/variance_threshold.py#L14

If you need any clarification on this, just let me know.

### Feature example

This would just consist of renaming the "variable thresholding" variables, functions, strings, etc... to another name, and possibly implementing the scikit-learn version of variable thresholding.

axiomcura:
Closed due to #719

PR #719, opened 2026-06-23
Add var_eps support to variance_threshold to filter low-variance features

### LINKED ISSUE #656: FeatureRequest: Numerical variance filtering on top of current NZV filter
### Feature type

- [x] Add new functionality

- [x] Change existing functionality

The current implementation of `variance_threshold` and its use in `feature_select` relies exclusively on frequency-based heuristics (NZV) rather than numeric variance:
- https://github.com/cytomining/pycytominer/blob/8a87aa6b473d741fd15871461f0e41969e3c178c/pycytominer/feature_select.py#L144-L151

Although NZV is a robust variance filter for count based and categorical features, it is insensitive and inappropriate for variance filtering on continuous-valued features. As a result, low-spread continuous columns can still pass the feature selection causing downstream issues. Please consider either adding a numeric variance-based filter (e.g., `.var(ddof=0) < eps`) or renaming/documenting this behavior more explicitly. Ideally, the implementation could be made dtype- or feature-semantic–aware, applying NZV-style filtering for discrete/quantized features and numeric variance thresholds for continuous features, ensuring appropriate behavior across different data modalities.

### Feature example

```python
def variance_threshold(
    population_df,
    features="infer",
    samples="all",
    freq_cut=0.05,
    unique_cut=0.01,
    var_eps=None,
):
    # existing setup:
    # - validate thresholds
    # - subset rows if samples != "all"
    # - infer / validate feature list

    if features == "infer":
        inferred_features = infer_cp_features(population_df)
    else:
        inferred_features = features

    population_df = population_df.loc[:, inferred_features]

    # proposed addition:
    # split features by semantic type and/or dtype
    discrete_features, continuous_features = infer_feature_type(
        population_df.loc[:, inferred_features]
    )

    excluded_features = []

    # existing NZV-style logic, applied only to discrete / quantized features
    if discrete_features:
        excluded_features_freq = population_df.loc[:, discrete_features].apply(
            lambda x: calculate_frequency(x, freq_cut),
            axis=0,
        )
        excluded_features.extend(
            excluded_features_freq[excluded_features_freq.isna()].index.tolist()
        )

        unique_ratio = (
            population_df.loc[:, discrete_features].nunique()
            / population_df.shape[0]
        )
        excluded_features.extend(
            unique_ratio[unique_ratio < unique_cut].index.tolist()
        )

    # proposed numeric variance logic, applied only to continuous features
    if continuous_features and var_eps is not None:
        low_variance = (
            population_df.loc[:, continuous_features].var(ddof=0) < var_eps
        )
        excluded_features.extend(
            low_variance[low_variance].index.tolist()
        )

    return list(set(excluded_features))
```

### Alternative Solutions

_No response_

### Additional information

_No response_

axiomcura:
# Description
This PR adds optional numerical variance filtering to `variance_threshold` through a new `var_eps` parameter. When `var_eps` is provided, features with population variance below that threshold are excluded alongside the existing frequency and uniqueness based near-zero variance filters. It also adds tests covering low-variance exclusion, the zero-threshold case, and invalid `var_eps` values.

Closes #656 ,   https://github.com/cytomining/pycytominer/issues/728,  https://github.com/cytomining/pycytominer/issues/634.

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a new `"frequency_threshold"` operation for feature exclusion based on frequency/uniqueness behavior.
  * Added `min_variance` support to variance-based feature exclusion, including in composite feature selection.
* **Bug Fixes**
  * Updated variance thresholding to use only `min_variance` with stricter input validation and consistent handling of inferred features and sample filtering.
* **Tests**
  * Reworked variance-threshold tests for `min_variance` behavior (low-variance removal, `0.0` keeps all, invalid inputs fail) and added coverage for feature inference and `samples` filtering.


pycytominer/operations/variance_threshold.py
```diff
@@ -17,6 +17,7 @@ def variance_threshold(
     samples: str = "all",
     freq_cut: float = 0.05,
     unique_cut: float = 0.01,
+    var_eps: float | None = None,
```
gwaybio: consider using the full name (variance_epsilon) to increase clarity - also, can you think of a clearer word than epsilon? Perhaps "remove_features_with_variance_less_than"?

pycytominer/operations/variance_threshold.py
```diff
@@ -43,6 +44,10 @@ def variance_threshold(
         Ratio (num unique features / num samples). Must range between 0 and 1.
         Remove features less than unique cut. A low unique_cut will remove features
         that have very few different measurements compared to the number of samples.
+    var_eps: float, default None
+        If not None, remove features with variance less than var_eps. A low var_eps will
+        remove features that have very low variance (e.g. this will remove a feature:
+        [1, 1, 1, 1, 1].
```
gwaybio: Based on #656 , we should be clear to document this argument to be made explicitly for continuous features. Right now  the example is not continuous. If we make a distinction here for continuous features, then we probably should also mirror this description in the other variables for categorical variables.

pycytominer/operations/variance_threshold.py
```diff
@@ -74,7 +83,10 @@ def variance_threshold(
     # Subset the DataFrame to only include the features of interest
     population_df = population_df.loc[:, inferred_features]
 
-    # Exclude features based on frequency
+    # initalize to store excluded features
```
gwaybio:
```suggestion
    # Initalize to store excluded features
```

Capitalization consistency

pycytominer/operations/variance_threshold.py
```diff
@@ -74,7 +83,10 @@ def variance_threshold(
...
+    excluded_features = []
+
+    # 1. Exclude features based on frequency
```
gwaybio: We exclude features based on what type of frequency? Low frequency?

pycytominer/operations/variance_threshold.py
```diff
@@ -102,12 +115,18 @@ def variance_threshold(
...
+    excluded_features.extend(excluded_features_unique)
+
+    # 3. Exclude features with low variance
+    # If var_eps is not None, calculate the variance of each feature and exclude those
+    # with variance less than var_eps
+    if var_eps is not None:
+        exclude_features_low_variance = population_df.var(ddof=0, skipna=True) < var_eps
+        excluded_features.extend(
+            exclude_features_low_variance[exclude_features_low_variance].index.tolist()
+        )
```
gwaybio: This is difficult to parse - is there a way to restructure this logic so the indexing variable is not the same as the dataframe used to index? I think my primary comment relates to variable naming but there is a separate comment also about pandas indexing logic - consider making this block clearer

pycytominer/operations/variance_threshold.py
```diff
@@ -102,12 +115,18 @@ def variance_threshold(
...
+
+    return list(dict.fromkeys(excluded_features))
```
gwaybio: i'm wondering why the decision to change the return here for a feature selection method? I think we should stay consistent with all other feature selection operations

tests/test_operations/test_variance_threshold.py
```diff
@@ -100,6 +100,50 @@ def test_variance_threshold():
...
+        "low_var": [1, 1, 1, 1, 1.001, 1.001],
+        "high_var": [0, 0, 10, 10, 20, 20],
+    }).reset_index(drop=True)
+
+    excluded_features = variance_threshold(
+        population_df=data_var_test_df,
+        features=data_var_test_df.columns.tolist(),
+        freq_cut=0,
+        unique_cut=0,
+        var_eps=0.0,
```
gwaybio: i recommend trying also a `None` here to test that behavior works

tests/test_operations/test_variance_threshold.py
```diff
@@ -100,6 +100,50 @@ def test_variance_threshold():
...
+    )
+
+    assert excluded_features == []
+
+
+@pytest.mark.parametrize("var_eps", [-0.1, 1, "0.1"])
```
gwaybio: Should it fail at 1? Do other variables in Pycytominer have this behavior? I presume 1 should work, since 1 should succeed at the main call `population_df.var(ddof=0, skipna=True) < var_eps`?

gwaybio (changes requested):
Thanks for the PR! Several specific in-line comments below.

Importantly, we should discuss a general comment, which is the decision to place this functionality inside `variance_threshold` and to not generate a separate function. If we decide to design a separate function, then we probably should rename this one (variance_threshold) as well. Some elements to consider in this decision include feature schemas, logging, and the planned `feature_select.parquet` summary file of features and which functions removed them.

axiomcura:
Good point @gwaybio I've been thinking about this.... and I agree that the current frequency thresholding implementation is conceptually different from variance thresholding. Frequency thresholding is generally intended for categorical or discrete features (e.g., cell neighbor counts), whereas variance thresholding is primarily designed for continuous-valued features. Because of these different assumptions, I think they would be clearer as separate feature selection methods rather than combining them under variance_threshold.

If we go in that direction, I would propose introducing a frequency_thresholding option within feature_select(), while reserving variance_threshold for continuous features only. I also think this would make logging and the planned feature_select.parquet summary more interpretable, since each removed feature could be attributed to the specific selection method that filtered it.

LMK what you think!

gwaybio:
> I think they would be clearer as separate feature selection methods rather than combining them under variance_threshold.

> If we go in that direction, I would propose introducing a frequency_thresholding option within feature_select(), while reserving variance_threshold for continuous features only. 

Sounds good. Please also keep in mind our planned schema upgrades. In other words, it may not make sense to "reserve variance_threshold for continuous features only" right now since any logic to infer feature type will be removed when the solution will be baked into the data type at instantiation. Is there a downside to running it on all non-metadata features (like we do currently for every other feature selection method)?

axiomcura:
Ahhh I see...  I discussed this with @wli51, we agreed  that frequency thresholding is unlikely to meaningfully affect continuous features, since exact value repetition is rare in continuous spaces. How ever, the behavior of variance thresholding on discrete features is less predictable and will depend heavily on parameter choices (`var_eps`).

Given the roadmap and the move toward schema-defined feature types, it may be better to no separate the functions and keep the implementation flexible for now. Supporting both variances and frequency thresholding within the current `variance_thresholding()`on  all non-metadata features. Then, once the feature schema is formalized at instantiation, we can refine or constrain the behavior of each method based on declared data types rather than implicit assumptions.

I like this approach it's much easier to maintain and update c: LMK what you think @gwaybio !

gwaybio:
> Supporting both variances and frequency thresholding within the current variance_thresholding()on all non-metadata features. Then, once the feature schema is formalized at instantiation, we can refine or constrain the behavior of each method based on declared data types rather than implicit assumptions.

My instinct is to separate the functions now, since they will be separate in the future. Where we draw the line is not performing dtype inference to drive specific behavior, since this is not what we currently do and our schema future will obviate this function.

pycytominer/operations/variance_threshold.py
axiomcura: I just added steps but lines that have not been changed have the explanation. c: 
https://github.com/cytomining/pycytominer/blob/eb1cbeaa24814a223c10b56865d1714527b45b31/pycytominer/operations/variance_threshold.py#L77-L79

pycytominer/operations/variance_threshold.py
axiomcura: reverted back!

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
 from pycytominer.cyto_utils.features import infer_cp_features
 
 
+def _subset_population_df(
```
gwaybio: it's a good idea in general to reduce duplicating functions, but this function really should be a cyto utility that all functions use. I'll elaborate below, but the `frequencey_threshold` function should be its own file, given the convention we use in `operations`.

I also believe that refactoring this function to a cyto util is beyond scope of this PR

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
...
+        inferred_features = features
+
+    return population_df.loc[:, inferred_features]
+
+
 def variance_threshold(
+    population_df: pd.DataFrame,
+    features: Union[str, list[str]] = "infer",
+    samples: str = "all",
+    min_variance: float = 1e-6,
```
gwaybio: how did you determine this default?

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
...
+        prefixed with "Cells", "Nuclei", or "Cytoplasm".
+    samples : str, default "all"
+        List of samples to perform operation on. The function uses a pd.DataFrame.query()
+        function, so you should  structure samples in this fashion. An example is
+        "Metadata_treatment == 'control'" (include all quotes).
+        If "all", use all samples to calculate.
+    min_variance: float, default 1e-6
+        Removes continuous features with variance less than this value.  A low value
+        will remove features that have very low variance (e.g. this will remove a
+        feature: [1.0000, 1.0001, 1.0000, 1.0001, 1.0000]).
```
gwaybio:
```suggestion
        Removes continuous features with variance less than this value.
```

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
...
+
+    Returns
+    -------
+    excluded_features : list[str]
+         List of features to exclude from the population_df.
+
+    """
+
+    # check if freq_cut and unique_cut are between 0 and 1
```
gwaybio: this comment no longer applies

pycytominer/operations/variance_threshold.py
```diff
@@ -148,3 +260,29 @@ def calculate_frequency(
         return np.nan
     else:
         return str(feature_column.name)
+
+
+def calculate_variance(population_df: pd.DataFrame, var_eps: float) -> list[str]:
```
gwaybio: do we need a dedicated function for this? It seems like this logic can be easily handled in the `variance_treshold` function

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
...
+
+    # Subset the population_df based on features and samples
+    population_df = _subset_population_df(
+        population_df=population_df,
+        features=features,
+        samples=samples,
+    )
+
+    # Exclude low-variance features (var < var_eps)
+    excluded_low_variance_features = calculate_variance(
```
gwaybio: i made a comment about this function below (it is likely not necessary to split a dedicated function for this), but also, the name of the function is non-specific and not accurate. It is doing more than calculating variance!

pycytominer/operations/variance_threshold.py
```diff
@@ -11,14 +12,128 @@
...
+        population_df=population_df, var_eps=min_variance
+    )
+
+    return excluded_low_variance_features
+
+
+def frequency_threshold(
```
gwaybio: To maintain parity with other feature selection operations, please move this function to a dedicated file.

pycytominer/operations/variance_threshold.py
```diff
@@ -92,7 +196,7 @@ def variance_threshold(
     # Get the number of unique features
     num_unique_features = population_df.nunique()
 
-    # Exclude features with too many (defined by unique_ratio) values in common, where
+    # 2. Exclude features with too many (defined by unique_ratio) values in common, where
```
gwaybio: typo? I don't think you meant to add `2.`?

pycytominer/operations/variance_threshold.py
```diff
@@ -116,6 +222,12 @@ def calculate_frequency(
     """Calculate frequency of second most common to most common feature.
     Used in pandas.apply()
 
+    This is done for discrete feature values by calculating the value counts of the
```
gwaybio: two comments on this comment block:

1. I would recommend avoiding stating that we do this for "discrete feature values". This is not currently accurate because do it for all features and also, the features are discrete/continuous, not the values.
2. It's not necessarily true "Example: [1, 1, 1, 1, 2, 2, ...] will return np.nan if freq_cut is 0.05." as it depends on what other values exist in the ...

tests/test_feature_select.py
```diff
@@ -299,13 +299,38 @@ def test_feature_select_get_na_columns_feature_infer():
 
 
 def test_feature_select_variance_threshold():
-    """
-    Testing feature_select and variance_threshold pycytominer function
-    """
+    """Test that feature_select removes low-variance continuous features."""
```
gwaybio: the second example feature "high_var" is not a continuous feature

gwaybio: i recommend avoiding using feature dtype nomenclature

gwaybio (changes requested):

pycytominer/operations/variance_threshold.py
axiomcura: I agree, that refactoring is beyond the scope of this PR. I will revert this implementation, move the frequency thresholding to a separate module, and create a new issue to address the refactoring of the population dataframe subsetting logic!

pycytominer/operations/variance_threshold.py
axiomcura: This is now within the function!

pycytominer/operations/variance_threshold.py
axiomcura: Fixed!

pycytominer/operations/variance_threshold.py
axiomcura: Done!

pycytominer/operations/variance_threshold.py
axiomcura: fixed!

tests/test_feature_select.py
axiomcura: fixed!

pycytominer/operations/variance_threshold.py
axiomcura: removed!

pycytominer/operations/variance_threshold.py
axiomcura: issue made! https://github.com/cytomining/pycytominer/issues/721

tests/test_operations/test_variance_threshold.py
axiomcura: The current `variance_threshold` implementation only accepts `float` values for `min_variance`. I'll update this to accept `int` values.

pycytominer/operations/variance_threshold.py
axiomcura: I used this practical guide on variance thresholding as a reference for choosing a conservative default: https://www.datasciencewithmarco.com/blog/a-practical-guide-to-feature-selection-using-sklearn#:~:text=by%20the%C2%A0author.-,From%20the%20figure%20above,-%2C%20we%20can%20now

The rationale is that features with variance below `1e-6` are nearly invariant, meaning they show very little change across samples. Because these features are unlikely to contribute meaningful signal and may mostly reflect numerical noise, removing them is a reasonable default. The threshold is intentionally conservative, so it should mainly remove features that are effectively constant while minimizing the risk of filtering out informative low-variance features.

axiomcura: > I used this practical guide on variance thresholding as a reference for choosing a conservative default: https://www.datasciencewithmarco.com/blog/a-practical-guide-to-feature-selection-using-sklearn#:~:text=by%20the%C2%A0author.-,From%20the%20figure%20above,-%2C%20we%20can%20now
> 
> The rationale is that features with variance below `1e-6` are nearly invariant, meaning they show very little change across samples. Because these features are unlikely to contribute meaningful signal and may mostly reflect numerical noise, removing them is a reasonable default. The threshold is intentionally conservative, so it should mainly remove features that are effectively constant while minimizing the risk of filtering out informative low-variance features.

LMK what you think!

gwaybio: sounds good to me - probably worth adding a comment to the docstring to this effect (if you don't have one already). The link also begs the question why we're not implementing this in sklearn?

gwaybio:
once you merge main the build integration should pass (fix in #720)

axiomcura:
#720 is merged @gwaybio and all tests passed.

gwaybio:
Thanks Erik, one remaining comment

pycytominer/operations/variance_threshold.py
axiomcura: That's a good point! I can switch this to scikit-learn’s implementation.

I initially used `.var()` because it is optimized for pandas DataFrames and has relatively low overhead. However, after looking more closely at scikit-learn’s implementation, it uses its own optimized variance calculation. I agree that using scikit-learn here makes the behavior more stable, consistent, and better maintained.

I’ll update this!

axiomcura:
@gwaybio done!

pycytominer/operations/frequency_threshold.py
```diff
@@ -0,0 +1,162 @@
...
+    if samples != "all":
+        population_df = population_df.query(expr=samples)
+    if features == "infer":
+        inferred_features = infer_cp_features(population_df)
+    if features == "infer":
+        inferred_features = infer_cp_features(population_df)
+    elif isinstance(features, list):
+        inferred_features = features
+    else:
+        raise ValueError('features must be a list of column names or "infer"')
```
gwaybio: this appears to be repeated code, please delete

pycytominer/operations/frequency_threshold.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Parameters
+    ----------
+    feature_column : pd.Series
+        Pandas series of the specific feature in the population_df
+    freq_cut : float, (suggested but unenforced default of 0.05)
+        Ratio (2nd most common feature val / most common). Must range between 0 and 1.
+        Remove features lower than freq_cut. A low freq_cut will remove features
+        that have large difference between the most common feature and second most
+        common feature. (e.g. this will remove a feature: [1, 1, 1, 1, 0.01, 0.01, ...])
```
gwaybio: this docstring isn't accurate to this function - it will return the feature name

pycytominer/operations/variance_threshold.py
```diff
@@ -34,117 +37,57 @@ def variance_threshold(
...
 
     """
 
-    # check if freq_cut and unique_cut are between 0 and 1
-    if not 0 <= freq_cut <= 1:
-        raise ValueError("freq_cut variable must be between (0 and 1)")
-    if not 0 <= unique_cut <= 1:
-        raise ValueError("unique_cut variable must be between (0 and 1)")
+    # Checking for min_variance type and value
+    if not isinstance(min_variance, (int, float)) or isinstance(min_variance, bool):
```
gwaybio: min_variance can be a bool? I think you can drop the logic after or

pycytominer/operations/variance_threshold.py
```diff
@@ -34,117 +37,57 @@ def variance_threshold(
...
+    if min_variance < 0:
+        raise ValueError("min_variance must be non-negative")
+
+    # Checking for features type and value
+    if features == "infer":
+        inferred_features = infer_cp_features(population_df)
+    elif isinstance(features, list):
+        inferred_features = features
+    else:
+        raise ValueError('features must be a list of column names or "infer"')
```
gwaybio: this is redundant with logic below (see line 72)

pycytominer/operations/frequency_threshold.py
gwaybio: I made a similar comment in the variance threshold function, except I noticed the sample query happens before the feature subsetting. If you remove the logic I flag here, then this function will do the oppositve (feature subsetting before sample query). Please make sure this is consistent between the two functions you change here, as well as the other feature selection functions in the package

tests/test_feature_select.py
```diff
@@ -403,7 +442,7 @@ def test_feature_select_all():
     result = feature_select(
         profiles=data_all_test_df,
         features=data_all_test_df.columns.tolist(),
-        operation=["drop_na_columns", "correlation_threshold", "variance_threshold"],
+        operation=["drop_na_columns", "correlation_threshold", "frequency_threshold"],
```
gwaybio: this function is called `test_feature_select_all` - can you think of a way to add variance threshold? You will likely have to add a new column (see pattern earlier in the function)

gwaybio (changes requested):
A few new items to clean up, nearly there!

(you'll notice I rebased as previous commits leaked in here when you merged main)

gwaybio:
please also make sure functionality is exposed in CLI

gwaybio:
@axiomcura - is this ready for me to re-review?

axiomcura:
@gwaybio Yes, It is ready for your re-review. I have also merged the latests changes into the this branch as well.

pycytominer/operations/variance_threshold.py
```diff
@@ -15,11 +14,15 @@ def variance_threshold(
     population_df: pd.DataFrame,
     features: Union[str, list[str]] = "infer",
     samples: str = "all",
-    freq_cut: float = 0.05,
-    unique_cut: float = 0.01,
+    min_variance: int | float = 1e-6,
```
gwaybio:
```suggestion
    min_variance: Union[int, float] = 1e-6,
```

For internal consistency

gwaybio (approved):
looks good @axiomcura - thanks for this contribution!

gwaybio:
I'm going to commit this change and then merge!

axiomcura:
Thanks @gwaybio Now merging!

PR #748, opened 2026-08-12
Fix Spherize: compute SVD on transformed (not raw) data (#747)

### LINKED ISSUE #747: Bug: ZCA Spherize does not whiten
### Issue description

Hi,

I noticed that Spherize with ZCA or ZCA-cor does not produce a whitened output. With the default center=True, the output is not decorrelated. I made a test script with test data here, comparing the pycytominer Spherize with the pypi package zca and one with ZCA implemneted on top of sklearn's PCA. 

An observation is that if you pre-center your inputs the ZCA version returns decorrelated outputs but it does not help ZCA-cor.

### Example code with output

This is my test script:
```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.11"
# dependencies = ["pycytominer", "zca", "scikit-learn"]
# ///
"""Check the pycytominer ZCA Spherize methods against two independent whitenings."""

import numpy as np
import pandas as pd
from sklearn.datasets import make_blobs
from sklearn.decomposition import PCA
from sklearn.preprocessing import StandardScaler
from zca import ZCA

from pycytominer.operations.transform import Spherize

def normalize_helper(x: np.ndarray, method: str) -> np.ndarray:
    """If method is ZCA-cor, center and scale it to unit variance."""
    if method == 'ZCA-cor':
        return StandardScaler(with_mean=True, with_std=True).fit_transform(x)
    else:
        return x


def sklearn_ref(x: np.ndarray, method: str) -> np.ndarray:
    """ZCA via sklearn.PCA"""
    x = normalize_helper(x, method=method)
    p = PCA(whiten=True).fit(x)
    y = p.transform(x)
    y = y @ p.components_  # rotate back (ZCA)
    return y


def zca_ref(x: np.ndarray, method: str) -> np.ndarray:
    """ZCA from the zca package"""
    x = normalize_helper(x, method=method)
    return ZCA().fit_transform(x)


def err(y) -> float:
    """max |cov(Y) - I|; 0 iff Y is white"""
    c = np.cov(np.asarray(y), rowvar=False)
    return float(np.abs(c - np.eye(c.shape[0])).max())


d = 6
x, _ = make_blobs(n_samples=400, n_features=d, centers=4, random_state=0)
df = pd.DataFrame(x, columns=[f"f{i}" for i in range(d)])

METHODS = ['ZCA', 'ZCA-cor']

res = pd.DataFrame(
    [
        {
            "sklearn": err(sklearn_ref(df, method=method)),
            "zca": err(zca_ref(df, method=method)),
            "Spherize": err(Spherize(method=method).fit_transform(df)),
            "Spherize(centered input)": err(Spherize(method=method).fit_transform(df - df.mean())),
        }
        for method in METHODS
    ],
    index=METHODS,
)
res.index.name = "max |cov-I|"
pd.set_option("display.float_format", "{:.4f}".format)
print(res.T)
```

The result is this :
```
max |cov-I|                 ZCA  ZCA-cor
sklearn                  0.0000   0.0000
zca                      0.0000   0.0000
Spherize                 0.6321   0.9662
Spherize(centered input) 0.0000   0.9852
```

### Expected behavior

The expected output would be zeroes also in the Spherize row.

agu2347:
## Bug

`Spherize` (PCA/ZCA/PCA-cor/ZCA-cor whitening transform) does not actually whiten its output: `cov(transform(X))` is far from the identity matrix, for every method, whenever the input isn't already centered (and, for the `-cor` variants, standardized) by the caller.

Fixes #747

## Root cause

`Spherize.fit()` builds `X_transformed` (mean-centered, and for `PCA-cor`/`ZCA-cor` additionally standardized) from the input `X`, but then computes the SVD used to build the whitening matrix `self.W` on the raw, untransformed `X`:

```python
_, Sigma, Vt = np.linalg.svd(X, full_matrices=True)
```

The derivation directly above this line in the existing code explicitly assumes *"X is mean-centered (zero mean for each feature)"* -- a precondition silently violated by passing the wrong variable. There's no error and no obviously wrong shape, so the bug is easy to miss: `transform()`'s output looks like a normal DataFrame, it's just not actually white. This matches the reporter's own observation exactly -- pre-centering the input before calling `Spherize` "fixes" ZCA (because a pre-centered `X` and `X_transformed` now coincide), but not ZCA-cor (which additionally needs unit-variance scaling that pre-centering alone doesn't provide).

## Fix

Use `X_transformed` (not `X`) in the SVD call. One-line change.

## Testing

Reproduced the reporter's own diagnostic (`max|cov(transform(X)) - I|`) against the current `main` branch and confirmed the failure:

```
PCA        max|cov(Y)-I| = 0.888968
ZCA        max|cov(Y)-I| = 0.632116
PCA-cor    max|cov(Y)-I| = 0.982251
ZCA-cor    max|cov(Y)-I| = 0.966227
```

With the fix, all four are `~0` (`1e-6` or below).

Added `test_spherize_whitens_data` to `tests/test_operations/test_transform.py`. The existing `test_spherize` only checks whitening to the nearest *integer* (via `.round()` on the covariance matrix before comparing) -- confirmed this passes both **with and without** this fix, i.e. it can't actually catch this bug. The new test uses data with an off-origin mean (so the bug can't hide behind already-near-zero-centered test data) and asserts `max|cov(Y) - I| < 1e-6` for every method. Reverting just the source fix makes the new test fail with `AssertionError: method=PCA: transformed data is not white (max|cov(Y) - I| = 0.987...)`, while the pre-existing (too-coarse) tests still pass either way -- confirming this is a genuine, tighter regression guard.

Ran `pytest tests/test_operations/ tests/test_normalize.py` (covers `Spherize` and its usage via `normalize()`): 73 passed, no failures.

`ruff check` and `ruff format --check` (per `.pre-commit-config.yaml`) are clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data sphering so covariance results remain accurate when centering or standardization is enabled.
  * Corrected handling of off-center input data across supported sphering methods.

* **Tests**
  * Added regression coverage to verify near-identity covariance results for centered, off-center data.

### DIFF
--- a/pycytominer/operations/transform.py
+++ b/pycytominer/operations/transform.py
@@ -139,7 +139,17 @@ class Spherize(BaseEstimator, TransformerMixin):
         # Get the eigenvalues and eigenvectors of the covariance matrix
         # by computing the SVD on the data matrix
         # https://stats.stackexchange.com/q/134282/8494
-        _, Sigma, Vt = np.linalg.svd(X, full_matrices=True)
+        #
+        # This must be the SVD of X_transformed (the centered/standardized
+        # matrix computed above), not of the raw input X: the derivation
+        # below explicitly assumes "X is mean-centered (zero mean for each
+        # feature)", and for PCA-cor/ZCA-cor X_transformed is additionally
+        # scaled to unit variance. Computing the SVD of the untransformed X
+        # silently breaks that precondition -- the resulting W still runs
+        # without error, but the transformed output is not actually
+        # whitened (cov(XW) != I) whenever X isn't already centered (and,
+        # for the -cor variants, scaled) on its own.
+        _, Sigma, Vt = np.linalg.svd(X_transformed, full_matrices=True)
 
         # if n <= d then Sigma has shape (n,) so it will need to be expanded to
         # d filled with the value r'th element of Sigma
--- a/tests/test_operations/test_transform.py
+++ b/tests/test_operations/test_transform.py
@@ -49,6 +49,41 @@ def test_spherize():
             assert int(result) == expected_result
 
 
+def test_spherize_whitens_data():
+    """The transformed output of Spherize should be white: cov(XW) should
+    be (near) the identity matrix for every method/center combination.
+
+    test_spherize() above only checks this to the nearest integer (via
+    `.round()`), which is far too coarse to catch a whitening bug: it was
+    passing even when `fit()` computed the SVD of the raw, uncentered
+    input instead of the mean-centered (and, for the -cor variants,
+    standardized) data -- see
+    https://github.com/cytomining/pycytominer/issues/747.
+    """
+    rng = np.random.default_rng(0)
+    # An off-origin mean makes an uncentered-SVD bug visible: with
+    # data centered at/near the origin already, `svd(X)` and
+    # `svd(X_transformed)` coincide, masking the bug.
+    data_off_center = pd.DataFrame(
+        rng.normal(loc=100, scale=25, size=(200, 5)),
+        columns=list("abcde"),
+    )
+
+    spherize_methods = ["PCA", "ZCA", "PCA-cor", "ZCA-cor"]
+    for method in spherize_methods:
+        scaler = Spherize(method=method, center=True)
+        scaler = scaler.fit(data_off_center)
+        transform_df = scaler.transform(data_off_center)
+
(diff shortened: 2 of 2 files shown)


gwaybio (approved):
thanks for catching this @agu2347 and @danr - it looks great. Thanks also for the clean PR. I'll merge this now and mint a new release.

Issue #427, opened 2024-08-22
FeatureRequest: Machine Learning Data Split Functionality


MikeLippincott:
### Feature type

- [X] Add new functionality

- [X] Change existing functionality

### General description of the proposed functionality

This functionality is multi-part:
Currently, machine learning data splits are performed after normalization and feature selection. 
This poses a potential for data leakage into the models.
The proposed fix is the implement functions that perform data splits prior to normalization.
To implement this the normalization and feature selection would need to be applied to the training split and then propagated to the validation, testing, [*holdout] splits.
I envision most of this functionality needing to be carried out by the user.
To do so, I suggest updating the noramlize and feature selection function and implement two more functions:
1. Normalize function: Save the transformation performed on the training data
2. Implement a normalize transform function to propagate the saved transformation to testing data splits
3. Feature select function: Save the feature selected columns list
4. Implement a feature selection column propagation to each test data split.

### Feature example

change function: 
def noramlize(*args, **kwargs, blah, blah, blah, save_xform:bool = False):
    normalize_function_magic_that_is_happinging

    if not save_xform:
        continue
    else:
        save_the_transform (probably as a numpy array), depending on method it could be parameters too (mean, std)

new function:
def apply_xform(*args, **kwargs):
    apply the saved xform the test data splits here

change function:
def feature_select(*args, **kwargs, save_selected_feature_list: bool = False):
    feature selection magic

    if not save_selected_feature_list:
        continue
    else:
        save the list of features to apply to a dataset

new function:
def apply_selected_features(*args, **kwargs):
    apply the selected features from train split to test split.

MikeLippincott:
This is something that I can implement. Do I need approval prior to opening a PR?

d33bs:
Thanks for adding this issue @MikeLippincott ! No special approval required before opening a PR, please feel free to propose changes and open when ready.

On the enhancement development: based on how you outlined the issue would it be possible to add a test which checks for leakage you mentioned? I imagine this would help prove the new capabilities in addition to making sure future changes don't reintroduce leakage. Totally open to your thoughts here (this isn't a requirement).

PR #760, opened 2026-09-14
add normalize fit transform functionality

MikeLippincott:
# Description
This PR updates the normalize function to add the ability to fit or apply transformations to data. This could be helpful for ML splits, and/or if data are very large and one wants to apply the transformation to chunked data.

# Checklist

## Summary by CodeRabbit

- **New Features**
  - Save fitted normalization transforms as reusable `.joblib` files.
  - Apply saved transforms without refitting.
  - Supports standardize, robustize, MAD robustize, and spherize methods.
  - Preserves feature order and normalization method in saved files.
  - Automatically restores inferred features when applying saved transforms.
  - Validates supplied features and reports mismatches.
  - Exposes transform save/load utilities through the package namespace.


MikeLippincott:
tests fail on main as well my branch - not sure which/how to reconcile. Will circle back here once addressed!

pycytominer/cyto_utils/normalize_transform_io.py
```diff
@@ -0,0 +1,88 @@
...
+        Where to write the transform. Written as a joblib file with a
+        ".joblib" suffix, containing the fitted scaler, method name, and
+        feature order.
+    """
+    method = method.lower()
+    if method not in avail_methods:
+        raise ValueError(f"method must be one of {avail_methods}")
+
+    output_path = _joblib_path(output_file)
+    output_path.parent.mkdir(parents=True, exist_ok=True)
```
gwaybio, line 48: Pycytominer should not be opinionated to create folders. Instead, we default to the behavior in the external function itself. See [output](https://github.com/cytomining/pycytominer/blob/main/pycytominer/cyto_utils/output.py) for examples

pycytominer/normalize.py
```diff
@@ -108,6 +114,21 @@ def normalize(
         Number of cumulative distribution function landmarks used for the inverse
         normal transformation. Values larger than the number of samples are capped
         at the number of samples. Only used when ``method="inverse_normal"``.
+    transform_output_file : str, optional
+        If provided, save the fitted transform to this path so it can be
+        reapplied later (e.g. to other subsets of data) via
+        `fitted_transform_file`. The transform is written as a single joblib
+        file with a ".joblib" suffix. Ignored if `fitted_transform_file` is
+        also provided, since no new fit is performed in that case. Not
+        supported for `method="inverse_normal"`.
```
gwaybio, line 123: why not supported for inverse_normal?

pycytominer/normalize.py
```diff
@@ -108,6 +114,21 @@ def normalize(
...
+    fitted_transform_file : str, optional
+        If provided, skip fitting a new transform and instead load a previously
+        saved transform (written by a prior call using `transform_output_file`)
+        and apply it directly to `profiles`. The `method`, `samples`, and
+        method-specific fitting parameters (e.g. `mad_robustize_epsilon`,
+        `spherize_center`, `spherize_method`, `spherize_epsilon`) are ignored, since
```
gwaybio, line 129: the docs indicate these parameters are saved, but it doesn't look like the function is actually saving them

pycytominer/normalize.py
```diff
@@ -189,39 +210,54 @@ def normalize(
...
-    elif method == "spherize":
-        scaler = Spherize(
-            center=spherize_center,
-            method=spherize_method,
-            epsilon=spherize_epsilon,
-            return_numpy=True,
+    if fitted_transform_file is not None:
+        # Load a previously-fit transform and apply it directly, skipping
+        # `method`/`samples`/method-specific fitting parameters entirely
+        fitted_scaler, method, saved_features = load_normalize_transform(
```
gwaybio, line 216: i agree that Pycytominer should save more than just the scalar to support data provenance, but currently the "method" variable isn't doing any work - should it be doing more? For example, should a user require that the input "method" matches the fitted scalar when a user is applying a precomputed scalar?

pycytominer/normalize.py
```diff
@@ -189,39 +210,54 @@ def normalize(
...
+            fitted_transform_file
         )
-    elif method == "inverse_normal":
-        scaler = InverseNormalTransform(n_quantiles=inverse_normal_n_quantiles)
 
-    if features == "infer":
-        features = infer_cp_features(profiles, image_features=image_features)
+        if features == "infer":
+            features = saved_features
+        elif list(features) != list(saved_features):
```
gwaybio, line 222: Should `saved_features` do more work? Should it reorder? 

What if a user inputs `features = "infer"`, this will raise a ValueError right? Should it?

pycytominer/normalize.py
```diff
@@ -319,12 +355,21 @@ def normalize(
...
+    if fitted_transform_file is None:
+        # Fit the sklearn scaler
+        if samples == "all":
+            fitted_scaler = scaler.fit(feature_df)
+        else:
+            # Subset to only the features measured in the sample query
+            fitted_scaler = scaler.fit(profiles.query(samples).loc[:, features])
+
+        if transform_output_file is not None:
+            save_normalize_transform(
```
gwaybio, line 367: what about other parameters? Spherize, for example, should save associated parameters? It seems our goal of data provenance stops short of saving parameters here.

tests/test_normalize.py
```diff
@@ -1040,3 +1040,92 @@ def test_output_type():
...
     # check to make sure both dataframes are the same regardless of the output_type
     pd.testing.assert_frame_equal(csv_df, parquet_df)
+
+
+@pytest.mark.parametrize(
+    "method", ["standardize", "robustize", "mad_robustize", "spherize"]
+)
+def test_normalize_transform_save_and_apply(method):
+    """
+    Fitting a transform on the control samples and saving it via
```
gwaybio, line 1050: nice!

tests/test_normalize.py
```diff
@@ -1040,3 +1040,92 @@ def test_output_type():
...
+    ]
+
+    np.testing.assert_allclose(
+        applied_result.loc[:, numeric_cols].values,
+        expected_subset_result.loc[:, numeric_cols].values,
+        rtol=1e-6,
+        atol=1e-8,
+    )
+
+    # features="infer" should recover the saved feature list and give the
```
gwaybio, line 1092: you are implicitly using features="infer" in the code below by accepting defaults, but this is fragile and will break if the default ever changes.

walkthroughs/using_cytotable_data_in_pycytominer.ipynb
```diff
@@ -1,559 +1,693 @@
 {
-  "cells": [
```
gwaybio, line 2: It is concerning that this file changed so much - was this intentional?

gwaybio (changes requested):
Thanks for this contribution! I made specific comments throughout, please consider these and the general comments below.

- Align appropriate data provenance
- Determine how much work each saved parameter in the joblib file should perform
- Add tests to confirm the joblib file successfully stores these additional parameters
- Update CLI to new functionality

This is also related to #427 - This PR represents the first step toward this goal; Mike, please think through if this PR (#760) should get all the way there. Are we all the way there already?

pycytominer/normalize.py
gwaybio: sklearn scalars also include the `feature_names_in_` attribute - what would it take to use this feature rather than rebuilding our own method? Are there version restrictions? Would we need to update our custom scalars here https://github.com/cytomining/pycytominer/blob/22654a5793b538190e4ea53d2b4ff1523574c975/pycytominer/operations/transform.py?

